Issue #54, opened 2026-03-15
WB-BH-05 Authoring loop sanity


skulmakov-oss:
## Scope
Review open → edit → save → check/compile/run/disasm from the editor shell.

## Acceptance Criteria
Dirty and saving indicators stay accurate; auto-save before actions only happens where intended; current-file actions never require manual path workarounds; results flow back into jobs/inspectors predictably.

## Priority
High

## Source
Generated from output/spreadsheet/workbench_beta_hardening_backlog.csv.

skulmakov-oss:
Started in stacked PR #68. This slice hardens the editor authoring loop by making autosave authoritative for current-file actions, adding current-file run/disasm, and routing results back into Diagnostics and Inspect predictably.

skulmakov-oss:
Integrated into main via fast-forward at 02a27d8.

PR #68, opened 2026-03-15
WB-BH-05 Authoring loop sanity

skulmakov-oss:
## Summary\n- make current-file authoring actions stop when autosave fails instead of silently continuing against stale disk state\n- add current-file run and disasm actions that derive the canonical artifact path without manual path workarounds\n- route check/failure flows to Diagnostics and inspectable run/disasm flows to Inspect predictably\n\n## Validation\n- npm run lint\n- npm run build\n- cargo check --manifest-path apps/workbench/src-tauri/Cargo.toml\n- cargo tauri build --debug --no-bundle\n\nRefs #54

### DIFF
--- a/apps/workbench/src/App.tsx
+++ b/apps/workbench/src/App.tsx
@@ -808,12 +808,12 @@ function App() {
 
   async function saveEditorFile(relativePath: string) {
     if (!selectedWorkspace) {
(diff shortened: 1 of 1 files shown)


skulmakov-oss:

Issue #55, opened 2026-03-15
WB-BH-06 Diagnostics fidelity and source navigation


skulmakov-oss:
## Scope
Audit parse, type, module, verify, and runtime diagnostics preservation.

## Acceptance Criteria
Severity, code, file, and range stay intact when present in tool output; grouped filters remain stable; opening a diagnostic lands in the correct file context; raw diagnostic block stays accessible.

## Priority
High

## Source
Generated from output/spreadsheet/workbench_beta_hardening_backlog.csv.

skulmakov-oss:
Started in stacked PR #69. This slice hardens diagnostics fidelity at the UX boundary by preserving jump context into the editor shell and landing source navigation at the reported file/line/column when the CLI output provides them.

skulmakov-oss:
Integrated into main via fast-forward at 02a27d8.

PR #69, opened 2026-03-15
WB-BH-06 Diagnostics fidelity and source navigation

skulmakov-oss:
## Summary\n- preserve diagnostics jump context all the way into the editor shell instead of only opening the file\n- focus the editor at the reported line and column when diagnostics provide them\n- keep raw diagnostic blocks and grouped filters intact while routing source navigation predictably back to /project\n\n## Validation\n- npm run lint\n- npm run build\n- cargo check --manifest-path apps/workbench/src-tauri/Cargo.toml\n- cargo tauri build --debug --no-bundle\n\nRefs #55

### DIFF
--- a/apps/workbench/src/App.tsx
+++ b/apps/workbench/src/App.tsx
@@ -116,6 +116,19 @@ type WorkspaceOpenOptions = {
   successMessage?: string | null
 }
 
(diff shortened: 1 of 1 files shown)


skulmakov-oss:

PR #73, opened 2026-03-15
WB-BH-10 inspectors readability on real artifacts

skulmakov-oss:
## Summary\n- make inspector artifact provenance explicit instead of implying a second model\n- label derived summaries and derived signals as secondary to preserved stdout/stderr\n- keep raw output authoritative while improving readability in the inspect ledger and detail view\n\n## Validation\n- npm run lint\n- npm run build\n- cargo check --manifest-path apps/workbench/src-tauri/Cargo.toml\n- cargo tauri build --debug --no-bundle


skulmakov-oss:

Issue #94, opened 2026-03-21
D20-C06 compound assignment


skulmakov-oss:
## Scope
- Pack: Density-20
- Wave: Wave C - API and Data Density
- Owner crate: sm-ir
- Touch crates: sm-front;sm-sema;sm-ir
- Priority: P0

## Acceptance Criteria
Operators like += and &= desugar to read-modify-write over existing scalar semantics, preserve overflow and type rules, and ship with parser, IR, and verified-path tests.

## PR Split
PR-1 syntax and operator table; PR-2 sema legality and IR desugaring; PR-3 tests/docs

## Boundary Rule
- Keep this change inside Semantic Core unless a later architecture decision explicitly broadens scope.
- Do not couple this issue to prom-* work in the first implementation wave.

skulmakov-oss:
Implemented and integrated on main at 62854ad.

PR #134, opened 2026-03-22
D20-C06 compound assignment

skulmakov-oss:
Refs #94.

Scope:
- narrow D20 slice only
- no broader language/runtime expansion beyond this feature
- stacked in validated dependency order

Validation:
- targeted crate tests for the touched layers
- cargo test --workspace


skulmakov-oss:
Integrated via fast-forward of main to 62854ad.

Issue #97, opened 2026-03-21
D20-D03 loop as expression plus break value


skulmakov-oss:
## Scope
- Pack: Density-20
- Wave: Wave D - Compile-time Contract Density
- Owner crate: sm-sema
- Touch crates: sm-front;sm-sema;sm-ir
- Priority: P2

## Acceptance Criteria
loop can yield a value via break expr with deterministic type unification, lowering stays within existing control-flow machinery, and verified tests cover success and type-error paths.

## PR Split
PR-1 syntax and AST; PR-2 sema break-value typing and lowering; PR-3 tests/docs

## Boundary Rule
- Keep this change inside Semantic Core unless a later architecture decision explicitly broadens scope.
- Do not couple this issue to prom-* work in the first implementation wave.

skulmakov-oss:
Integrated on main at bf0116c via mini-wave consolidation.

PR #146, opened 2026-03-22
D20-D03 loop as expression plus break value

skulmakov-oss:
Implements #97 as a narrow control-flow slice.\n\nScope:\n- loop as expression\n- break value typing\n- existing label/store/load lowering path\n- docs/tests sync\n\nValidation:\n- cargo test -p sm-front\n- cargo test -p sm-ir\n- cargo test --workspace


skulmakov-oss:

Issue #101, opened 2026-03-21
D20P-B02 UFCS and method-call sugar


skulmakov-oss:
## Scope
- Pack: Density-20 Plus
- Wave: Wave B+ - Call Density
- Owner crate: sm-sema
- Touch crates: sm-front;sm-sema;sm-ir
- Priority: P2

## Acceptance Criteria
Method-style calls over existing free functions resolve through ordinary call semantics, name lookup rules are explicit, and disasm/verified tests prove no object model or hidden dispatch path was introduced.

## PR Split
PR-1 parse surface and precedence; PR-2 sema resolution and IR desugaring; PR-3 tests/docs

## Boundary Rule
- Keep this change inside Semantic Core unless a later architecture decision explicitly broadens scope.
- Do not couple this issue to prom-* work in the first implementation wave.

skulmakov-oss:
Integrated on main at bf0116c via mini-wave consolidation.

PR #147, opened 2026-03-22
D20P-B02 UFCS and method-call sugar

skulmakov-oss:
Implements #101 as postfix call sugar over existing free-function calls.\n\nScope:\n- receiver.name(args...) surface\n- ordinary call desugaring\n- no field access or object model\n- docs/tests sync\n\nValidation:\n- cargo test -p sm-front\n- cargo test -p sm-ir\n- cargo test --workspace


skulmakov-oss:

Issue #102, opened 2026-03-21
D20P-C01 let-else


skulmakov-oss:
## Scope
- Pack: Density-20 Plus
- Wave: Wave C+ - Control Density
- Owner crate: sm-sema
- Touch crates: sm-front;sm-sema;sm-ir
- Priority: P1

## Acceptance Criteria
let-else lowers to explicit failure/early-exit flow over existing guard semantics, works first for destructuring-capable patterns only, and diagnostics explain unsupported targets and pattern forms.

## PR Split
PR-1 surface syntax; PR-2 sema target restrictions and lowering; PR-3 tests/docs

## Boundary Rule
- Keep this change inside Semantic Core unless a later architecture decision explicitly broadens scope.
- Do not couple this issue to prom-* work in the first implementation wave.

skulmakov-oss:
Integrated on main at bf0116c via mini-wave consolidation.

PR #148, opened 2026-03-22
D20P-C01 let-else

skulmakov-oss:
Implements #102 as a narrow tuple-target let-else slice.\n\nScope:\n- tuple-only let-else\n- else return only\n- flat name/_/quad-literal items\n- explicit TupleGet/CmpEq/Ret lowering\n- docs/tests sync\n\nValidation:\n- cargo test -p sm-front\n- cargo test -p sm-ir\n- cargo test --workspace


skulmakov-oss:

PR #166, opened 2026-03-23
REC2-02 record let-else over explicit field patterns

skulmakov-oss:
## Summary
- add narrow record let-else over explicit field patterns
- keep the slice refutable only through explicit quad literal field targets
- add lowering, verifier, VM, and spec coverage for the new surface

## Validation
- cargo test -p sm-front
- cargo test -p sm-ir
- cargo test -p sm-verify
- cargo test -p sm-vm
- cargo test --workspace


skulmakov-oss:
Integrated via fast-forward of main to db973fa

PR #168, opened 2026-03-23
D20P-D01 record punning and field shorthand

skulmakov-oss:
## Summary
- add record field shorthand sugar inside canonical nominal record forms
- support punning in record literals, copy-with, destructuring bind, and record let-else
- keep the feature as parser-level sugar only with no separate runtime path

## Validation
- cargo test -p sm-front
- cargo test -p sm-ir
- cargo test --workspace


skulmakov-oss:
Integrated via fast-forward of main to db973fa

PR #370, opened 2026-04-27
GAP: scope line ending normalization for source files

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_line_ending_normalization_scope.md`
- scope the frontend gap exposed by `weather_station.sm`: LF/CRLF fragility
- define acceptance tests for line-ending-independent parsing and stable diagnostics

## Scope
- no parser changes yet
- no release claim widening

## Validation

- normalize line endings at the frontend input boundary
- add LF/CRLF fixtures and diagnostics regression tests

### DIFF
--- a/docs/roadmap/full_readiness/gap_line_ending_normalization_scope.md
+++ b/docs/roadmap/full_readiness/gap_line_ending_normalization_scope.md
@@ -1,0 +1,78 @@
+# GAP: Line Ending Normalization
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / frontend robustness  
+Source trigger: `weather_station.sm` canonical-program trial
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes

This gap should be fixed at the **lexer input boundary**, before tokenization.

Rationale:
- line-ending normalization in the parser is too late;
- parser-level normalization risks shifting error spans after tokens already exist;
- diagnostics must preserve stable line/column positions across LF and CRLF variants.

Recommended test matrix:

```text
LF-only source
CRLF-only source
mixed LF+CRLF source in one file
source with trailing newline
source without trailing newline
```

Acceptance refinement:
- `smc check` should behave the same for LF and CRLF versions of the same source;
- mixed line-ending policy must be explicit: either normalize deterministically or reject with a dedicated diagnostic;
- span/caret diagnostics must remain stable after normalization.

PR #371, opened 2026-04-27
GAP: scope comment lexing robustness for source files

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_comment_lexing_robustness_scope.md`
- scope the frontend gap exposed by `weather_station.sm`: leading comments / decorative comment text fragility
- define acceptance tests for ordinary line comments and explicit non-ASCII comment policy

## Scope
- no parser changes yet
- no release claim widening

## Validation

- make line comments robust in the Rust-like executable frontend
- add positive comment fixtures and diagnostics regressions

### DIFF
--- a/docs/roadmap/full_readiness/gap_comment_lexing_robustness_scope.md
+++ b/docs/roadmap/full_readiness/gap_comment_lexing_robustness_scope.md
@@ -1,0 +1,79 @@
+# GAP: Comment and Source Text Lexing Robustness
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / frontend robustness  
+Source trigger: `weather_station.sm` canonical-program trial
+
+## Problem
+
+The weather-station trial exposed that valid-looking source files with leading comments or decorative comment text can fail before ordinary parsing.
+
+A production-ready source frontend must handle ordinary source comments predictably.
+
+## Goal
+
+Make comment handling explicit and robust for the Rust-like executable Semantic surface.
+
+## Scope
+
+Define and implement comment handling policy for:
+
+- leading `//` line comments before the first `fn`;
+- blank lines around comments;
+- comment-only lines between declarations;
+- ASCII comment text;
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes

This PR actually covers two distinct frontend robustness bugs:

1. **Leading comments**
   - The parser/frontend path should not fail because the first token-like text in a file is `//`.
   - A line comment before the first top-level item should be skipped or diagnosed by explicit policy.

2. **Non-ASCII inside comments**
   - The lexer should skip UTF-8 bytes inside `// ...` until newline without trying to tokenize them as identifiers/operators.
   - UTF-8 should be allowed only inside comments and string literals for this slice.
   - Do **not** widen identifier syntax to Unicode in this PR.

Boundary rule:

```text
UTF-8 in comments/string literals: admitted or explicitly diagnosed
UTF-8 in identifiers: out of scope / must remain rejected unless separately scoped
```

Recommended tests:

```text
// leading comment
fn main() { return; }

// comment with Cyrillic/Kazakh text
fn main() { return; }

fn main() {
    // inline UTF-8 comment
    return;
}

fn функция() { return; }  // should remain rejected unless Unicode identifiers are separately admitted
```

skulmakov-oss:
## Probe Pack v0 update

Current local probe result shows that UTF-8 comments already pass the current `smc` binary:

```text
probe_comments_utf8: check passed
probe_comments_utf8: compile passed
compiled size: 163 bytes
```

This changes the implementation reading of this PR:

- the UTF-8-in-comments path appears already satisfied in the current binary;
- this PR may still be useful as a hardening/regression scope;
- remaining value is to add explicit fixtures so this behavior cannot regress;
- Unicode identifiers should still remain out of scope unless separately admitted.

Recommended follow-up:

```text
Keep as regression-hardening scope, not as a confirmed active blocker.
```

PR #372, opened 2026-04-27
GAP: scope f64 relational operator surface

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_f64_relational_surface_scope.md`
- scope the everyday-expressiveness gap exposed by `weather_station.sm`: direct `f64` relational operators
- define semantic policy questions, positive/negative tests, and acceptance criteria

## Scope
- no operator implementation yet
- no release claim widening

## Validation

- support admitted `f64 < <= > >=` paths through source → IR → SemCode → verify → VM
- remove `gt`/`lt` helper workaround from canonical examples once implemented

### DIFF
--- a/docs/roadmap/full_readiness/gap_f64_relational_surface_scope.md
+++ b/docs/roadmap/full_readiness/gap_f64_relational_surface_scope.md
@@ -1,0 +1,95 @@
+# GAP: f64 Relational Operator Surface
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / everyday expressiveness  
+Source trigger: `weather_station.sm` canonical-program trial
+
+## Problem
+
+The weather-station trial had to define helper functions such as `gt(x, y)` and `lt(x, y)` using `abs()` because direct `f64` relational operators were not usable in that path.
+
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes

The existing `gt()` / `lt()` helper workaround via `abs()` is acceptable as a temporary source-level workaround, but it should not become canonical language style.

Implementation should keep two distinct type-checker paths:

```text
f64/i32/u32/fx relational operators: admitted by explicit numeric rules
quad relational operators: rejected
```

Important language boundary:
- `>` / `<` on `f64` and admitted numeric families should be allowed once implemented;
- `>` / `<` on `quad` must remain forbidden;
- `quad` ordering is not part of this feature.

Required negative regression:

```semantic
fn bad(q: quad) -> bool {
    return q > T;
}
```

Expected result: deterministic type error, not lowering failure.

Recommended positive cases:

```semantic
fn f(x: f64, y: f64) -> bool { return x > y; }
fn g(x: i32, y: i32) -> bool { return x < y; }
```

If `NaN` can be represented or produced, its relational semantics must be explicitly documented before merge.

PR #374, opened 2026-04-27
GAP: scope record executable readiness for canonical programs

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_record_executable_readiness_scope.md`
- scope the practical-programming gap exposed by weather/snake trials: record values need a reliable executable path
- define minimal record surface, required tests, acceptance criteria, and explicit out-of-scope items

## Scope
- no record implementation changes yet
- no UI / Workbench scope
- no release claim widening

## Validation

- support minimal record declaration/literal/field-access/function-parameter path through source → IR → SemCode → verify → VM
- add positive/negative record fixtures and canonical record-based example

### DIFF
--- a/docs/roadmap/full_readiness/gap_record_executable_readiness_scope.md
+++ b/docs/roadmap/full_readiness/gap_record_executable_readiness_scope.md
@@ -1,0 +1,106 @@
+# GAP: Record Executable Readiness for Canonical Programs
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / stable language surface  
+Source trigger: initial snake/weather program trials
+
+## Problem
+
+Canonical practical programs naturally want `record` values for state, sensor packets, and domain objects. In the weather-station and snake trials, record-based source was not reliable enough for the current executable example path.
+
+This creates a mismatch between the documented language surface and the practical canonical examples that should prove the language.
+
+## Goal
+
+Close the minimal executable record path needed for canonical non-UI programs.
+
+## Required minimal surface
+
+```semantic
+record SensorState {
+    temp: f64,
+    pressure: f64,
+}
+
+fn read_temp(s: SensorState) -> f64 {
+    return s.temp;
+}
+
+fn main() {
+    let s: SensorState = SensorState { temp: 28.5, pressure: 1013.0 };
+    return;
+}
+```
+
+## Scope
+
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes

This is the widest technical gap in the group because the minimal record path crosses multiple layers:

```text
frontend -> semantic analysis -> IR lowering -> SemCode emission -> verifier -> VM
```

Minimum surface to close first:

```semantic
record Point { x: f64, y: f64 }

fn use_point(pt: Point) -> f64 {
    return pt.x;
}

fn main() {
    let p: Point = Point { x: 1.0, y: 2.0 };
    let v: f64 = p.x;
    return;
}
```

Important diagnostic clue for Codex:
- current `record` appears to parse partially, but valid syntax can fail with `expected '}'`;
- likely cause: parser does not register `record` correctly as a top-level file item in the executable grammar path, or record body parsing exits through the wrong production.

Implementation order recommendation:

```text
1. parser top-level item registration for `record`
2. record declaration AST/sema ownership check
3. record literal type checking
4. field access type checking
5. IR representation/lowering
6. SemCode emission/version handling
7. verifier admission
8. VM runtime value/field access
9. positive and negative fixtures
```

Keep out of scope for this slice:
- record destructuring;
- record let-else;
- punning;
- copy-with;
- mutable update;
- schema-to-record generation.

skulmakov-oss:
## Probe Pack v0 update

Current local probe result shows that the minimal executable record path already passes in the current `smc` binary:

```text
probe_record_minimal:
  check: pass
  compile: pass
  verify: pass
  run-smc: pass
  .smc size: 348 bytes
  functions: 3
  header: SEMCODE1
  epoch: 0.2
```

Disassembly confirms the expected executable path exists:

```text
MAKE_RECORD
RECORD_GET
STORE_VAR / LOAD_VAR
CALL
RET
```

This changes the implementation reading of this PR:

- record declaration/literal/field access/function parameter baseline appears already functional;
- this PR should be treated as regression-capture and documentation alignment, not as a confirmed active implementation blocker;
- the remaining useful work is adding canonical fixtures and negative tests for missing/extra/duplicate/wrong-type fields.

Recommended follow-up:

```text
Convert from implementation-blocker scope to record executable regression-hardening scope.
```

PR #376, opened 2026-04-27
GAP: scope match terminal CFG recognition

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_match_terminal_cfg_scope.md`
- scope the Probe Pack v0 finding that `match` with all returning arms can pass check but fail compile return analysis
- define terminality rules, minimal reproducer, tests, and acceptance criteria

## Probe evidence
- `check` can pass for a function where every `match` arm returns
- `compile` can fail with `function may exit without returning F64`
- adding artificial fallthrough `return <value>;` after `match` works around the issue

## Scope
- no CFG/control-flow implementation yet
- no release claim widening

## Validation

### DIFF
--- a/docs/roadmap/full_readiness/gap_match_terminal_cfg_scope.md
+++ b/docs/roadmap/full_readiness/gap_match_terminal_cfg_scope.md
@@ -1,0 +1,98 @@
+# GAP: Match Terminal Control-Flow Recognition
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / source semantics and lowering  
+Source trigger: Probe Pack v0
+
+## Problem
+
+Probe Pack v0 shows that a function containing a `match` whose every arm returns can pass `smc check` but fail at compile time with:
+
+```text
+function 'classify' may exit without returning F64
+```
+
+A manual fallthrough `return <value>;` after the `match` works around the issue.
+
+This indicates that the compiler/check path does not currently recognize fully-returning `match` expressions/statements as terminal control flow.
+
+## Goal
+
+Teach the source/lowering/control-flow analysis that a `match` is terminal when all admitted arms are terminal.
+
+## Minimal reproducer
+
+```semantic
+fn classify(q: quad) -> f64 {
+    match q {
+        T => { return 1.0; }
+        F => { return 0.0; }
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes — check/compile stage gap

Minimal repro pattern:

```semantic
fn f(q: quad) -> f64 {
    match q {
        T => { return 1.0; }
        F => { return 2.0; }
        S => { return 3.0; }
        _ => { return 0.0; }
    }
}
```

Observed behavior:

```text
smc check:   OK
smc compile: function may exit without returning F64
```

This is a stage-consistency gap:

- semantic/source checking accepts the match form;
- compile/lowering return analysis does not treat a `match` with wildcard/default arm and all-returning branches as terminal.

Implementation options:

1. propagate match exhaustiveness/terminality information from sema into the compile/CFG stage; or
2. compute terminality directly in the CFG/lowering pass for the admitted match form.

Required rule for this slice:

```text
match is terminal if every admitted arm is terminal and the match is exhaustive by current source semantics.
```

Workaround currently required:

```semantic
match q { ... }
return 0.0; // artificial fallthrough
```

The fix should remove the need for this unreachable fallthrough return.

skulmakov-oss:
## Implementation notes — check/compile stage gap

Minimal repro pattern:

```semantic
fn f(q: quad) -> f64 {
    match q {
        T => { return 1.0; }
        F => { return 2.0; }
        S => { return 3.0; }
        _ => { return 0.0; }
    }
}
```

Observed behavior:

```text
smc check:   OK
smc compile: function may exit without returning F64
```

This is a stage-consistency gap:

- semantic/source checking accepts the match form;
- compile/lowering return analysis does not treat a `match` with wildcard/default arm and all-returning branches as terminal.

Implementation options:

1. propagate match exhaustiveness/terminality information from sema into the compile/CFG stage; or
2. compute terminality directly in the CFG/lowering pass for the admitted match form.

Required rule for this slice:

```text
match is terminal if every admitted arm is terminal and the match is exhaustive by current source semantics.
```

Workaround currently required:

```semantic
match q { ... }
return 0.0; // artificial fallthrough
```

The fix should remove the need for this unreachable fallthrough return.

PR #377, opened 2026-04-27
GAP: scope diagnostic span accuracy

skulmakov-oss:
## Summary
- add `docs/roadmap/full_readiness/gap_diagnostic_span_accuracy_scope.md`
- scope the Probe Pack v0 finding that semantic errors often report `line 1:1` instead of the real offending expression
- define span targets, snapshot tests, and acceptance criteria

## Probe evidence
- `if q` where `q: quad` reports at line 1:1 instead of the `if` condition
- `i32 + i32` reports at line 1:1 instead of the binary expression
- unary `-1` reports at line 1:1 instead of the unary expression

## Scope
- no diagnostic implementation yet
- no release claim widening

## Validation

### DIFF
--- a/docs/roadmap/full_readiness/gap_diagnostic_span_accuracy_scope.md
+++ b/docs/roadmap/full_readiness/gap_diagnostic_span_accuracy_scope.md
@@ -1,0 +1,100 @@
+# GAP: Diagnostic Span Accuracy
+
+Status: implementation-scope draft  
+Track: Semantic Full Readiness / diagnostics quality  
+Source trigger: Probe Pack v0
+
+## Problem
+
+Probe Pack v0 shows that several semantic errors report the location as line `1:1`, pointing at the first function declaration, even when the actual error is deeper in the file.
+
+Examples observed:
+
+- `if q` where `q: quad` reports line `1:1` instead of the `if` condition;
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
## Implementation notes — likely default/global span leak

Observed diagnostic pattern:

```text
Error [E0201]: <message> at line 1:1
  1 | fn <first_function_in_file> {
    | ^
```

This happens even when the real error is deeper in the file.

Likely causes to inspect:

- spans initialized to `(0,0)` / file start and not updated during AST traversal;
- semantic diagnostics emitted through a global/default source mark;
- expression-level errors using the enclosing function span instead of the offending expression span;
- lowered/type-check nodes losing their original source span.

Fast regression fixture:

```semantic
fn ok1() -> quad { return T; }

fn ok2() -> quad { return T; }

fn bad(q: quad) -> bool {
    if q {
        return true;
    }
    return false;
}
```

Expected after fix:

```text
caret points at `q` or the `if q` condition in `bad`, not at `fn ok1` line 1.
```

Acceptance refinement:

- at least one diagnostic fixture must place the first error in the third function;
- the snapshot must prove the caret does not collapse to the first function declaration;
- same test should be rerun after #370 line-ending normalization to ensure spans remain stable for LF and CRLF.

PR #1383, opened 2026-07-07
experiment(post-ui): separate fixture identity checks from reader shape checks

skulmakov-oss:
## Goal

Separate the fixture identity check from the shape check of the trust fields by emitting a specific trust_shape diagnostic.

This ensures the identity snapshot rejection does not obscure the reader's shape extraction abilities.


skulmakov-oss:
@codex review

Issue #1526, opened 2026-07-19
architecture: define Semantic Hub secure external tool runtime


skulmakov-oss:
# Goal

Define **Semantic Hub** as the protected external-tool execution boundary for Semantic.

Semantic Hub is not a folder of optional dependencies and not a general plugin marketplace. It is a controlled runtime layer that allows Semantic, Workbench, Semantic Studio, ALM-oriented systems, and future host applications to use external computational tools without granting those tools direct authority over Semantic Core, verifier, VM, capability, audit, project, or release truth.

Core formula:

```text
Semantic Hub
=
secure external tool boundary
+ isolated tool workers
+ typed message passing
+ capability admission
+ supervision
+ resource budgets
+ audit and provenance
```

Canonical rule:

```text
External tools provide bounded computational capability.
Semantic Hub governs execution and effects.
Semantic remains the owner of meaning, admission, and canonical truth.
```

---

# Why this is needed

Semantic is developing several future surfaces that will need external tools:

```text
- Workbench
- Semantic Studio
- Graph & Skill Toolchain
- ALM
- local vector indexes
- SMT/SAT solvers
- databases
- code analyzers
- compilers
- local models
- simulation engines
- import/export bridges
- industrial protocol adapters
```

Without a dedicated boundary, these integrations will naturally drift toward unsafe direct calls:

```text
Semantic component
  -> arbitrary crate / DLL / process / socket
  -> unrestricted state or filesystem access
  -> untracked result
```

That model would create several architectural failures:

```text
- direct mutable access to Semantic state;
- external code inside trusted address space without policy;
- host-effect bypass around prom-cap / prom-audit;
- hidden network or filesystem access;
- no resource accounting;
- no stable identity or provenance;
- no failure containment;
- no restart or quarantine policy;
- tool output accidentally promoted to Semantic truth;
- each integration inventing its own protocol and lifecycle.
```

Semantic Hub exists to prevent this fragmentation.

---

# Relationship to existing architecture issues

This issue is related to, but does not duplicate:

```text
#1370 — Semantic platform dependency graph and sequencing map
#1371 — Semantic threat model and untrusted project policy
#1373 — Semantic extension and plugin boundary
#1374 — Semantic artifact provenance and signing chain
#1369 — Workbench before Semantic Studio
```

The separation is:

```text
#1373 Extension boundary:
  defines what extensions may display, request, transform, and propose.

Semantic Hub:
  defines how external computational tools are admitted, isolated,
[...]

PR #1566, opened 2026-08-01
fix(hub): close Windows drive-prefix path gap, reconcile stale Hub v0 docs

skulmakov-oss:
## Summary

Small follow-up to the Semantic Hub v0 completion work that landed on `main` in `cb995efd` (bundled into #1558). That work was implemented across two sessions (one hit usage limits mid-way; an external tool continued it), and got committed/pushed directly to `main` inside an unrelated, already-merged PR instead of through its own reviewed PR. This PR is the honest correction pass on top of that:

- **Security fix**: `TurboVecAdapter`'s `checked_private_file_path` (`crates/semantic-hub-turbovec/src/storage.rs`) rejected traversal via `/`, `\`, `.`, `..`, but not a bare Windows drive-relative name (e.g. `C:evil.tvim.txn` — contains neither `/` nor `\`). `PathBuf::join` treats an argument carrying its own prefix as *replacing* the base path outright rather than appending to it, so such a name would fully escape the scoped storage root. Unreachable by any current caller (every caller builds the name from a charset-restricted `IndexName` plus a digit-only transaction id), but a real gap in a `pub(crate)` safety boundary meant for future private-file additions, on the OS this project targets. Fixed by requiring the name to parse as exactly one `Normal` path component; regression test added.
- **Docs**: `docs/architecture/semantic_hub_v0.md`'s fault-taxonomy and transaction/recovery sections still claimed `PersistenceFailed`/`RecoveryRequired` do not exist as `HubFault` variants and are only ever wrapped as `ToolDeclaredFailure`. That's stale relative to what actually shipped: `runtime.rs`'s `hub_fault_from_tool_error` maps them to their own distinct `HubFault` variants. Reconciled both sections (fault count corrected 25 → 28).
- **Harness report**: rewrote `.harness/reports/SEMANTIC-HUB-V0-COMPLETION.md` to accurately describe the final architecture and the irregular delivery path, instead of the stale draft that predated the later extensions.
- Removed `scratch_full.diff`, a stray raw diff dump accidentally committed to the repo root during that delivery.

Relates to / part of closing out #1526 (the substantive Hub v0 completion work — bounded session executor, write-ahead transaction/recovery, streaming audit, generic provenance envelope, extended fault taxonomy, `smc hub session` — is already live on `main` as of `cb995efd`; this PR closes the remaining correctness/doc gaps found during a full post-hoc review of that commit).

## Test plan
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy -p semantic-hub -p semantic-hub-turbovec -p smc-cli --all-targets --all-features -- -D warnings` clean
- [x] `cargo test -p semantic-hub -p semantic-hub-turbovec -p smc-cli` — 336 passed, 0 failed
- [x] `cargo test -p semantic_language --test hub_cli --release` — 34 passed, 0 failed
- [x] `cargo test -p semantic_language --test public_api_contracts --release` — 4 passed, 0 failed
- [x] `cargo check --workspace --all-targets --all-features` clean
- [x] New regression test covers the Windows drive-prefix case specifically


skulmakov-oss:
Both findings are real, fixed in 3763c18d: the drive-prefix rejection now checks the `<ascii-letter>:` pattern explicitly (platform-independent) instead of relying on `Path::components()`, which only special-cases that syntax on a Windows target -- the original fix would have wrongly accepted the traversal on Linux/macOS CI. Also caught and fixed two more leftover "21 fault codes"/"four additions" references in hub_api_v0.md that the first pass missed.

PR #1600, opened 2026-08-10
fix(runtime): charge application builtins against max_effect_calls (R2)

skulmakov-oss:
## Summary

R2 / SSF04-FIX-02 from #1598: application host builtins (`args_read`, `stdin_read_text`, `stdout_write`, `stderr_write`, `path_inspect`, `fs_read_text`, `fs_write_text`, `time_duration_ms`), dispatched through generic `Opcode::Call`, never consumed `ExecutionConfig.quotas.max_effect_calls`. Only the dedicated `Gate*/Pulse*/State*/Event*/Clock*` opcodes called `bump_effect_calls`. A loop of application-boundary calls could run unbounded until an unrelated quota (steps/calls) eventually intervened.

## Fix

`ApplicationVmHost` (the `VmHostBridge` used for application-boundary execution) now carries its own `RuntimeQuotas` snapshot and `effect_calls` counter, and charges exactly one effect-call unit per **admitted** application host operation — after the existing capability check (and, for writes, the observation check) and immediately before the actual host dispatch. This means:

- Quota exhaustion blocks the host effect itself (never partially executes it).
- Capability-denied attempts are **not** charged (decision documented below).
- Dedicated host opcodes (Gate/Pulse/State/Event/Clock) are untouched — they already charge via `bump_effect_calls(vm)` at their own call sites, unaffected by this change.

### Decision: denied attempts do not consume quota

Ordering is: capability check → (observation check) → quota charge → host dispatch. A capability-denied call therefore surfaces as `CapabilityDenied`, not `QuotaExceeded`, and leaves the quota untouched. Regression: `capability_denied_application_builtins_are_not_charged_against_the_effect_quota`.

## Tests

New `tests/ssf04_effect_quota.rs` (6 tests):
- `admitted_application_host_effects_up_to_the_quota_all_pass`
- `the_fifth_admitted_effect_would_be_denied_but_the_fourth_stays_under_quota_of_four`
- `a_single_application_host_effect_exceeds_a_zero_quota_deterministically`
- `capability_denied_application_builtins_are_not_charged_against_the_effect_quota`
- `every_application_builtin_kind_is_charged_against_the_effect_quota` (all 8 builtins, one call each)
- `the_eighth_distinct_builtin_kind_is_blocked_by_a_seven_call_quota`

Confirmed RED against pre-fix `main` (all four "must exceed quota" assertions failed with `()` instead of `QuotaExceeded`), GREEN after the fix.

## Evidence

- `cargo test --test ssf04_effect_quota --test ssf04_application_boundary`: 15/15 PASS
- `cargo test -p sm-vm`: 61/61 PASS
- `cargo test -p sm-front -p sm-ir -p sm-verify -p sm-emit`: PASS
- `cargo test -p semantic_language`: 718/719 PASS (1 pre-existing unrelated failure: `canonical_pack_all_sm_files_are_fmt_clean_and_lexically_sound` fails only under local Windows `core.autocrlf=true` checkouts — confirmed via `git diff` that the flagged file is byte-identical to `origin/main`; not present on Linux CI, out of scope for R2)
- `cargo test --test public_api_contracts --test trust_boundary_guards`: 9/9 PASS
- `cargo fmt --check`: clean
- `cargo clippy -p sm-vm -p semantic_language --tests -- -D warnings`: clean

Progresses #1598.

### DIFF
--- a/crates/sm-vm/src/semcode_vm.rs
+++ b/crates/sm-vm/src/semcode_vm.rs
@@ -777,6 +777,8 @@ pub fn run_verified_entry_semcode_with_application_host_and_capabilities_and_con
         host,
         capabilities,
         observed: false,
+        quotas: vm.config.quotas,
+        effect_calls: 0,
     };
     let mut observation = HelloObservationRuntime::discard();
     exec_loop(&mut vm, &mut bridge, &mut observation).map(|_| ())
@@ -1603,6 +1605,8 @@ struct ApplicationVmHost<'a, H: ApplicationHostAbi, C: CapabilityChecker> {
     host: &'a mut H,
     capabilities: &'a C,
     observed: bool,
+    quotas: RuntimeQuotas,
+    effect_calls: usize,
 }
 
 impl<'a, H: ApplicationHostAbi, C: CapabilityChecker> ApplicationVmHost<'a, H, C> {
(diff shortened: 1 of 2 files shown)


skulmakov-oss:
@codex review

Please review exact current HEAD `0eddcab3325cab41ff4272476acd38fa0644cd1a` for unresolved P1/P2 findings, with particular attention to: quota-vs-capability failure ordering in `ApplicationVmHost`, whether all eight application builtins are correctly charged exactly once, and whether dedicated host opcodes (Gate/Pulse/State/Event/Clock) remain unaffected.

PR #1605, opened 2026-08-10
fix(smc-cli): frame application stderr to keep it distinct from audit evidence (R7)

skulmakov-oss:
## Summary

R7 / SSF04-FIX-03 from #1598: application-controlled `stderr_write` payload and the CLI's trusted audit records (`ApplicationAuditRecord::render`, always starting with `schema=`) share process stderr. Application text was written raw via `io::stderr().write_all(text.as_bytes())`: payload without a trailing newline could concatenate onto a following audit line, and payload could embed a literal audit-schema-looking line that would be visually indistinguishable from genuine trusted evidence.

## Fix

Chosen design (option 2 from the issue's three acceptable bounded designs — "unambiguous structured/framed audit output"): added `write_framed_application_stderr`, which prefixes every physical line of application-controlled stderr output with `application| ` and always newline-terminates it, regardless of whether the application's own text was newline-terminated. Genuine audit lines never carry that prefix and always start with `schema=`, so the two evidence classes stay unambiguous even when application text embeds newlines or the literal audit schema string.

A true separate sink (option 1) was not chosen: it would add a new file descriptor/channel, which the checklist explicitly wants avoided ("avoids new ambient capability") when framing alone is sufficient.

Redaction/hashing in the audit record itself (`ApplicationAuditRecord::render`, `record_denied`) is untouched — this is a framing fix at the raw-byte-write site only, not an audit-format change. `stdout_write` is untouched (out of scope: the CLI has no trusted stdout evidence stream to protect).

## Tests

Four new unit tests in `crates/smc-cli/src/application_host.rs::tests`:
- `framed_stderr_prefixes_every_line_and_always_newline_terminates`
- `framed_stderr_cannot_impersonate_a_trusted_audit_line` — a forged payload containing the literal audit schema string still renders prefixed, never bare
- `framed_stderr_of_an_unterminated_payload_cannot_merge_onto_a_following_line`
- `framed_stderr_of_empty_text_writes_nothing`

Evidence:
- `cargo test -p smc-cli`: 136/136 PASS
- `cargo test -p semantic_language`: 718/719 PASS (same 1 pre-existing unrelated CRLF-checkout failure as R1-R6)
- `cargo fmt --check` / `cargo clippy -p smc-cli --tests -- -D warnings`: clean

Progresses #1598.

### DIFF
--- a/crates/smc-cli/src/application_host.rs
+++ b/crates/smc-cli/src/application_host.rs
@@ -8,6 +8,24 @@ pub(crate) const APPLICATION_AUDIT_SCHEMA: &str = "semantic.foundation.applicati
 const MAX_APPLICATION_TEXT_BYTES: usize = 16 * 1024 * 1024;
 const MAX_APPLICATION_PATH_BYTES: usize = 4096;
 
+/// Every physical stderr line written on behalf of the application carries this
+/// prefix. Trusted audit lines (rendered separately by the CLI, see
+/// `ApplicationAuditRecord::render`) always start with `schema=` and never with
+/// this prefix, so a payload that embeds literal audit-schema-looking text still
+/// renders as unambiguously application-controlled, and cannot concatenate onto
+/// a following trusted line since every framed line is newline-terminated here
+/// regardless of whether the application's own text was.
+const APPLICATION_STDERR_LINE_PREFIX: &str = "application| ";
+
+fn write_framed_application_stderr<W: Write>(writer: &mut W, text: &str) -> io::Result<()> {
+    for line in text.lines() {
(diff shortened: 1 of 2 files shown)


skulmakov-oss:
@codex review

Please review exact current HEAD `d5d37f85e41ad02f72426cc7e4ad59b950be11b4` for unresolved P1/P2 findings, with particular attention to: whether the per-line framing truly prevents impersonation and concatenation in all cases (embedded CR without LF, very large payloads, multiple stderr_write calls in sequence), whether the audit record's own schema string could itself ever appear unprefixed on stderr from a source other than ApplicationAuditRecord::render, and whether stdout_write genuinely has no analogous trusted-stream conflict.

PR #1606, opened 2026-08-10
fix(smc-cli): reject non-UTF-8 test source paths before sorting (R8)

skulmakov-oss:
## Summary

R8 / SSF05-FIX-01 from #1598: `discover_project_test_sources` sorted discovered `tests/**/*.sm` files by `path.strip_prefix(project_root).to_string_lossy()` — a lossy Unicode conversion. On Unix, two distinct non-UTF-8 filenames can lossy-convert to the same replacement-character string, so the sort key no longer reflects real file identity, leaving unspecified filesystem `read_dir` order to decide their relative execution order.

## Fix

Reject non-UTF-8 relative test paths deterministically at discovery, before sorting (`relative.to_str().is_none()` check right where the path is found), and sort by the real UTF-8 `str` (not a lossy conversion) once every discovered path is guaranteed valid.

I also investigated the checklist's "verify entry/root symlink protection across check, compile, verify, run through shared authority; add only missing coverage" item. Every caller of `resolve_project_root_check_entry` funnels through the shared `read_source_with_package_admission` → `admit_package_entry_module`, which already canonicalizes the entry and rejects it if the resolved path falls outside the package's `module_root` — this already correctly rejects a symlinked entry that escapes the project, with **no production code change needed**. I only added test coverage proving this holds across `check`/`compile`/`run`, satisfying "add only missing coverage" without introducing a new mechanism.

Documented the UTF-8 requirement in `docs/spec/project_model_v0.md`.

## Tests

New `tests/ssf05_test_path_utf8_and_entry_symlink.rs`:
- `valid_utf8_test_paths_execute_in_normalized_lexical_order` (cross-platform) — positive regression, ran and passed locally
- `non_utf8_test_source_paths_reject_deterministically` (`#[cfg(unix)]`) — two distinct invalid-UTF-8 filenames that both lossy-convert to the same string; rejected deterministically across two runs
- `project_entry_symlink_escaping_the_module_root_is_rejected_across_check_compile_run` (`#[cfg(unix)]`) — proves the existing shared admission authority already rejects a project entry symlinked outside the module root, for all three commands

Note: the two `#[cfg(unix)]` tests need real symlinks / non-UTF-8 `OsStr` paths, which this Windows dev environment can't exercise — they're gated to run on Linux CI, where this repo's `test-std` job runs.

Evidence:
- `cargo test -p smc-cli`: 132/132 PASS
- `cargo test --test ssf05_test_path_utf8_and_entry_symlink`: 1/1 PASS locally (the cross-platform one); full 3/3 expected on Linux CI
- `cargo test -p semantic_language`: 718/719 PASS (same 1 pre-existing unrelated CRLF-checkout failure as R1-R7)
- `cargo fmt --check` / `cargo clippy -p smc-cli -p semantic_language --tests -- -D warnings`: clean

Progresses #1598.

### DIFF
--- a/crates/smc-cli/src/app.rs
+++ b/crates/smc-cli/src/app.rs
@@ -2845,9 +2845,15 @@ fn discover_project_test_sources(project_root: &Path) -> Result<Vec<PathBuf>, St
                         path.display()
                     )
                 })?;
-                canonical.strip_prefix(root).map_err(|_| {
+                let relative = canonical.strip_prefix(root).map_err(|_| {
                     format!("test source '{}' escapes the project root", path.display())
                 })?;
+                if relative.to_str().is_none() {
+                    return Err(format!(
+                        "test source '{}' is not valid UTF-8; project test paths must be UTF-8",
+                        path.display()
+                    ));
+                }
                 tests.push(canonical);
             }
         }
@@ -2875,7 +2881,8 @@ fn discover_project_test_sources(project_root: &Path) -> Result<Vec<PathBuf>, St
     tests.sort_by_key(|path| {
         path.strip_prefix(project_root)
             .expect("discovered test remains inside project root")
-            .to_string_lossy()
+            .to_str()
+            .expect("test source path validated as UTF-8 during discovery")
             .replace('\\', "/")
     });
     Ok(tests)
--- a/crates/smc-cli/src/package_manifest.rs
+++ b/crates/smc-cli/src/package_manifest.rs
@@ -444,6 +444,7 @@ enum ProjectRootResolutionCode {
     SemanticTomlReadFailed,
     SemanticTomlManifest(SemanticTomlManifestErrorCode),
     SemanticTomlEntryMissing,
+    EntrySymlinkOrReparse,
     MissingProjectManifest,
 }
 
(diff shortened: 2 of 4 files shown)


skulmakov-oss:
@codex review

Please review exact current HEAD `43ae32d9f7b33404951090a97716faf917c04051` for unresolved P1/P2 findings, with particular attention to: whether rejecting non-UTF-8 test paths at discovery time (before sort) is the correct fix vs. e.g. needing to also validate at the tests/ root level, and whether the claim that admit_package_entry_module already rejects symlinked project entries escaping the module root (via canonicalize + strip_prefix) actually holds for all three of check/compile/run -- I could not run the added #[cfg(unix)] tests locally (Windows dev environment) so this needs real CI verification.

skulmakov-oss:
CI caught a real gap in the first submission: `project_entry_symlink_escaping_the_module_root_is_rejected_across_check_compile_run` failed for `check` — my assumption that `admit_package_entry_module`'s canonicalize+containment check already protected the project entry was wrong (it only applies when the symlink target happens to fall under some *other* manifest's scope; a symlink to a bare file with no ancestor manifest bypassed it entirely). Fixed in `resolve_project_root_check_entry_structured` with an explicit is-symlink-or-reparse check on the project entry, mirroring the policy `smc test`'s own source discovery already uses.

@codex review

Please review exact current HEAD `5a2d3754c2102457f8a8ded8f8af4058698cb89b` for unresolved P1/P2 findings, with particular attention to: whether `entry_is_symlink_or_reparse`/`require_entry_not_symlink_or_reparse` correctly cover both the `semantic.toml` entry path and the `Semantic.package` default `src/main.sm` path, whether this could reject any currently-valid project layout, and whether there's a similar unguarded gap elsewhere in `resolve_project_root_check_entry_structured`'s callers.

PR #1610, opened 2026-08-11
ssf/07: pin measured/fx cross-family arithmetic bounds

skulmakov-oss:
## Summary

SSF-07 (issue #1578) gap analysis verified that the source contract's excluded-arithmetic boundary for measured numeric forms is enforced in code (`typecheck.rs`'s narrow-slice-gap check plus ordinary type-mismatch rejection) but was pinned by only one test (`measured_fx_addition_still_reports_narrow_slice_gap`). This slice closes that coverage gap with no production code change.

Adds five sibling tests next to the existing precedent, covering:
- `measured_fx_subtraction_reports_narrow_slice_gap` — fx `-` hits the same narrow-slice-gap rejection as `+`
- `measured_arithmetic_rejects_mul_div_mod` — mul on a measured `f64` operand is rejected
- `measured_arithmetic_rejects_mismatched_units` — addition across different unit symbols (`f64[m]` + `f64[s]`) is rejected
- `measured_f64_addition_typechecks` — positive case: same-unit `f64[m]` addition typechecks
- `measured_i32_unary_minus_rejects` — unary minus on a measured `i32` is rejected

All six tests (five new + the precedent) pass against unmodified code — pure pinning, zero production diff outside the doc update below.

Also updates the `measured numeric forms` row in `semantic_stable_foundation_matrix.md` with the new evidence pointer.

## Test plan
- [x] `cargo test -p sm-front measured_ --lib` — 6/6 new+precedent tests pass
- [x] `cargo test -p sm-front --lib` — 459/459 pass
- [x] `cargo fmt --all` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `tools/7hell/run.ps1` — all 7 gates pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1609, opened 2026-08-11
SSF-07: freeze i32 arithmetic overflow contract

skulmakov-oss:
## Summary

Second implementation slice for SSF-07 (#1578, "Type and abstraction closure"), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608) (blanket/generic impl rejection).

`docs/spec/foundation_source_profile_v1.md` disclaimed i32's overflow policy as "decisions remain owned by SSF-07," but the behavior already exists and is fully deterministic:
- `AddI32`/`SubI32`/`MulI32` (`crates/sm-vm/src/semcode_vm.rs:1822-1850`) wrap silently via `wrapping_add`/`wrapping_sub`/`wrapping_mul` — no trap, no panic.
- `DivI32`/`ModI32` (`semcode_vm.rs:1852-1870`, via `i32_div_raw`/`i32_mod_raw` at `2893-2907`) trap deterministically on division/modulo by zero (`RuntimeTrap::DivisionByZero`) and on the `i32::MIN / -1` overflow edge case (`RuntimeTrap::ArithmeticOverflow`, via `checked_div`/`checked_rem`).

None of this had regression coverage anywhere in the repo — a future change could have silently flipped wrapping into panicking, or vice versa, with nothing to catch it.

## Regression coverage

Six new tests in `crates/sm-vm/src/semcode_vm.rs`, added next to the existing `vm_traps_on_fx_division_by_zero` precedent and following its exact pattern (`compile_program_to_semcode` + `run_semcode`, in-language `assert()` for the wrap cases, `RuntimeError::Trap(...)` matching for the trap cases):
- `vm_wraps_i32_addition_past_max`, `vm_wraps_i32_subtraction_past_min`, `vm_wraps_i32_multiplication_past_max` — pin the wrap-around behavior.
- `vm_traps_on_i32_division_min_by_negative_one`, `vm_traps_on_i32_division_by_zero`, `vm_traps_on_i32_modulo_by_zero` — pin the trap behavior.

All six pass immediately against unmodified code — **this is a documentation+test freeze, not a bugfix**. `i32::MIN` can't be written as a literal directly (`parse_i32_literal` parses unsigned digit text before unary negation applies), hence the `0 - 2147483647 - 1` construction.

[...]

PR #1608, opened 2026-08-11
SSF-07: reject blanket/generic impls, pin deterministic rejection for deferred trait forms

skulmakov-oss:
## Summary

First implementation slice for SSF-07 (#1578, "Type and abstraction closure"), the phase activated after SSF-06's repair program closed (#1598, PR #1597).

`docs/spec/foundation_source_profile_v1.md` explicitly lists trait objects, associated types, blanket impls, specialization, and default methods as "deterministically unsupported forms" that must fail before execution with a canonical diagnostic. Four of the five already did, incidentally, at parse time. **Blanket/generic impls did not**: `ImplDecl.type_params` is documented as "Empty in first-wave canonical form," but nothing enforced it — `impl<T> Trait for T` (or any impl declaring an unused type parameter) silently typechecked as an ordinary impl for a nominal type literally named by the parameter, contradicting the spec's own claim.

`validate_trait_coherence` now rejects any impl with non-empty `type_params` before the existing duplicate-impl check. No `.sm` fixture anywhere in the repo uses `impl<...>` syntax, so this closes a real, previously-undetected spec-violating gap with no legitimate-usage risk.

## Regression coverage

- `blanket_impl_with_type_params_is_rejected` (`typecheck.rs`) — the real fix; confirmed RED (silently accepted) before the fix, GREEN after.
- `default_method_in_trait_body_is_rejected`, `associated_type_in_trait_is_rejected`, `dyn_trait_object_syntax_is_rejected` (`parser.rs`) — pin the three forms that were already correctly, if incidentally, rejected at parse time.
- Specialization needs no new test: already covered by `trait_self_contract_allows_multiple_impl_targets`, and now fully moot since a true blanket impl can no longer be constructed to overlap with a concrete one.

## Test plan
- [x] `cargo test -p sm-front`: 454/454 PASS (450 baseline + 4 new).
- [x] `cargo test -p sm-ir -p sm-sema -p sm-verify -p sm-vm`: all PASS, no regressions.
- [x] `cargo fmt --check` / `cargo clippy -p sm-front --all-targets -- -D warnings`: clean.
- [x] `tools/7hell/run.ps1`: ALL 7 GATES PASSED.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1578, opened 2026-08-09
SSF-07 — Type and abstraction closure


skulmakov-oss:
Parent: #1569
Depends on: #1577

Priority: P7.

Status gate: **BLOCKED until SSF-06 closes.** This phase closes or explicitly bounds abstractions required by Stable Foundation; it is not a general type-system expansion program.

## Goal
Finish or explicitly bound the numeric, text, collection, closure, generic, trait, and pattern semantics required by the frozen Stable Foundation contour.

## Safe execution steps
1. Read SSF-00…06 outputs and build a gap table only for abstractions already selected by the Stable Foundation target.
2. Separate contract gaps from implementation bugs; do not add new abstractions simply because adjacent work is convenient.
3. **Numeric policy:** decide intended `u32` role, overflow behavior, and conversions across i32/u32/f64/fx/measured values before coding.
4. **Text policy:** freeze UTF-8 semantics, indexing/slicing behavior, escapes, formatting boundary, and whether char/byte carriers are included or deferred.
5. **Collections:** freeze Sequence/Map equality, ordering, mutability/persistence, and iteration guarantees.
6. **Closures:** freeze first-wave capture/invocation semantics; explicitly classify multi-argument, mutable capture, host transport, and async behavior.
7. **Generics/traits:** prove deterministic monomorphisation/coherence; classify associated types, trait objects, blanket impls, specialization, and defaults as included or deferred.
8. **Patterns:** align parser, typechecker, exhaustiveness, lowering, ownership interaction, and diagnostics.
9. For each included abstraction, add positive end-to-end tests plus deterministic negative cases before or with implementation changes.
10. Keep each implementation slice narrow; no global refactor unless a demonstrated invariant cannot be preserved otherwise.
11. Add canonical examples for included ordinary-program use cases and explicit rejection examples for deferred forms.
12. Run focused frontend/typechecker/lowering/verifier/VM tests, then repository-wide regression/qualification gates required by touched layers.

## Acceptance criteria
- [ ] Canonical ordinary programs require no undocumented workaround for included abstractions.
- [ ] Numeric overflow/conversion policy is explicit and tested.
- [ ] UTF-8/text and collection observability rules are explicit and tested.
- [ ] Included closures/generics/traits/patterns execute end-to-end.
- [ ] Deferred advanced features have explicit non-goals and deterministic diagnostics.
- [ ] No stable abstraction depends on unstable internal crate APIs.
[...]

PR #1611, opened 2026-08-12
ssf/07: reject impl methods not declared by their trait

skulmakov-oss:
## Summary

Fourth implementation slice for SSF-07 (#1578), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608) (blanket/generic impl rejection), [#1609](https://github.com/skulmakov-oss/Semantic/pull/1609) (i32 overflow contract), and [#1610](https://github.com/skulmakov-oss/Semantic/pull/1610) (measured/fx arithmetic bounds).

`validate_impl_conformance` (`crates/sm-front/src/typecheck.rs`) checks that every trait-declared method is present in an impl with matching arity/params/return type — but never checked the reverse direction: an impl could freely define extra methods not declared by the trait, and they typechecked silently.

This is safe to close now because impl methods have **no call syntax anywhere in the language** outside one hardcoded case: `build_fn_table` never registers `program.impls[].methods`, and UFCS `x.method(args)` desugars to a plain free-function call (`Expr::Call`), not trait dispatch. The only place any impl method is ever invoked is the single hardcoded Iterable `next()` for-loop desugaring. So an "extra" impl method was never callable in the first place — tightening this turns a silent no-op into a deterministic diagnostic, per SSF-07's "no silently unsupported forms" charter, without risk of breaking any real program.

## Change

Adds a symmetric loop to `validate_impl_conformance`, mirroring the existing "every trait method must be in impl.methods" loop: for each impl method, reject if the trait doesn't declare a method by that name. Follows the exact shape of the already-merged blanket-impl fix (PR #1608).

## Test coverage

Three new tests in `crates/sm-front/src/typecheck.rs`, next to the existing M9.2 Wave-3 coherence tests:
- `impl_extra_method_not_in_trait_is_rejected` — the real fix (RED before, GREEN after).
- `impl_duplicate_method_name_within_impl_is_rejected` — pins a previously-untested existing branch.
- `impl_of_unknown_trait_is_rejected` — pins another previously-untested existing branch.

Confirmed via TDD: before the fix, `impl_extra_method_not_in_trait_is_rejected` failed (extra method silently accepted) while the other two already passed (pure pins of existing code). After the one-loop fix, all 22 trait/impl-related tests in the file pass.

## Docs updated
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md` — "Broader static traits/protocols and impl coherence" row states the frozen bidirectional conformance contour.
- `docs/spec/foundation_source_profile_v1.md` — closures/static-dispatch section distinguishes the now-closed static conformance check from the still-deferred general dispatch/UFCS gap.

## Test plan
- [x] `cargo test -p sm-front --lib`: 476/476 PASS
- [x] `cargo fmt -p sm-front` / workspace fmt: clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (CI's exact invocation): clean
- [x] Slice verified via an exhaustive multi-agent gap-analysis workflow before implementation (confirmed low-risk, zero blast radius since impl methods have no dispatch path)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1612, opened 2026-08-12
ssf/07: pin multi-arg closure rejection as incidental, not explicit

skulmakov-oss:
## Summary

Fifth implementation slice for SSF-07 (#1578), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608), [#1609](https://github.com/skulmakov-oss/Semantic/pull/1609), [#1610](https://github.com/skulmakov-oss/Semantic/pull/1610), and [#1611](https://github.com/skulmakov-oss/Semantic/pull/1611).

`semantic_stable_foundation_matrix.md` claimed "captureful and multi-arg exclusions are explicit" for first-wave closures. The captureful half is genuinely explicit — a dedicated "capture-free only" diagnostic, test-pinned. The multi-arg half is not: the grammar has **no multi-parameter lambda production at all**. `(x, y) => x + y` never gets closure-aware handling — `starts_short_lambda_head()` requires `Ident` then `FatArrow`, so `(x, y)` (next token is `,`) falls straight into ordinary tuple-expression parsing and fails later with a generic, unrelated error. `Closure(f64, f64 -> f64)` as a type annotation fails only because `parse_type()` parses exactly one param type before requiring `->`. Both rejections are deterministic, but incidental — a side effect of the grammar shape, not a dedicated diagnostic — and neither had any regression test anywhere in the repo before this PR.

## Test coverage

Two new parser-level tests in `crates/sm-front/src/parser.rs`, next to the existing `rustlike_parser_rejects_captureful_short_lambda`:
- `rustlike_parser_rejects_multi_arg_short_lambda` — pins `(x, y) => x + y` failing with the exact generic message `"expected ';'"` (confirmed empirically, not guessed).
- `rustlike_parser_rejects_multi_arg_closure_type_annotation` — pins `Closure(f64, f64 -> f64)` failing with `"expected '->' between Closure parameter and return types"`.

Both pass immediately against unmodified code — pure pin, zero production change.

## Docs updated
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md` — the "Immutable first-wave closures" row now distinguishes the two exclusions instead of calling both "explicit," aligning with the already-accurate "not part of the current contract" / "Explicit Non-Goals" wording already used in `docs/spec/source_semantics.md` and `first_class_closures_full_scope.md`.

## Test plan
- [x] `cargo test -p sm-front --lib`: 478/478 PASS
- [x] `cargo fmt -p sm-front` / workspace fmt: clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (CI's exact invocation): clean
- [x] Slice verified via an exhaustive multi-agent gap-analysis workflow before implementation (confirmed low-risk)


skulmakov-oss:
@codex review

PR #1613, opened 2026-08-12
ssf/07: promote std.math sqrt/abs to Selected

skulmakov-oss:
## Summary

Sixth implementation slice for SSF-07 (#1578), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608), [#1609](https://github.com/skulmakov-oss/Semantic/pull/1609), [#1610](https://github.com/skulmakov-oss/Semantic/pull/1610), [#1611](https://github.com/skulmakov-oss/Semantic/pull/1611), and [#1612](https://github.com/skulmakov-oss/Semantic/pull/1612).

`std.math` sat entirely in Deferred status, but `sqrt` and `abs` are already Selected-quality today: both call straight into Rust's IEEE-754 `f64::sqrt`/`f64::abs` (`crates/sm-vm/src/semcode_vm.rs`), with no host effect and no VM quota charge. Unlike `sin`/`cos`/`tan`/`pow`, they don't depend on platform libm transcendental implementations.

`abs` is bit-exact for every possible `f64` input, including `NaN`/infinities. `sqrt` is IEEE-754 correctly-rounded and cross-platform bit-exact **for `x >= 0.0` only**; for `x < 0.0` it returns `NaN`, and while the *semantic* result (`sqrt(x) != sqrt(x)`) is guaranteed, IEEE 754 does not mandate a specific `NaN` bit pattern, so that edge is explicitly excluded from the frozen contract.

## Real bug caught by review, now fixed

`@codex review` caught that promoting these names without reserving them was unsafe: a user-defined `fn sqrt(x: f64) -> f64 { return 42.0; }` typechecked cleanly, but the VM's builtin dispatch intercepts `"sqrt"` by bare string match *before* function-frame dispatch, so the builtin silently ran instead — `assert(sqrt(4.0) == 42.0)` failed at runtime despite a clean `check`. This PR now includes a real code fix: `sqrt`/`abs` are reserved function names, rejected deterministically at typecheck time (mirroring the existing `APPLICATION_BUILTIN_NAMES` pattern for host-boundary names like `stdout_write`), verified end-to-end via the actual CLI.

## Change

Promotes `sqrt`/`abs` to Selected in `semantic.foundation.std/0.1`, scoped precisely (see above), with the name-collision hole closed.

## Changes
- `crates/sm-front/src/lib.rs`: new `STDLIB_MATH_BUILTIN_NAMES` reservation list + `build_fn_table` check; two new unit tests (the real fix, plus a pin for the previously-untested `APPLICATION_BUILTIN_NAMES` rejection path).
- `docs/spec/foundation_stdlib_v0.md`: `std.math` row moved to Selected; new `### std.math` subsection covering the `sqrt` domain scoping, the `NaN`-bit-pattern exclusion, and the name-reservation rule; "Deferred families" narrowed to the four transcendental builtins.
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md` and `standard_library_v0_evidence.md`: matching status/evidence updates.
- `tests/ssf_stdlib_v0_contract.rs`: updated the required-string pin that would otherwise break on the doc rewording; added `sqrt`/`abs` to the builtin-presence check.
- `examples/canonical/stdlib_v0_helpers/src/main.sm`: new `assert_math_helpers()`, called from `main()`, plus the matching guard-test addition.
- `tests/fixtures/pcc8_stdlib/positive_math_basic.sm` + `tests/pcc8_stdlib_acceptance.rs`: fixture pinning `sqrt`/`abs` correctness, same-input determinism, and the `NaN` domain boundary.

## Test plan
- [x] `cargo test -q --test ssf_stdlib_v0_contract`: 3/3 PASS
- [x] `cargo test -q --test canonical_examples`: 5/5 PASS
- [x] `cargo test -q --test pcc8_stdlib_acceptance`: 5/5 PASS
- [x] `cargo test -p sm-front --lib`: 480/480 PASS
- [x] `cargo fmt --all` / `cargo clippy --workspace --all-targets -- -D warnings` (CI's exact invocation): clean
- [x] Manually verified via `smc check`/`run` that the shadowing repro now fails deterministically at check-time instead of silently misbehaving at runtime
- [x] Slice went through 2 real review-caught issues (sqrt NaN-domain overclaim, then the name-shadowing bug) before landing clean — both fixed with evidence, not just narrowed claims


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
Pushing back on the last finding with primary-source evidence rather than implementing it — this is the one place in this review chain where the automated finding appears to be factually incorrect, not just imprecisely worded.

**Claim under review:** "the VM directly calls Rust's `f64::sqrt`, whose official documentation classifies its precision as non-deterministic across platforms, Rust versions, and potentially invocations."

**Rust's actual documentation for `f64::sqrt`** (https://doc.rust-lang.org/std/primitive.f64.html#method.sqrt), section "Precision":

> The result of this operation is guaranteed to be the rounded infinite-precision result. It is specified by IEEE 754 as `squareRoot` and guaranteed not to change.

This is the opposite of the "Unspecified precision" disclaimer that genuinely does apply to `exp`, `sin`, `cos`, `ln`, `powf`, etc. (verified in the same fetch — those methods carry: *"The precision of this function is non-deterministic. This means it varies by platform, Rust version, and can even differ within the same execution from one invocation to the next."*). `sqrt` is explicitly carved out as the one exception, because IEEE 754-2008 mandates `squareRoot` as a correctly-rounded operation — which is exactly the distinction this PR's documentation already draws between `sqrt`/`abs` (promoted) and `sin`/`cos`/`tan`/`pow` (kept Deferred, precisely because they carry that disclaimer and `sqrt` does not).

So the PR's claim — `sqrt(x)` for `x >= 0.0` is cross-platform bit-exact — is directly supported by Rust's own documented guarantee, not contradicted by it. I'm not making a change based on this finding; resolving the thread with this evidence attached. Happy to revisit if there's a counter-source, but I've checked the primary one.

skulmakov-oss:
@codex review

PR #1614, opened 2026-08-12
ssf/07: freeze std.text API surface

skulmakov-oss:
## Summary

Seventh (and per the original gap-analysis workflow, last) implementation slice for SSF-07 (#1578), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608), [#1609](https://github.com/skulmakov-oss/Semantic/pull/1609), [#1610](https://github.com/skulmakov-oss/Semantic/pull/1610), [#1611](https://github.com/skulmakov-oss/Semantic/pull/1611), [#1612](https://github.com/skulmakov-oss/Semantic/pull/1612), and [#1613](https://github.com/skulmakov-oss/Semantic/pull/1613).

The gap analysis found **no code-vs-spec drift** for `std.text` — every builtin's arity, admitted-type set, and error text already matches `docs/spec/foundation_stdlib_v0.md` exactly. The actual gap was evidence, not implementation: the existing drift guard (`tests/ssf_stdlib_v0_contract.rs`) only checks doc-text substrings, not runtime behavior, and `+` against `u32`/`Map`/`Record` operands had **zero test coverage anywhere in the repo** (only `i32`/`bool`/`quad` were tested).

## Change

Adds `tests/ssf07_text_family_freeze.rs` as the single canonical SSF-07 compatibility-freeze pointer for `std.text`, following the established `smc_cli::run(["check", path])` + error-substring pattern used by `tests/pcc_text_negative.rs` (an earlier draft of this slice incorrectly proposed calling `typecheck.rs`'s private `typecheck_source()` from `tests/` — that's unreachable from the workspace-root integration-test binary, caught and corrected before implementation).

Reuses existing `.sm` fixtures where they already exist and adds four new ones for the previously-untested operand types (`text_plus_u32.sm`, `text_plus_map.sm`, `text_plus_record.sm`, `text_plus_sequence.sm`) — **all verified empirically against the actual CLI before writing any assertion**: `i32`/`u32`/`bool`/`quad`/`Map`/`Record` all reject with the identical `"text concatenation currently admits only text + text operands"` message, but `Sequence` hits a *different, earlier* code path (the ordered-sequence operator gate fires before the text-specific check ever runs) with a distinct message — pinned as its own case rather than assumed to match the others.

**Zero production code change** — this documents and freezes already-correct behavior, same pattern as the i32-overflow, measured-arithmetic, and std.math slices.

## Explicitly out of scope
`std.seq`/`std.map` and the sequence-indexing scope boundary (whether `Sequence(T)` indexing belongs inside `std.seq`'s family boundary, per the row-67-vs-row-104 ambiguity the gap analysis flagged) remain open for a follow-up slice, per the analysis's own narrowest-first sequencing.

## Docs updated
- `docs/spec/foundation_stdlib_v0.md` — `std.text` section states the frozen operator boundary (both rejection paths) and points at the new test.
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md` — `std.text` row's evidence/routing updated; status label left unchanged (already correctly Gate-1-backed, unlike the std.math case in #1613).
- `docs/roadmap/stable_foundation/standard_library_v0_evidence.md` — `std.text` compatibility-baseline column points at the new consolidated test file.

## Test plan
- [x] `cargo test -q --test ssf07_text_family_freeze`: 3/3 PASS
- [x] `cargo test -q --test ssf_stdlib_v0_contract --test pcc_text_negative --test pcc8_stdlib_diagnostics`: all PASS (sanity check, confirms no regression to existing doc-substring pins)
- [x] `cargo fmt --all` / `cargo clippy --workspace --all-targets -- -D warnings` (CI's exact invocation): clean
- [x] Every new fixture's exact error text verified against the real `smc` CLI before writing the corresponding assertion, including the sequence case's distinct message


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1615, opened 2026-08-13
ssf/07: document the already-enforced match-scrutinee and exhaustiveness boundary

skulmakov-oss:
## Summary

Eighth implementation slice for SSF-07 (#1578), following [#1608](https://github.com/skulmakov-oss/Semantic/pull/1608)–[#1614](https://github.com/skulmakov-oss/Semantic/pull/1614).

**This started as a documentation-only PR and is no longer one.** The original assumption — that the `match`/pattern boundary needed zero new code and only needed its already-enforced behavior written down — was wrong. Round-by-round review (`@codex review`, ~20 rounds) progressively found that several of the behaviors being "documented" were themselves broken or inconsistent, culminating in three real phase-consistency defects in `crates/sm-front` and `crates/sm-ir`. This PR now fixes all three, in addition to the documentation work.

## What review found and what changed

1. **`u32` `match` was never actually executable.** Every `u32` match arm either compiled and then trapped at runtime (`runtime type mismatch: CmpEq/CmpNe operands must have same runtime type`, small values) or failed to lower at all for large values, because `crates/sm-ir/src/legacy_lowering.rs`'s pattern-literal lowering unconditionally emitted an `i32` carrier (`LoadI32`) regardless of the scrutinee's actual type. **Fix:** `expect_int_match_pattern` now takes the scrutinee type and emits the matching carrier (`LoadU32` for `u32`, with `u32::try_from` bound validation instead of `i32::try_from`). `u32` match is now Included and executable across its full domain (`0` through `u32::MAX` = `4294967295`), for both literal and range arm forms.
2. **The exclusive equal-bound range pattern (`5..5`) silently miscompiled.** Lowering checked only `start == end`, ignoring the parser-preserved `inclusive` flag, so the semantically-empty exclusive range `5..5` was lowered exactly like the inclusive `5..=5` and incorrectly matched the value `5`. **Fix:** the equal-bounds fast path now additionally requires `range.inclusive`; an exclusive equal-bound range falls through to the same deterministic "range lowering is not yet implemented" rejection every other unsupported range form already used. `5..5` can no longer compile into literal `5` — it is now a compile-time rejection, not silent wrong behavior.
3. **Or-pattern match arms (`A | B`) were phase-inconsistent.** Typecheck and exhaustiveness accepted them for every admitted scrutinee family, then lowering failed — with a diagnostic that varied by family and by whether a wildcard arm was present, in one case actively confusing ("quad match pattern requires quad scrutinee" for an *enum* scrutinee). **Fix:** or-patterns are now rejected deterministically at typecheck time (`build_and_apply_match_plan` in `crates/sm-front/src/typecheck.rs`), with one diagnostic worded identically across all six admitted scrutinee families (`quad`, `i32`, `u32`, enum/ADT, `Option(T)`, `Result(T, E)`) and regardless of wildcard presence. `if let` is unaffected — it uses a separate entry point.

All three fixes are backward-compatible with everything Foundation Source 1.1 actually promised: no previously-Included form stopped working. Two are purely additive (`u32` match went from always-broken to working); the third converts a previously undocumented, buggy edge case into either correct behavior (`5..5`) or a cleaner, earlier, uniform rejection (or-patterns).

## Architecture decisions

- **`u32` match: Included, real implementation.** The VM already had full `u32` comparison support (`LoadU32`/`CmpEq` already worked correctly for non-match `u32` equality); the gap was purely in match-lowering always choosing the `i32` carrier. Implementing the correct carrier selection was a small, bounded, two-call-site fix — not a redesign — so "fake through `i32`" was never necessary and full inclusion was the smallest correct architecture.
- **Exclusive-range (`5..5`): reject, don't implement empty-range semantics.** Genuine multi-value range lowering doesn't exist yet at all (a separate, pre-existing M9.4 Wave 1 boundary); implementing true "matches nothing" semantics for one degenerate case ahead of general range support would be a bigger, less coherent change than routing it to the same pre-existing rejection path every other unsupported range form already uses.
- **Or-patterns: reject at typecheck, don't implement lowering.** Real or-pattern lowering across all six families (expanding each alternative into its own check, or its own tag-check-and-bind for the sum families) is architecturally bounded but touches ~7-8 lowering call sites plus exhaustiveness reasoning — a materially larger, riskier change than the u32/range fixes. Or-patterns were never part of the Included executable surface under 1.0 or 1.1 (this PR's own earlier rounds had already established that), so rejecting them cleanly and uniformly, rather than implementing lowering, is the smallest architecture that removes the phase inconsistency without expanding scope beyond what this slice was chartered to fix.

## Qualification matrix

A new granular, per-form, per-phase supplement to the existing summary evidence row:

- `docs/roadmap/stable_foundation/ssf07_pattern_qualification_matrix.md`

covers every `i32`/`u32` range form (literal, inclusive/exclusive singleton, multi-value, zero/max/oversized/negative boundaries, suffixed bounds) and every or-pattern family/wildcard combination, each with its exact phase-by-phase (parse/typecheck/exhaustiveness/lower/verify/run) result and pinning fixture — not just a frontend-only test standing in for full qualification.

## Regression evidence

New fixtures under `examples/qualification/match_surface/`, wired into `tests/match_surface_qualification.rs`:

- **Positive (full pipeline, `check`→`compile`→`verify`→`run-smc`):** `positive_u32_match_full_domain` (literal + range forms across `0`, ordinary, `i32::MAX`, `i32::MAX+1`, `u32::MAX`), `positive_i32_plain_literal_match`.
- **Negative — typecheck-phase (`assert_check_rejects`):** all 9 or-pattern combinations (`quad`, `i32`, `u32`, enum/Option/Result × with/without wildcard), plus the pre-existing suffixed-bound and new negative-bound range rejections.
- **Negative — lowering-phase (`assert_compile_rejects`):** exclusive singleton, multi-value, exclusive multi-value, and oversized-singleton ranges, for both `i32` and `u32`.

The two fixtures that used to pin the *old, wrong* behavior (`defect_u32_literal_match_runtime_trap`, `defect_u32_range_match_runtime_trap`, `defect_i32_exclusive_singleton_range_miscompilation`) were deleted — they were pinning bugs that no longer exist.

Also updated 4 pre-existing `sm-front` unit tests (`crates/sm-front/src/typecheck.rs`) that asserted or-pattern match arms typecheck successfully — they now assert the deterministic rejection instead, since that's a deliberate behavior change, not a regression.

## Foundation Source version transition

`semantic.foundation.source/1.1` → `1.2`. `1.1` was a pure documentation clarification; `1.2` is not — it contains the three real fixes above, all backward-compatible per the version policy. Synchronized every current-facing authority: `README.md`, `docs/spec/foundation_source_profile_v1.md`, `syntax.md`, `types.md`, `source_semantics.md`, `source_style.md`, `diagnostics.md`, `foundation_stdlib_v0.md`, `controlled_application_boundary_v0.md`, `logos.md`, `stable_public_language_contract.md`, `rustlike_logos_coherence_decision.md`, `standard_library_v0_evidence.md`, and the `stdlib_v0_helpers` canonical example README. Extended `tests/ssf_language_contract_drift.rs` with a new `ssf_07_pattern_surface_authorities_stay_in_sync` guard so a future version bump or match-allowlist change that misses one of these authorities fails CI instead of surfacing as another review round.

## Commands run

```
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test -p sm-front --lib
cargo test -p sm-ir --lib
cargo test -p sm-vm --lib
cargo test --test match_surface_qualification
cargo test --test ssf_language_contract_drift
cargo test --test frontend_boundaries
cargo test --test trust_boundary_guards
cargo test --test public_api_contracts
cargo test --test canonical_examples
cargo test --workspace
tools/7hell/run.ps1
```

Full `cargo test --workspace` result: all green except the same 3 pre-existing local-checkout artifacts in `canonical_source_style.rs` (CRLF vs LF, caused by `core.autocrlf=true` on this Windows machine, unrelated to this PR — those files are untouched per `git diff`).

## Remaining non-goals (deliberate, not defects)

- General multi-value range-pattern lowering (`1..=5`, `1..5`) remains a known, pre-existing M9.4 Wave 1 boundary, unchanged by this PR — still deterministically rejected, not silently wrong.
- Or-pattern *lowering* (as opposed to rejection) remains unimplemented; it is now a clean, deliberate, uniformly-diagnosed typecheck-time boundary rather than a phase-inconsistent one.
- `u32` general arithmetic (`+`, `-`, `*`, `/`, `%`), conversions, and overflow policy remain deferred to a later SSF-07 slice — this PR only fixes `u32` `match`, not `u32` arithmetic.

## Why this is safe to freeze

Every form in the qualification matrix now ends in exactly one of the two valid states: a fully-qualified Included executable, or a deterministic pre-artifact rejection with a stable, tested diagnostic. No form is left where one compiler phase accepts it and a later phase fails unexpectedly, executes different semantics, or produces a runnable-but-wrong artifact.

## Test plan
- [x] `cargo fmt --all -- --check`: PASS
- [x] `cargo clippy --workspace --all-targets -- -D warnings`: PASS
- [x] `cargo test -p sm-front --lib`: 480 passed
- [x] `cargo test -p sm-ir --lib`: 99 passed
- [x] `cargo test -p sm-vm --lib`: 69 passed
- [x] `cargo test --test match_surface_qualification`: PASS
- [x] `cargo test --test ssf_language_contract_drift`: PASS (both the pre-existing and new guard)
- [x] `cargo test --workspace`: PASS (except the 3 known-unrelated local CRLF artifacts noted above)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1616, opened 2026-08-14
fix(cache): invalidate semantic results across compiler revisions

skulmakov-oss:
## Finding

While qualifying SSF-07 PR #1615 (which changes real `sm-front`/`sm-ir` semantics), a real cache correctness defect surfaced: after rebuilding `smc-cli` with changed typecheck/lowering behavior, `smc check` could still return an old cached result until `.semantic-cache` was deleted or `--no-cache` was used.

## Violated invariant

Cached semantic results are valid only for the compiler semantics that produced them. Specifically, `cache HIT` must never survive a compiler generation change while `smc check --no-cache` would produce a different result for the same source. Separately (added in the round covering identity *construction*): if the compiler cannot construct a complete, lossless, deterministic identity for the build that produced a cache entry, it must not guess, silently omit input, or normalize two distinct identities into one — uncertainty during identity construction must fail the build, not narrow the identity. Runtime cache *lookup* behavior stays fail-open-to-MISS throughout (`CacheReason::{ToolchainMismatch, FeatureMismatch, FingerprintMismatch, ...}` → fresh check); that is correct and unchanged.

## Root cause

Confirmed from code, not assumed. `current_toolchain_hash()`'s default value was `fnv1a64(format!("smc-cli:{}", CARGO_PKG_VERSION))` — a pure function of the crate's package version alone. `CARGO_PKG_VERSION` is `0.1.0` and does not change on ordinary commits. `load_cache_entry_ex` (the `smc check` semantic-pass cache) and `load_blob_pack_ex` (the AST/IR/SMC dump caches) both gate correctness on this single `toolchain_hash` field — so a rebuild that changes typecheck/lowering behavior without a version bump produced the *identical* hash as the build before it, and an old cache entry stayed a HIT.

## Architecture evolution

This PR went through five review-driven rounds, each closing a real gap the previous round left open:

1. **Package-version-only identity → source-content generation.** Introduced `crates/smc-cli/build.rs` computing `SM_COMPILER_SOURCE_HASH` from a hand-listed `SEMANTIC_CRATES` set's `.rs` files, folded into `current_toolchain_hash()`/`current_feature_hash()` via a new pure `toolchain_identity_tag` helper.
2. **Full local dependency closure + repository-relative paths + Cargo feature identity.** Round-1 review found the hand-listed crate set was incomplete (missed real dependencies like `sm-profile`, `ton618-core`), the hashed paths were absolute (breaking checkout-path independence), and Cargo feature flags (`profile-rust`/`profile-logos`/`debug-symbols`) weren't part of the identity at all. Replaced the hand list with `discover_local_dependency_closure` (transitive scan of `smc-cli`'s own `Cargo.toml` dependency graph), rooted all hashed paths at `crates/`, and added `SM_ENABLED_FEATURES` (sorted `CARGO_FEATURE_*`) into `current_feature_hash()`.
3. **Manifest identity.** Round-2 review found a manifest-only change (e.g. removing a hardcoded feature from one crate's path-dependency on another) could change compiled behavior without touching any `.rs` file or `smc-cli`'s own features. Extended hashing to include each closure member's `Cargo.toml` content, not just its `src/`.
4. **Fail-closed identity construction + Cargo.lock-based closure discovery.** A follow-up architecture audit found the identity-*construction* code itself wasn't fail-closed (see below), and that the hand-written `Cargo.toml` text scanner for dependency discovery wasn't provably correct (a real, if not yet triggered, gap: `crates/prom-ui-demo/Cargo.toml`'s `path = "../../experiments/ui-shell-kit"` would have been mis-extracted by the old single-level `../<name>` scan). Replaced the scanner with `Cargo.lock`-derived closure discovery, and closed every silent-fallback path in identity construction. **A design mistake in this round's first draft — hashing `Cargo.lock`'s full content to capture external-dependency drift — was caught empirically, not by review**: `Cargo.lock` is listed in this repository's own `.gitignore` and is independently resolved per checkout, so two fresh `git worktree`s of the *identical* commit resolved different external-dependency patch versions immediately. Hashing it would have made the same committed source produce different identities across checkouts. Corrected to use `Cargo.lock` only for local-crate closure discovery (deterministic, since it derives from tracked `Cargo.toml` files), never as hashed content.
5. **Watch-before-exists + delimited hashing.** Round-4 review on that commit found two more gaps in the round's own new code: `hash_optional_file` only registered `cargo:rerun-if-changed` for a per-crate `build.rs` that already existed, so a crate that later gained one wouldn't trigger a rebuild of the identity; and `mix_labeled` concatenated each file's label and content with no delimiter, so a label/content boundary shift (`label "ab"+content "cd"` vs `label "abc"+content "d"`) produced the identical hash. Fixed by printing the watch directive unconditionally before the existence check, and by length-prefixing every label/content chunk before hashing it.

## Current identity (final)

```
toolchain_hash = fnv1a64("smc-cli:" + CARGO_PKG_VERSION + ":" + SM_COMPILER_SOURCE_HASH)
feature_hash   = fnv1a64("debug_assertions=...;target_pointer_width=...;features=" + SM_ENABLED_FEATURES)
```

`SM_COMPILER_SOURCE_HASH` (`build.rs`) = FNV-1a over, for every crate in `smc-cli`'s local semantic dependency closure (discovered from `Cargo.lock`, currently 20 crates: `prom-abi`, `prom-audit`, `prom-cap`, `prom-refs`, `prom-ui`, `prom-ui-runtime`, `semantic-core-quad`, `semantic-hub`, `semantic-hub-turbovec`, `sm-emit`, `sm-format`, `sm-front`, `sm-ir`, `sm-profile`, `sm-runtime-core`, `sm-sema`, `sm-verify`, `sm-vm`, `smc-cli`, `ton618-core`): its `Cargo.toml` content, its optional `build.rs` content (defensive; none exists today besides `smc-cli`'s own), and every `.rs` file under `src/`, each labeled by its path relative to `crates/` (never the absolute checkout path). `SM_ENABLED_FEATURES` = sorted `CARGO_FEATURE_*` env vars Cargo sets for this build.

The `SM_TOOLCHAIN_HASH`/`SM_FEATURE_HASH`/`SM_CACHE_SCHEMA`/`SM_CAPS_HASH` env overrides are untouched and still take priority for anyone who sets them. A *malformed* override value is deliberately left falling through to the computed default rather than erroring — documented in `app.rs`: the default is itself a complete, fail-closed identity, so this can never cause a false HIT the way an incomplete construction-time fallback could.

**Why content-hash, not git-commit-SHA** (the task's originally-suggested option): the actual incident that motivated this PR was a *local rebuild from uncommitted source edits* during #1615's qualification. A commit SHA doesn't change until you commit, so it would not have caught that. A content fingerprint catches both committed and uncommitted changes, needs no git binary, and is deterministic/stable across rebuilds of unchanged source.

## Fail-closed identity construction (round 4)

Fail-open paths found and fixed in `build.rs`:
- Unreadable manifest silently became "no dependencies" → the whole `Cargo.toml` text-scanning discovery approach is gone, replaced by `Cargo.lock` parsing.
- Unreadable expected `src/` directory silently hashed nothing → `hash_dir` now returns `Result` and fails the build.
- Directory-enumeration errors were silently dropped (`.filter_map(|e| e.ok())`) → entries collected through `Result`, failing on the first error.
- File-read failures were silently ignored while the file was still watched for changes → `hash_file` now fails the build.
- `strip_prefix(...).unwrap_or(path)` silently fell back to the *absolute* path if a path was ever outside the identity root → now a hard error.
- `to_string_lossy()` could silently collapse two distinct non-UTF-8 paths onto the same label → replaced with `to_str()` + hard error.
- Backslash-to-slash folding was unconditional, wrong on Unix (`\` is an ordinary filename byte there) → now `cfg!(windows)`-gated.
- `Cargo.lock` dependency names with no matching `[[package]]` entry anywhere now fail the build instead of being silently treated as absent from the closure.

## Repair

- `crates/smc-cli/build.rs` — computes `SM_COMPILER_SOURCE_HASH`/`SM_ENABLED_FEATURES`, fail-closed throughout, `Cargo.lock`-based closure discovery.
- `crates/smc-cli/src/app.rs` — `current_toolchain_hash()`/`current_feature_hash()`/`current_cache_schema_version()`/`current_caps_hash()` fold in the build-time identity via `toolchain_identity_tag`; `DL-XXX` placeholder comments replaced with the real `DL-024`/`DL-025` references; malformed-override handling documented.
- `crates/smc-cli/tests/build_identity_fail_closed.rs` (new) — 13 tests exercising `build.rs`'s own logic directly (pulled in via `#[path]`, since a build script has no test target of its own `cargo test` runs).

## Regression evidence

Reproduced empirically before writing any test: cached a `smc check` result, edited `crates/sm-front/src/typecheck.rs` (**uncommitted**), rebuilt, reran `check` on the same source without touching `.semantic-cache` → printed `smc check passed` (fresh), not `smc check passed (cached)`. Reverted before writing the test suite. Checkout-path independence (round 4) was verified the same way: built the same uncommitted source from two different absolute `git worktree` paths, confirmed identical `SM_COMPILER_SOURCE_HASH` only after the `Cargo.lock`-content-hashing correction above.

Added tests (`crates/smc-cli/src/app.rs`, `mod tests`):
- `semantic_cache_toolchain_hash_differs_across_compiler_generations` — direct test of `toolchain_identity_tag`'s core claim.
- `semantic_cache_hits_for_same_compiler_generation` — same generation + same source fingerprint → HIT.
- `semantic_cache_misses_when_compiler_generation_changes` — the key regression: `Miss(ToolchainMismatch)` across generations.
- `semantic_cache_still_misses_when_source_fingerprint_changes` — same generation, different fingerprint → `Miss(FingerprintMismatch)`, unregressed.
- `semantic_check_command_discards_cache_entry_from_another_compiler_generation` — exercises `cmd_check` end to end.
- `build_script_identity_env_vars_are_wired_into_the_runtime_formulas` — locks in that `current_toolchain_hash()`/`current_feature_hash()` actually consume what `build.rs` emits.

Added `crates/smc-cli/tests/build_identity_fail_closed.rs` (13 tests): deterministic/lossless identity for identical input, checkout-path independence (real filesystem trees under two different roots), Unix backslash-vs-slash non-collision, non-UTF-8 path rejection, file/directory/`Cargo.lock` read-failure → error (not `Vec::new()`/empty hash), `Cargo.lock` trailing-comma parsing, dependency-closure completeness including a dangling-reference hard failure and a missing-entry-crate hard failure, and (round 5) a direct reproduction of the label/content boundary-shift collision proving it no longer occurs.

Round 5 (the two remaining review findings on round 4's own new code) verified empirically before trusting the fix: rebuilt and inspected cargo's own captured build-script stdout (`target/debug/build/smc-cli-*/output`), confirming `cargo:rerun-if-changed` is now present for every closure crate's `build.rs` path even when absent (`sm-format`, `sm-front`, `sm-ir`, `prom-abi`, etc.), not just `smc-cli`'s own.

`VersionMismatch`/`ChecksumMismatch` coverage is unmodified and unregressed (`cache_version_mismatch_is_ignored`, `cache_checksum_mismatch_is_ignored`).

Env-var mutation was deliberately avoided in the app.rs tests (a real parallel-test race risk, since `std::env::set_var` is process-global) in favor of the codebase's own established direct-`PackHeader`-construction pattern.

## Non-goals

Explicitly deferred, not touched:
- `cache_file_for_root`'s non-UTF-8/lossy canonical-path identity (separate, already-known defect class).
- `incremental.rs`'s backslash/path-normalization cache-identity issues (separate, already-known defect class).
- PR #1615's pattern-matching semantics — untouched.
- `CACHE_SCHEMA_VERSION` was **not** bumped — this repair is a structural fix to what participates in toolchain identity going forward, not a one-time cache-format change.
- No new pinning policy for external `Cargo.lock` dependencies — that file is gitignored in this repository by pre-existing convention; this PR does not change that.

## Qualification (round 5, final)

```
cargo fmt --all -- --check                                        PASS
cargo clippy -p smc-cli --all-targets -- -D warnings               PASS
cargo test -p smc-cli                                              PASS (165 passed: 152 lib + 13 new)
cargo test --test frontend_boundaries                              PASS (2 passed)
cargo test --test trust_boundary_guards                            PASS (5 passed)
cargo test --test public_api_contracts                             PASS (4 passed)
cargo test --workspace --no-fail-fast                              PASS (only the 3 pre-existing
                                                                     local-checkout CRLF failures in
                                                                     canonical_source_style.rs remain,
                                                                     unrelated to this branch -
                                                                     core.autocrlf=true on this Windows
                                                                     checkout, untouched files, same 3
                                                                     failures on main)
tools/7hell/run.ps1                                                 PASS (ALL 7 GATES PASSED)
```

CI: all checks green on the final HEAD. `@codex review` on the exact final HEAD (`7e50d2a07b`): "Didn't find any major issues." All 6 review threads across all 5 rounds resolved. **0 unresolved P1, 0 unresolved P2.**

## Decision log

- DL-024 (issue #1598): the original compiler-generation cache-invalidation finding, root cause, and decision.
- DL-025 (issue #1598): the round-4 fail-closed identity-construction finding, the `Cargo.lock`-content-hashing correction, and the malformed-override decision.

### DIFF
--- a/crates/smc-cli/build.rs
+++ b/crates/smc-cli/build.rs
@@ -6,7 +6,11 @@
 //!   verification, execution), discovered as `smc-cli`'s full transitive
 //!   local-path dependency closure rather than a hand-maintained list, so a
 //!   newly-added semantic dependency is picked up automatically instead of
-//!   silently falling outside the fingerprint.
+//!   silently falling outside the fingerprint. Includes each closure
+//!   member's `Cargo.toml` content, not just its `.rs` files: a manifest
+//!   can change compiled behavior (e.g. a crate hardcoding
+//!   `features = [...]` on one of its own path-dependencies) without
+//!   touching a single source file.
 //! - `SM_ENABLED_FEATURES` — the Cargo feature flags this specific build of
 //!   `smc-cli` was compiled with (e.g. `profile-rust`, `debug-symbols`),
 //!   since those change lowering/verification behavior independently of
@@ -46,11 +50,9 @@ fn main() {
 
     let mut hash: u64 = 0xcbf2_9ce4_8422_2325;
     for crate_name in &closure {
-        hash_dir(
-            &crates_dir,
-            &crates_dir.join(crate_name).join("src"),
-            &mut hash,
-        );
+        let crate_dir = crates_dir.join(crate_name);
+        hash_file(&crates_dir, &crate_dir.join("Cargo.toml"), &mut hash);
+        hash_dir(&crates_dir, &crate_dir.join("src"), &mut hash);
     }
     println!("cargo:rustc-env=SM_COMPILER_SOURCE_HASH={:016x}", hash);
 
@@ -70,7 +72,6 @@ fn discover_local_dependency_closure(crates_dir: &Path, entry_crate: &str) -> BT
             continue;
         }
         let manifest_path = crates_dir.join(&name).join("Cargo.toml");
-        println!("cargo:rerun-if-changed={}", manifest_path.display());
         for dep in local_path_dependencies(&manifest_path) {
             stack.push(dep);
         }
@@ -118,17 +119,25 @@ fn hash_dir(crates_dir: &Path, dir: &Path, hash: &mut u64) {
         if path.is_dir() {
             hash_dir(crates_dir, &path, hash);
         } else if path.extension().is_some_and(|ext| ext == "rs") {
-            if let Ok(bytes) = fs::read(&path) {
-                let relative = path.strip_prefix(crates_dir).unwrap_or(&path);
-                fnv1a64_update(
-                    hash,
-                    relative.to_string_lossy().replace('\\', "/").as_bytes(),
-                );
-                fnv1a64_update(hash, &bytes);
-            }
-            println!("cargo:rerun-if-changed={}", path.display());
+            hash_file(crates_dir, &path, hash);
         }
     }
+}
+
+/// Hashes one file's crates-relative path and content, and registers it
+/// with `cargo:rerun-if-changed`. A missing file is skipped rather than
+/// treated as an error, matching `hash_dir`'s tolerance of a stripped-down
+/// checkout.
+fn hash_file(crates_dir: &Path, path: &Path, hash: &mut u64) {
+    if let Ok(bytes) = fs::read(path) {
+        let relative = path.strip_prefix(crates_dir).unwrap_or(path);
+        fnv1a64_update(
+            hash,
+            relative.to_string_lossy().replace('\\', "/").as_bytes(),
+        );
+        fnv1a64_update(hash, &bytes);
+    }
+    println!("cargo:rerun-if-changed={}", path.display());
 }
 
 fn fnv1a64_update(hash: &mut u64, bytes: &[u8]) {
--- a/crates/smc-cli/build.rs
+++ b/crates/smc-cli/build.rs
@@ -1,0 +1,80 @@
+//! Computes a deterministic content fingerprint over the source of every
+//! crate that determines `smc check`'s semantic behavior (parse, typecheck,
+//! lowering, verification, execution) and exposes it to the crate as
+//! `SM_COMPILER_SOURCE_HASH` via `env!()`.
+//!
+//! This exists so the on-disk semantic cache (see `current_toolchain_hash`
+//! in `src/app.rs`) can tell two compiler builds apart even when neither
+//! `Cargo.toml`'s package version nor any environment variable changed. A
+//! `smc-cli` rebuilt from edited (including uncommitted) source in these
+//! crates gets a different hash than the build that produced an existing
+//! cache entry, so that entry is correctly treated as stale.
+//!
+//! Deliberately content-based rather than git-commit-based: the defect this
+//! repairs was observed after a local rebuild from *uncommitted* source
+//! edits, where the git HEAD commit does not change at all.
+
+use std::fs;
+use std::path::Path;
+
+/// Crates whose source content determines what a semantic-cache entry means.
+/// Kept in sync manually; adding a new crate to the check/compile/verify/run
+/// pipeline should add it here too.
+const SEMANTIC_CRATES: &[&str] = &[
+    "sm-front",
+    "sm-ir",
+    "sm-sema",
+    "sm-emit",
+    "sm-verify",
+    "sm-vm",
+    "smc-cli",
+];
+
+fn main() {
+    let manifest_dir =
+        std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR set by cargo");
+    let crates_dir = Path::new(&manifest_dir)
+        .parent()
+        .expect("smc-cli manifest dir has a crates/ parent");
+
+    let mut hash: u64 = 0xcbf2_9ce4_8422_2325;
+    for crate_name in SEMANTIC_CRATES {
+        let src_dir = crates_dir.join(crate_name).join("src");
+        hash_dir(&src_dir, &mut hash);
+    }
+
+    println!("cargo:rustc-env=SM_COMPILER_SOURCE_HASH={:016x}", hash);
+}
+
+/// Recursively hashes every `.rs` file's path (relative ordering matters,
+/// hence the sort) and content, and registers each with
+/// `cargo:rerun-if-changed` so editing any of them re-runs this script.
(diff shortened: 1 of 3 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1802, opened 2026-08-17
fix(sm-format): make untrusted offset arithmetic checked

### LINKED ISSUE #1736: FA-05-006: unchecked decoder offset arithmetic can panic on malformed SemCode on 32-bit targets
Umbrella: #1617
Module: 05 — `sm-format`
Severity: **P2**
Classification: **FAIL-OPEN · QUALIFICATION-GAP · FALSE-READY**

## Trigger

`decode_semcode_envelope()` reads the attacker-controlled function `code_len` as `u32`, casts it to `usize`, and performs unchecked arithmetic:

```rust
let code_len = read_u32_le(bytes, &mut cursor)? as usize;
if cursor + code_len > bytes.len() { ... }
let code_slice = &bytes[cursor..cursor + code_len];
```

On a 32-bit target, a tiny malformed artifact can choose a near-`u32::MAX` `code_len` so `cursor + code_len` overflows `usize`.

The same low-level format helpers (`read_u16_le`, `read_u32_le`, `read_f64_le`, `read_utf8`) also use unchecked `*i + N` / `*i + len` arithmetic on their public cursor argument.

## Claim / confidence source

The audit governing invariant and SemCode trust-boundary contract require malformed/truncated/unsupported artifacts to produce deterministic rejection, not process panic or build-profile-dependent behavior. `sm-format` is presented as the neutral format/decode authority used before verifier admission.

## Actual guarantee

On 64-bit canonical decoding, the `u32` field is too small to overflow ordinary `usize` arithmetic, so this specific artifact path is not a current x86-64 exploit.

On 32-bit:

- debug builds can panic at the overflowing addition;
- release builds can wrap the end offset, potentially pass the `> bytes.len()` test, and then panic on an invalid slice range.

Thus the same malformed bytes do not deterministically return `DecodeError` across supported Rust target widths/build arithmetic behavior.

## Gap

The format authority exposes `Result`-based decoding but does not make offset construction itself fallible. An untrusted length field can escape the intended error model on 32-bit targets.

This finding is deliberately scoped to decoder robustness / target-width qualification; it does not claim a demonstrated 64-bit verifier bypass.

## Evidence

- `crates/sm-format/src/semcode_decode.rs::decode_semcode_envelope`
- `crates/sm-format/src/local_format.rs::{read_u16_le,read_u32_le,read_f64_le,read_utf8}`
- `docs/spec/semcode.md` structural rejection contract

skulmakov-oss:
## Summary

Closes #1736 (FA-05-006): `decode_semcode_envelope`'s function `code_len` bounds check, `local_format.rs`'s five low-level readers (`read_u8`/`read_u16_le`/`read_u32_le`/`read_f64_le`/`read_utf8`), and the `DBG0`/`OWN0` section tag-sniffs all used raw `+` on attacker-controlled length fields. On a 32-bit target, a tiny malformed artifact can choose a near-`u32::MAX` `code_len` so the cursor addition wraps `usize`, producing an out-of-range slice-index panic instead of a deterministic decode error.

- **Genuinely reproduced**, not simulated: installed `i686-pc-windows-msvc` and confirmed the exact pre-fix panic (`slice index starts at 18 but ends at 2`) on a real 32-bit release build; post-fix the same artifact returns `Err(TruncatedFunction)`.
- **Fix**: every real bounds-check site now goes through a `checked_add`-based primitive (`checked_read_end` in `local_format.rs`; `checked_end` in `semcode_decode.rs`) that fails closed on overflow, mapping it to the same error the ordinary out-of-bounds case already produces.
- **One deliberate, narrow exception**: `diag_offset` uses `saturating_add` for diagnostic-only error-offset fields that never gate accept/reject (every call site is already inside an error-construction closure triggered by an already-failed `read_*` call). Confirmed by an independent rule-compliance review pass that no call site's result influences an accept/reject decision.
- **Regression coverage is width-independent**: rather than relying solely on a manual 32-bit run, tests force the cursor to `usize::MAX - 1` directly, so the exact overflow path is exercised on ordinary 64-bit CI too (`checked_end_rejects_cursor_near_usize_max_on_any_target_width`, plus the `local_format.rs` reader-level equivalents).
- Updated `docs/spec/semcode.md`'s Structural Contract with a new subsection documenting the invariant, including the distinction between the `code_len` check (a real accept/reject gate) and the `DBG0`/`OWN0` tag-sniffs (a lookahead probe for an optional section — a failed probe means "section absent," not rejection).

This PR's diff was adversarially reviewed by 5 independent agents (correctness, completeness, test-coverage, spec-accuracy, no-fallback-rule-compliance) before being opened; 4 real findings were confirmed and fixed pre-push (2 test-coverage gaps closed via the new width-independent tests above, 2 spec-doc inaccuracies corrected), 1 stylistic nitpick was independently refuted.

## Test plan

- [x] `cargo test -p sm-format --all-features` — 19/19 pass
- [x] `cargo test -p sm-format --target i686-pc-windows-msvc --release` — 19/19 pass, including genuine 32-bit reproduction/fix confirmation
- [x] `cargo test -p sm-verify --all-features --features sm-ir/profile-rust` — 81/81 pass, including a new `verify_semcode_token`-boundary panic-safety test
- [x] `cargo test -p sm-ir` — 104/104 pass
- [x] `cargo test -p sm-vm --all-features` — 23/23 pass
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` (per-package) — clean
- [x] `cargo check --workspace --all-features` / `--no-default-features` — clean
- [x] Full `scripts/admission_guard.ps1` PR-ready gate — `local_ci passed` (exit 0), twice (before and after the post-review fixes)

### DIFF
--- a/crates/sm-format/src/local_format.rs
+++ b/crates/sm-format/src/local_format.rs
@@ -688,30 +688,44 @@ pub fn write_f64_le(out: &mut Vec<u8>, v: f64) {
     out.extend_from_slice(&v.to_le_bytes());
 }
 
-pub fn read_u8(bytes: &[u8], i: &mut usize) -> Result<u8, SemcodeFormatError> {
-    if *i >= bytes.len() {
+/// #1736 (FA-05-006): shared bounds-check primitive for every low-level
+/// reader below. Uses `checked_add`, never raw `+`, so a maliciously large
+/// `start` (e.g. a cursor advanced by an attacker-controlled length field)
+/// cannot wrap past `usize::MAX` and produce a false in-bounds result -
+/// which is exactly how the pre-fix `*i + width > bytes.len()` comparison
+/// could be defeated on 32-bit targets.
+fn checked_read_end(
+    bytes_len: usize,
+    start: usize,
+    width: usize,
+) -> Result<usize, SemcodeFormatError> {
+    let end = start
+        .checked_add(width)
+        .ok_or(SemcodeFormatError::UnexpectedEof)?;
+    if end > bytes_len {
         return Err(SemcodeFormatError::UnexpectedEof);
     }
+    Ok(end)
+}
+
+pub fn read_u8(bytes: &[u8], i: &mut usize) -> Result<u8, SemcodeFormatError> {
+    let end = checked_read_end(bytes.len(), *i, 1)?;
     let v = bytes[*i];
-    *i += 1;
+    *i = end;
     Ok(v)
 }
 
 pub fn read_u16_le(bytes: &[u8], i: &mut usize) -> Result<u16, SemcodeFormatError> {
-    if *i + 2 > bytes.len() {
-        return Err(SemcodeFormatError::UnexpectedEof);
-    }
-    let v = u16::from_le_bytes([bytes[*i], bytes[*i + 1]]);
-    *i += 2;
+    let end = checked_read_end(bytes.len(), *i, 2)?;
+    let v = u16::from_le_bytes(bytes[*i..end].try_into().unwrap());
+    *i = end;
     Ok(v)
 }
 
 pub fn read_u32_le(bytes: &[u8], i: &mut usize) -> Result<u32, SemcodeFormatError> {
-    if *i + 4 > bytes.len() {
-        return Err(SemcodeFormatError::UnexpectedEof);
-    }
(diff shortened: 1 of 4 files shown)


skulmakov-oss:
@codex review

Issue #1771, opened 2026-08-16
FA-09-003: MAP_GET missing-key path swallows invalid default-register read and substitutes Unit


skulmakov-oss:
Umbrella: #1617
Module: **09 — `sm-vm`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · FALSE-READY · SEMANTIC-DRIFT**
Related: #1756 (`FA-07-016`), #1770 (`FA-09-002`)

## Trigger

Execute verified SemCode containing `Opcode::MapGet` where:

1. the requested key is absent from the map; and
2. `default_reg` is numerically within the verifier's register budget but is not present in the current VM frame.

The verifier records the default register only as a register reference and checks the maximum numeric register against `max_registers`; it does not prove that the register has been defined/materialized.

The VM executes missing-key `MapGet` as:

```rust
let result = pairs
    .iter()
    .find(|(k, _)| k == &key)
    .map(|(_, v)| v.clone())
    .unwrap_or_else(|| get_reg(vm, frame_idx, default_reg).unwrap_or(Value::Unit));
```

## Claim / confidence source

`docs/spec/vm.md` says malformed bytecode must not be treated as successful execution and the VM owns safe execution failure reporting.

`get_reg()` itself classifies a register that is outside the materialized frame as `RuntimeError::BadFormat("read invalid reg ...")`.

## Actual guarantee

`MapGet` catches that VM structural error and converts it to the valid runtime value `Value::Unit`.

A concrete reachable shape is:

```text
verified register budget = 4096
frame initially materializes r0..r15
MAP_GET ... default_reg=r100
key is absent
→ get_reg(r100) = BadFormat
→ unwrap_or(Value::Unit)
→ destination = Unit
→ execution continues successfully
```

No verifier bypass beyond the already-recorded lack of definite assignment in #1756 is required: `r100` is a legal numeric reference under the current verifier budget.

## Gap

This is an explicit error-to-plausible-value coercion inside the VM:

```text
invalid default register
→ BadFormat
→ swallowed
→ Unit
→ success
```

It is independent of #1770's broader register-initialization problem. Even if frame slots stopped using Unit as an uninitialized sentinel, this `unwrap_or(Value::Unit)` would still convert a real `get_reg` error into success.

## Cross-phase impact

```text
sm-verify accepts in-range default register
→ VerifiedSemCode
→ MAP_GET key miss
[...]

Issue #1770, opened 2026-08-16
FA-09-002: frame register storage materializes undefined registers as Unit and permits successful reads


skulmakov-oss:
Umbrella: #1617
Module: **09 — `sm-vm`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · SEMANTIC-DRIFT · FALSE-READY**
Related: #1756 (`FA-07-016`)

## Trigger

Execute verified SemCode that reads an in-budget register for which no value has been defined on the relevant path.

The verifier-side root cause is already recorded in #1756: `sm-verify` proves register range but not definite assignment.

The VM adds a separate fail-open root cause in its frame representation.

`push_frame()` allocates at least 16 registers regardless of arguments:

```rust
let initial_reg_count = 16usize.max(args.len());
let mut regs = vec![Value::Unit; initial_reg_count];
```

`write_reg()` grows the vector with:

```rust
frame.regs.resize(required, Value::Unit);
```

Thus register storage cannot distinguish:

```text
real language Unit value
from
never-defined register slot
```

## Claim / confidence source

`docs/spec/vm.md` states that the VM executes verified SemCode with safe execution failure reporting and that malformed bytecode must not be treated as successful execution.

#1756 currently records the verifier gap and expects an undefined register read to surface as VM `BadFormat` through `get_reg()`.

## Actual guarantee

That expectation is only true for register numbers outside the currently materialized vector.

For `r0..r15`, `push_frame()` has already materialized a legitimate-looking `Value::Unit` before any instruction executes. A program such as conceptually:

```text
RET r0
```

can therefore return `Unit` successfully even though no instruction defined `r0`.

The same problem extends to holes created by high-register writes:

```text
write r100
→ resize registers 16..100 with Unit
→ later read r50
→ successful Unit, although r50 was never defined
```

## Gap

The VM's physical storage representation erases the distinction between “uninitialized” and the valid Semantic value `Unit`.

This turns an invalid/unknown runtime state into a plausible language value and successful continuation rather than deterministic rejection.

It is related to but independent from #1756:

- #1756: verifier does not prove definition before use;
- this finding: VM representation itself converts some undefined states to `Unit` success.

Fixing only verifier dataflow would leave the VM's local invariant incapable of detecting this class if another producer/boundary later bypassed that proof.

## Evidence

- `crates/sm-vm/src/semcode_vm.rs::push_frame`
[...]

Issue #1756, opened 2026-08-15
FA-07-016: verifier checks register bounds but admits register reads with no proven definition


skulmakov-oss:
Umbrella: #1617
Module: **07 — `sm-verify`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · IDENTITY/TRUST-GAP · FALSE-READY**

## Trigger

Construct a function that reads a register that is within the numeric register budget but has never been initialized on the relevant execution path.

A minimal entrypoint example is conceptually:

```text
main:
    RET r0
```

with `has_src = 1` and no preceding instruction that defines `r0`.

`sm-verify::decode_operands()` calls `mark_reg(r0)`, and `verify_function_code()` proves only:

```text
max referenced register < max_registers
```

There is no register-definition / definite-assignment dataflow proof.

The VM reads registers through:

```rust
fn get_reg(...) -> Result<Value, RuntimeError> {
    ...
    .ok_or_else(|| RuntimeError::BadFormat(format!("read invalid reg r{}", r)))
}
```

so the admitted example reaches VM structural `BadFormat`.

## Claim / confidence source

The verifier is documented as the canonical public admission layer. `docs/spec/semcode.md` states that the VM is not the primary structural admission gate, and `docs/spec/verifier.md` says malformed/contract-inconsistent bytecode is verifier-rejected before successful VM execution.

## Actual guarantee

The verifier proves register **range**, not register **existence/definition**. Any in-range register number is treated as a valid operand reference regardless of whether execution can materialize a value there.

## Gap

A bytecode stream with a structurally impossible register read can obtain `VerifiedSemCode` and only later be classified by the VM as `BadFormat`.

This is broader than an ordinary runtime type trap: the VM itself identifies the state as malformed bytecode/register state.

## Evidence

- `crates/sm-verify/src/lib.rs::decode_operands`
- `crates/sm-verify/src/lib.rs::verify_function_code` — max-register proof only
- `crates/sm-vm/src/semcode_vm.rs::get_reg`
- `docs/spec/verifier.md`
- `docs/spec/semcode.md`

Phase A evidence only. **No repair during Phase A.**

skulmakov-oss:
## Phase A strengthening note from Module 09 (`sm-vm`)

Audit HEAD remains `4ac4d1a169902777e760e23eaee92aa233edc04d`.

[...]

Issue #1750, opened 2026-08-15
FA-07-010: verifier known-function-first resolution can bypass builtin capability proof while VM dispatches builtin-first


skulmakov-oss:
Umbrella: #1617
Module: **07 — `sm-verify`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · SEMANTIC-DRIFT · IDENTITY/TRUST-GAP · FALSE-READY**
Related: #1653 (`FA-02-021`)

## Trigger

Construct SemCode that defines an ordinary function with a name also recognized by the VM builtin dispatcher, then calls that name under a header lacking the builtin's required capability.

A minimal verifier-specific example is a function named `to_text` under a header without `CAP_TEXT_VALUES`.

Program-level verifier resolution is:

```rust
if known_functions.contains(callee.as_str()) {
    continue;
}
if let Some(required_capabilities) = builtin_call_required_capabilities(callee) {
    // capability proof
}
```

So a same-named internal function wins and the builtin capability check is skipped.

VM `Opcode::Call` has the opposite precedence:

```rust
if let Some(result) = try_eval_builtin_call(..., &callee, &args)? {
    // builtin executes
} else {
    push_frame(vm, &callee, ...)?;
}
```

Thus the runtime builtin wins before the internal function frame is considered.

## Claim / confidence source

`docs/spec/verifier.md` and `docs/spec/semcode.md` state that the verifier proves capability consistency with actual operation usage. The source semantics also state that user-defined functions win over same-named builtins, making the current cross-layer priority mismatch explicit.

## Actual guarantee

The verifier's capability proof depends on a call-target interpretation different from the VM's actual dispatch interpretation.

For example:

```text
define internal function "to_text"
CALL "to_text"
header lacks CAP_TEXT_VALUES
→ verifier sees known function → no builtin capability requirement → ACCEPT
→ VM sees builtin name first → executes builtin to_text
```

The same structural problem applies to the other runtime-intercepted builtin names.

## Gap

`VerifiedSemCode` can claim capability consistency under one target-resolution model while execution uses another. This invalidates the admission proof itself and can allow builtin behavior not represented by the admitted header capability envelope.

This is distinct from #1653: that issue records source/user-function wrong-code. This issue records the **verifier capability-proof bypass** caused by the same cross-layer precedence disagreement.

## Evidence

[...]

PR #1805, opened 2026-08-18
fix(sm-verify): reject reachable function fallthrough

### LINKED ISSUE #1745: FA-07-005: verifier admits functions that fall off the instruction stream and later fail VM execution as BadFormat
Umbrella: #1617
Module: **07 — `sm-verify`**
Severity: **P1**
Classification: **FAIL-OPEN · IDENTITY/TRUST-GAP · FALSE-READY**

## Trigger

Construct a syntactically decodable function whose instruction stream ends after a non-terminating instruction, for example:

```text
LOAD_BOOL r0, 1
<end of function>
```

or an empty instruction stream after the function-local tables/sections.

`verify_function_code()` walks instructions until `cursor == code.len()`, validates each decoded operand, and then returns `Ok(PendingVerifiedFunction { ... })`. It does not prove that control flow terminates through `RET` or another admitted terminal condition.

At runtime the VM detects the fallthrough only after execution advances beyond the final instruction:

```rust
if pc >= instr_rel_len {
    return Err(RuntimeError::BadFormat(format!(
        "pc out of range in '{}': {}", ...
    )));
}
```

## Claim / confidence source

`docs/spec/verifier.md` defines `sm-verify` as the public admission layer and says malformed/contract-inconsistent bytecode must be rejected before successful VM execution. `docs/spec/semcode.md` says the VM is not the primary structural admission gate.

## Actual guarantee

The verifier proves only that the bytes can be sequentially decoded and that explicit jump targets are valid. It does not prove that a function has a structurally valid exit path.

Therefore a function that necessarily runs off the end can still obtain a canonical `VerifiedSemCode` token and then fail with VM `BadFormat`.

## Gap

A structurally non-executable function is promoted to verified state. The canonical token therefore does not prove enough to prevent a basic bytecode-format failure during execution.

## Evidence

- `crates/sm-verify/src/lib.rs::verify_function_code`
- `crates/sm-vm/src/semcode_vm.rs` execution loop — `pc >= instr_rel_len` → `BadFormat`
- `docs/spec/verifier.md`
- `docs/spec/semcode.md`

skulmakov-oss:
Closes #1745

Umbrella: #1617

## Root cause

`verify_function_code()` decoded each instruction and validated explicit jump targets, but it did not prove that control flow reachable from function entry remained inside the executable instruction stream. Sequentially decodable bytecode could therefore reach `code.len()` and still receive `VerifiedSemCode`; the VM later rejected the same artifact at its defensive `pc >= instr_rel_len` guard as `RuntimeError::BadFormat`.

## Pre-fix reproduction

Before the repair, a disposable regression harness demonstrated all three facts:

- `LOAD_BOOL r0, true` followed by EOF was admitted by `verify_semcode()`.
- An empty executable instruction stream was admitted by `verify_semcode()`.
- Executing the admitted non-empty artifact reached `RuntimeError::BadFormat("pc out of range in 'main': 4")`.

The reproduction passed 3/3 assertions against unmodified `main`. The permanent rejection tests were then added first and failed exactly the three intended cases before the implementation (87 existing tests passed, 3 new tests failed).

## Repaired invariant

`VerifiedSemCode` now proves that every control-flow successor reachable from function entry is either a valid decoded instruction start or an admitted terminal condition. A reachable end-of-stream fallthrough is rejected deterministically as `VerificationCode::ReachableFunctionFallthrough` before the token is issued.

The verifier reuses the instruction starts, decoded next offsets, and jump targets collected by its existing instruction walk; it does not introduce a second bytecode decoder.

## CFG rule

The successor classification is derived from the current VM implementation:

- `RET`: terminal, no successor.
- `JMP`: explicit target only.
- `JMP_IF`: explicit target plus decoded next instruction.
- Every other opcode, including `CALL` and `CLOSURE_CALL`: decoded next instruction.

Existing explicit jump range and instruction-boundary checks remain intact and retain `InvalidJumpTarget` diagnostics.

This is a structural closure proof, not a termination proof. Closed infinite loops remain valid, unreachable trailing instructions do not create reachable fallthrough, and the verifier does not require the final encoded instruction to be `RET`.

## Regression matrix

- REJECT reachable ordinary instruction followed by EOF.
- REJECT empty executable instruction stream.
- REJECT conditional branch with a valid explicit branch and EOF fallthrough.
- ACCEPT a normal function ending in `RET`.
- ACCEPT an unconditional jump to a valid return instruction boundary.
- ACCEPT a structurally closed infinite loop with unreachable trailing fallthrough.
- ACCEPT valid `CALL` and `CLOSURE_CALL` fallthrough to `RET`.
- Preserve the existing invalid jump-target tests and diagnostics.
- Guard the complete public `VerificationCode` variant vocabulary, including `ReachableFunctionFallthrough`.

Tests build SemCode through the IR emitter; they do not search for raw-byte coincidences.

## Validation

- `cargo test -p sm-verify --features sm-ir/profile-rust,sm-ir/profile-logos,sm-ir/debug-symbols --lib` — PASS, 90/90.
- `cargo test -p sm-vm --lib` — PASS, 69/69.
- `cargo test --test public_api_contracts` — PASS, 5/5.
- `cargo clippy --workspace --all-targets -- -D warnings` — PASS.
- `git diff --check` and `git diff --cached --check` — PASS.
- Repository Windows-safe all-workspace formatting gate (`Invoke-WorkspaceFmtCheck`) — PASS for all 36 packages.
- `pwsh -NoProfile -File scripts/admission_guard.ps1 -PRReady` — PASS (workspace check, clippy, all-package format check, full workspace tests, public API contracts).

The literal one-shot `cargo fmt --all -- --check` was also attempted in the original worktree, a substituted drive path, and a disposable short worktree; Cargo failed before rustfmt with Windows `os error 206` because the aggregate argument list exceeds the platform command-line limit. The repository's authoritative per-package Windows-safe equivalent passed all 36 packages and is the formatter used by `-PRReady`.

## Scope

This PR repairs FA-07-005 / #1745 only: verifier-owned reachable control-flow closure, focused verifier regressions, the narrow public diagnostic contract guard, and minimum SemCode/verifier specification updates. It does not change the VM or remove its defensive PC bounds check.

#1746 debug-PC validation, #1750 builtin/function resolution, #1756 definite register assignment, #1770 register representation, #1771 MAP_GET behavior, pending-policy findings, ownership findings, and all other findings are untouched. No termination proof, CFG optimization, dead-code elimination, VM refactor, or unrelated cleanup is included.

### DIFF
--- a/crates/sm-verify/src/lib.rs
+++ b/crates/sm-verify/src/lib.rs
@@ -56,6 +56,11 @@ pub enum VerificationCode {
     /// header predates the revision whose contract actually admits this
     /// opcode's semantics.
     OpcodeRequiresNewerHeader,
+    /// A control-flow successor reachable from the function entry falls off
+    /// the executable instruction stream instead of reaching another
+    /// instruction boundary or an admitted terminal instruction (`RET`).
+    /// Closed loops remain admissible; this code does not imply termination.
+    ReachableFunctionFallthrough,
 }
 
 #[cfg(feature = "std")]
@@ -547,6 +552,7 @@ fn verify_function_code(
     let instr_start = cursor;
     let instr_len = code.len().saturating_sub(instr_start);
     let mut instr_starts = Vec::new();
+    let mut instruction_successors = Vec::new();
     let mut jump_targets = Vec::new();
     let mut string_refs = Vec::new();
     let mut max_register: Option<usize> = None;
@@ -577,6 +583,23 @@ fn verify_function_code(
             ),
         })?;
         let refs = decode_operands(name, code, &mut cursor, offset, opcode, true)?;
+        let next_offset = cursor - instr_start;
+        let jump_target = refs.jump_targets.first().copied();
+        let successors = match (opcode, jump_target) {
+            (Opcode::Ret, _) => InstructionSuccessors::None,
+            (Opcode::Jmp, Some(target)) => InstructionSuccessors::One(target),
+            (Opcode::JmpIf, Some(target)) => InstructionSuccessors::Two(target, next_offset),
+            (Opcode::Jmp | Opcode::JmpIf, None) => {
+                return Err(reject_one(
+                    name,
+                    VerificationCode::InvalidJumpTarget,
+                    offset,
+                    "jump instruction has no decoded target",
+                ));
+            }
+            _ => InstructionSuccessors::One(next_offset),
+        };
+        instruction_successors.push(successors);
         let min_rev = opcode.minimum_semcode_revision();
         if header.rev < min_rev {
             return Err(reject_one(
@@ -654,6 +677,8 @@ fn verify_function_code(
         }
     }
 
+    verify_reachable_control_flow(name, instr_len, &instr_starts, &instruction_successors)?;
+
     let mut call_targets = Vec::new();
     for (offset, sid, usage) in string_refs {
         if sid >= string_count {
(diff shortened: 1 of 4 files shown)


skulmakov-oss:
@codex review

Issue #1822, opened 2026-08-21
prom-runtime: ExecutionSession admission ignores configured runtime quotas


skulmakov-oss:
Umbrella #1617

## Summary

`prom-runtime::ExecutionSession` and `prom-runtime::GateExecutionSession` (`crates/prom-runtime/src/lib.rs`) each already carry a caller-supplied `self.config: ExecutionConfig` — including explicit `kernel_bound()` constructors — and execute with `self.config`. But their `run_verified_semcode` / `run_verified_semcode_entry` methods still admit via the plain, hardcoded-`VerifiedLocal` `sm_verify::verify_semcode_token(bytes)` instead of `sm_verify::verify_semcode_token_with_quotas(bytes, self.config.quotas)` (added in #1757, umbrella #1617).

This is the same class of admission/execution quota-profile mismatch #1757 fixed for `sm-vm`'s byte-based compatibility shims — a `kernel_bound()`-constructed `prom-runtime` session still cannot admit any SemCode exceeding `VerifiedLocal`'s 4096-register budget, even though it is explicitly configured to execute under `KernelBound` (8192).

## Origin

Discovered during adversarial review of #1757's PR (#1821) — both the semantic and qualification review lenses independently flagged this. #1757 deliberately left `prom-runtime` out of scope (a separate, unresearched crate with its own dedicated test suite), fixing only the `sm-vm` sibling gap it also found (`run_verified_semcode_with_host_and_capabilities`) directly in that PR. This issue tracks the `prom-runtime` half.

## Exact locations (confirmed live, may drift — re-verify before implementing)

- `crates/prom-runtime/src/lib.rs`: `ExecutionSession::run_verified_semcode` / `run_verified_semcode_entry`
- `crates/prom-runtime/src/lib.rs`: `GateExecutionSession::run_verified_semcode` / `run_verified_semcode_entry`

Both already have a `kernel_bound()` constructor (`ExecutionSession::kernel_bound(...)`, `GateExecutionSession::kernel_bound(...)`) — a *different, unrelated* API from `sm_runtime_core::RuntimeQuotas::kernel_bound()`; do not confuse the two when investigating.

## Required invariant

```text
ExecutionSession.config.quotas
        ↓
verifier admission
        ↓
same quotas
        ↓
runtime execution
```

The same `self.config.quotas` that governs execution must also govern admission for the same call.

## Required regression

```text
KernelBound + r5000 -> admission passes
[...]

Issue #1775, opened 2026-08-16
FA-09-007: invalid host AbiValue::Quad bytes are masked into valid quad values by VM bridge


skulmakov-oss:
Umbrella: #1617
Module: **09 — `sm-vm`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · IDENTITY/TRUST-GAP · SEMANTIC-DRIFT · FALSE-READY**

## Trigger

A `PrometheusHostAbi` implementation returns an `AbiValue::Quad(u8)` whose byte is outside the canonical four-value quad domain, for example:

```rust
AbiValue::Quad(4)
AbiValue::Quad(0xff)
```

`prom-abi` exposes `AbiValue::Quad(u8)`, so such values are representable at the host ABI boundary.

The VM host bridge converts incoming ABI values with:

```rust
AbiValue::Quad(q) => Value::Quad(u8_to_quad(q))
```

and `u8_to_quad` implements:

```rust
match v & 0b11 {
    0 => QuadVal::N,
    1 => QuadVal::F,
    2 => QuadVal::T,
    _ => QuadVal::S,
}
```

## Claim / confidence source

The host ABI is a trust boundary. Invalid/unsupported boundary values must not silently acquire valid Semantic meaning.

The VM already has explicit host-ABI error plumbing (`RuntimeError::HostAbi`) for failures returned by the host adapter, and `docs/spec/vm.md` requires safe execution failure reporting rather than best-effort reinterpretation.

## Actual guarantee

Every possible `u8` becomes a valid quad after masking the low two bits.

Examples:

```text
AbiValue::Quad(4)    -> 4 & 3 = 0 -> QuadVal::N
AbiValue::Quad(5)    -> 5 & 3 = 1 -> QuadVal::F
AbiValue::Quad(0xff) -> 3         -> QuadVal::S
```

The VM cannot distinguish a canonical host value from malformed high-bit data after conversion.

## Gap

An invalid host-bound value is silently normalized into a valid executable Semantic value:

```text
malformed host quad byte
→ VM bitmask normalization
→ valid N/F/T/S
→ successful execution
```

This weakens both ABI integrity and quad evidence semantics. It is the same governing fail-open shape as non-canonical SemCode literal normalization, but occurs at the **host ABI → VM value** boundary rather than the verifier → VM bytecode boundary.

[...]

Issue #1758, opened 2026-08-15
FA-07-018: with_decoded_envelopes is documented workspace-internal but is a callable public API


skulmakov-oss:
Umbrella: #1617
Module: **07 — `sm-verify`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P3**
Classification: **FALSE-READY · INERT-CONTRACT**

## Trigger

Use `VerifiedSemCode::with_decoded_envelopes` from a crate outside the Semantic workspace.

The method is documented as:

```text
Workspace-internal accessor for VM preparation.
Exposes the decoded internal state safely without creating a stable public API.
```

but its Rust visibility is:

```rust
#[doc(hidden)]
pub fn with_decoded_envelopes<...>(...)
```

`#[doc(hidden)]` affects generated documentation only. It does not restrict downstream crates from naming and calling a `pub` method, and the parameter types are themselves public.

## Claim / confidence source

The source comment explicitly claims this mechanism is workspace-internal and does not create a stable public API.

## Actual guarantee

The method is part of the externally callable Rust surface. The repository's textual public API inventory also sees public items in `sm-verify/src/lib.rs`, reinforcing that the actual visibility is public even if rustdoc hides it.

## Gap

The implementation mechanism does not enforce the ownership statement in its own documentation. An internal VM-preparation seam is exposed as an ordinary downstream callable API.

This is P3: the method is read-only over an already-verified token, so no verifier bypass or semantic mutation is demonstrated. It is a public-boundary honesty/maintenance defect.

## Evidence

- `crates/sm-verify/src/lib.rs::VerifiedSemCode::with_decoded_envelopes`
- `tests/public_api_contracts.rs`

Phase A evidence only. **No repair during Phase A.**

Issue #1755, opened 2026-08-15
FA-07-015: public API inventory snapshots sm-verify lib.rs but not its public admission submodule surfaces


skulmakov-oss:
Umbrella: #1617
Module: **07 — `sm-verify`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **FALSE-READY · QUALIFICATION-GAP**

## Trigger

Change a public type/function inside either public verifier submodule without changing the `pub mod` declaration in `src/lib.rs`:

- `hello_pending_admission.rs`
- `hello_real_semcode_admission.rs`

`tests/public_api_contracts.rs` includes only:

```text
crates/sm-verify/src/lib.rs
```

for the verifier crate. Its textual snapshot scanner does not recursively expand Rust modules. It therefore records the two `pub mod ...` declarations but not the public structs, enums and functions defined inside those files.

By contrast the same inventory explicitly snapshots an additional source file for `sm-vm`, showing that multi-file public surfaces are expected to be listed separately when they matter.

## Claim / confidence source

The repository uses the checked-in public API inventory as a release-facing contract drift guard, and `sm-verify` appears in that guard.

## Actual guarantee

Only the public items physically written in `sm-verify/src/lib.rs` are frozen. Public admission APIs in the two submodule files can change while the `sm-verify` snapshot remains green.

## Gap

The qualification signal “sm-verify public API is inventoried” is stronger than the actual proof boundary. The most policy-like provisional verifier APIs are outside the snapshot despite being publicly reachable through the crate.

## Evidence

- `tests/public_api_contracts.rs::TARGETS`
- `crates/sm-verify/src/lib.rs` — `pub mod hello_pending_admission`, `pub mod hello_real_semcode_admission`
- `crates/sm-verify/src/hello_pending_admission.rs`
- `crates/sm-verify/src/hello_real_semcode_admission.rs`

Phase A evidence only. **No repair during Phase A.**

PR #1826, opened 2026-08-23
fix(sm-vm): bind verified entry token to execution target

### LINKED ISSUE #1772: FA-09-004: verified entry token binding can be bypassed by canonical function helper's independent func_name
Umbrella: #1617
Module: **09 — `sm-vm`**
Severity: **P1**
Classification: **IDENTITY/TRUST-GAP · FALSE-READY · SEMANTIC-DRIFT**

## Trigger

Create a `VerifiedEntrySemCode` token resolved for function `A`, then call the public canonical function-invocation helper with an independent function name `B` from the same verified artifact:

```rust
let entry_a = token.require_entry("A")?;
run_verified_function_semcode_with_args(&entry_a, "B", args)
```

`run_verified_function_semcode_with_args()` accepts both:

```rust
token: &VerifiedEntrySemCode
func_name: &str
```

and delegates to `run_verified_function_semcode_with_args_and_config()`.

That implementation prepares the verified artifact, checks/uses `func_name`, and calls:

```rust
push_frame(&mut vm, func_name, args, None)?;
```

It does not require `func_name == token.entry()`.

## Claim / confidence source

`VerifiedEntrySemCode` is documented in `sm-verify` as:

> a verified artifact that has also been confirmed to contain a specific entry point function

and as the **second boundary in the canonical verified execution path, directly preceding execution in `sm-vm`**.

`docs/spec/vm.md` likewise says canonical token-first execution uses `run_verified_entry_semcode*`, where entry resolution has already occurred.

The function helper itself is documented as the “Canonical verified function invocation path with arguments and structured return value” and says it requires a `VerifiedEntrySemCode` token ensuring bytecode verification has passed.

## Actual guarantee

The token proves one specific entry was resolved, but this public helper does not consume that entry identity as the execution target. It uses a second caller-controlled string instead.

Possession of a token for `A` therefore suffices to invoke any existing function `B` in the same artifact through this helper.

## Gap

The entry-resolved token's identity binding is not authoritative at the function-invocation boundary:

```text
VerifiedEntrySemCode(entry=A)
→ independent func_name=B
→ push_frame(B)
→ B executes
```

This makes the entry token weaker than its type/documentation imply. It proves “some named entry was checked” rather than “this is the entry being executed” on this public path.

This does not bypass whole-artifact verification: `B` must exist in the same verified artifact. The finding is specifically about **entry identity binding**, not artifact admission.

## Cross-phase impact

```text
sm-verify require_entry(A)
→ VerifiedEntrySemCode bound to A
→ sm-vm canonical function helper accepts B
→ execution starts at B
```

## Evidence

- `crates/sm-verify/src/lib.rs::VerifiedEntrySemCode`
- `crates/sm-verify/src/lib.rs::VerifiedSemCode::require_entry`
- `crates/sm-vm/src/semcode_vm.rs::run_verified_function_semcode_with_args`
- `crates/sm-vm/src/semcode_vm.rs::run_verified_function_semcode_with_args_and_config`
- `crates/sm-vm/src/semcode_vm.rs::push_frame`
- `docs/spec/vm.md`

## Repair note

Phase B only: either execute `token.entry()` directly on the canonical entry-bound path, or use a token/type whose resolved function identity matches the function invocation being requested. Do not retain an entry-bound token plus an unconstrained second target identity and call that binding canonical.

skulmakov-oss:
Closes #1772
Umbrella #1617

## Root cause

`run_verified_function_semcode_with_args` / `run_verified_function_semcode_with_args_and_config` accepted both a `VerifiedEntrySemCode` token (proving one specific entry was resolved) and an independent `func_name: &str`, and used only `func_name` to select the execution target:

```rust
pub fn run_verified_function_semcode_with_args_and_config(
    token: &VerifiedEntrySemCode<'_, '_>,
    func_name: &str,
    args: Vec<Value>,
    config: ExecutionConfig,
) -> Result<Value, RuntimeError> {
    ...
    push_frame(&mut vm, func_name, args, None)?;
    ...
}
```

`token.entry()` was never consulted. Possession of a token bound to `"A"` therefore sufficed to invoke any function `"B"` in the same verified artifact through this public helper.

## Live API audit

Every public function in `crates/sm-vm/src/semcode_vm.rs` accepting `&VerifiedEntrySemCode`:

| API | uses `token.entry()`? | separate target string? | can diverge? |
|---|---|---|---|
| `run_verified_entry_semcode_with_host_and_capabilities_and_config` | yes | no | no |
| `run_verified_entry_semcode_with_application_host_and_capabilities_and_config` | yes | no | no |
| `run_verified_entry_semcode` | yes (delegates) | no | no |
| `run_verified_entry_semcode_with_config` | yes | no | no |
| `run_verified_entry_semcode_with_profile` (`vm-profile` feature) | yes | no | no |
| **`run_verified_function_semcode_with_args`** | **no (delegates to sibling below)** | **yes (`func_name`)** | **yes — the bug** |
| **`run_verified_function_semcode_with_args_and_config`** | **no** | **yes** | **yes — the bug, root implementation** |

Only this one function (and its thin wrapper) has the defect — every other token-first API already derives its execution target exclusively from `token.entry()`. No host-aware sibling of `run_verified_function_semcode_with_args` exists that would need a separate fix.

## Pre-fix reproduction

Built via `emit_ir_to_semcode` with two functions returning distinct sentinels (`a_fn` → `I32(111)`, `b_fn` → `I32(222)`). Resolved `token.require_entry("a_fn")`, then called `run_verified_function_semcode_with_args(&entry_a, "b_fn", vec![])`.

**PRE-FIX: `Ok(Value::I32(222))`** — `b_fn`'s body genuinely executed, not merely "some function existed" or a generic success. Confirmed by running this exact test against unmodified code before applying any fix.

## Repair design

Chose the strongest option per the task's own preference order: **removed `func_name` entirely** rather than keeping it with an equality check. The execution target is now always `token.entry()`, making the divergence structurally inexpressible rather than merely runtime-rejected.

This was safe to do without any compatibility-preserving fallback because a full repo-wide call-site audit found **all 15 call sites** — 13 in `sm-vm`'s own test suite, plus `examples/workbench_semantic/src/main.rs` and `examples/quad_logic_calculator/src/main.rs` — always passed a `func_name` string identical to whatever name `require_entry(...)` had just resolved. The parameter was 100% redundant everywhere it was used correctly; the only "use" for divergence would have been the exploit path itself. No caller needed arbitrary-target invocation independent of a bound entry token, so no `VerifiedSemCode` + fresh `require_entry(target)` redesign was needed anywhere.

The internal `!program.functions.contains_key(func_name)` defense-in-depth check was kept (now checking `token.entry()` instead of the removed parameter) rather than deleted — `require_entry`'s own contract already guarantees `token.entry()` names a real function in the artifact, but this check is retained as a deliberate belt-and-suspenders invariant proof on `prepare_verified_execution`'s output, not because it is currently reachable.

## Public API

```diff
-pub fn run_verified_function_semcode_with_args(token: &VerifiedEntrySemCode<'_, '_>, func_name: &str, args: Vec<Value>) -> Result<Value, RuntimeError>
+pub fn run_verified_function_semcode_with_args(token: &VerifiedEntrySemCode<'_, '_>, args: Vec<Value>) -> Result<Value, RuntimeError>

-pub fn run_verified_function_semcode_with_args_and_config(token: &VerifiedEntrySemCode<'_, '_>, func_name: &str, args: Vec<Value>, config: ExecutionConfig) -> Result<Value, RuntimeError>
+pub fn run_verified_function_semcode_with_args_and_config(token: &VerifiedEntrySemCode<'_, '_>, args: Vec<Value>, config: ExecutionConfig) -> Result<Value, RuntimeError>
```

`tests/golden_snapshots/public_api/sm_vm_semcode_vm.txt` updated to match, confirmed via `public_api_inventory_matches_checked_in_contract_snapshots` — the delta is exactly these two signatures, nothing else moved. Documented as an intentional, deliberate narrowing of an unsound identity surface (correctness hardening, not scope creep).

All 15 call sites updated:
- `crates/sm-vm/src/lib.rs`: 13 test call sites (mechanical — dropped the redundant string argument).
- `examples/quad_logic_calculator/src/main.rs`, `examples/workbench_semantic/src/main.rs`: same mechanical fix. `quad_logic_calculator` is a workspace member and was compiled/clippy'd clean as part of the full workspace gates below. `workbench_semantic` is **excluded from the workspace** (`Cargo.toml`'s `exclude = ["examples/workbench_semantic"]`) and requires separate native-GUI setup to build standalone — not independently build-verified in this environment, but its edit is mechanically identical (same code shape, same call pattern) to `quad_logic_calculator`'s, which did compile clean.
- Two of `sm-vm`'s own pre-existing tests (`single_argument_initializes_only_r0`, `multiple_arguments_initialize_exactly_supplied_registers`, from #1770) had been (unintentionally) relying on the very divergence this PR closes — resolving one `"main"`-bound token and then invoking `"read_r0"`/`"read_r1"` through it. Fixed by resolving a separate entry token per target function (`token.require_entry("read_r0")`, `token.require_entry("read_r1")`, etc.) — the same `VerifiedSemCode` supports resolving multiple independent entries, so this is a direct, no-compromise adaptation, not a workaround.

## Regression matrix (`crates/sm-vm/src/semcode_vm.rs`, `semcode_vm::tests`)

1. `entry_token_a_executes_a` — token(A) + request A → A executes.
2. `entry_token_execution_target_is_not_independently_selectable` — token(A) + "request B" is no longer expressible at all; there is no `func_name` parameter through which a caller holding `entry_a` could name `b_fn`. The bug is eliminated by construction (the crate's own successful compilation is part of the proof), not by a runtime check.
3. `entry_token_b_executes_b` — independently resolving token(B) → B executes normally.
4. `entry_token_missing_entry_still_fails_at_resolution` — unknown entry name still produces the existing `EntryResolutionError::MissingEntry`, unchanged.
5. `entry_token_config_aware_sibling_uses_token_entry_and_delivers_args` — the `_and_config` sibling obeys the identical rule and correctly delivers arguments to the bound target.
6. Argument delivery is also covered by item 5 and by the many pre-existing single/multi-argument tests in this module and `crates/sm-vm/src/lib.rs`, all passing unchanged.
7. Ordinary canonical verified-entry execution is unchanged — covered by the full pre-existing `vm_runs_*`/`test_N_*` suite (~90 tests), passing unchanged.

## Doc

`docs/spec/vm.md`, Standard Execution Rule section:

> `VerifiedEntrySemCode`'s entry identity is authoritative for execution. Possession of a token resolved for one entry does not authorize selecting a different function from the same verified artifact through an independent identifier. Every public execution API accepting a `VerifiedEntrySemCode` derives its execution target from `token.entry()` alone.

`docs/architecture/svm_verified_execution_core.md` was checked and does not reference this API — no update needed there.

## Validation

- `cargo test -p sm-vm --all-features`: 129+1+23 pass (124 baseline + 5 new #1772 regression tests)
- `cargo test -p sm-verify --features sm-ir/profile-rust`: 106/106 pass (unchanged)
- `cargo test --test public_api_contracts`: 5/5 pass — the one intentional signature-delta snapshot update
- `cargo test --lib` (root crate): 7/7 pass, unaffected
- `cargo clippy -p sm-vm --all-targets --all-features -- -D warnings`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean (includes `quad_logic_calculator`)
- `cargo fmt` clean on all touched crates
- `git diff --check` / `git diff --cached --check`: clean
- `pwsh -NoProfile -File scripts/admission_guard.ps1 -PRReady`: **PASS**
- `pwsh -NoProfile -File tools/7hell/run_ci.ps1`: **PASS**

## Scope

`crates/sm-verify/`, `crates/sm-runtime-core/`, `crates/sm-format/`, `crates/sm-ir/` untouched. No SemCode format/revision change. Token-first `run_verified_entry_semcode*` APIs (already correct, per the audit table) untouched. `#1775`, `#1822`, `#1755`, `#1758` untouched — independent, separate repairs in this same campaign.

### DIFF
--- a/crates/sm-vm/src/lib.rs
+++ b/crates/sm-vm/src/lib.rs
@@ -39,7 +39,7 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("get_num").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "get_num", vec![]).expect("run");
+        let res = run_verified_function_semcode_with_args(&entry, vec![]).expect("run");
         assert_eq!(res, Value::I32(42));
     }
 
@@ -49,8 +49,8 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("add_five").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "add_five", vec![Value::I32(10)])
-            .expect("run");
+        let res =
+            run_verified_function_semcode_with_args(&entry, vec![Value::I32(10)]).expect("run");
         assert_eq!(res, Value::I32(15));
     }
 
@@ -60,7 +60,7 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("get_quad").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "get_quad", vec![]).expect("run");
+        let res = run_verified_function_semcode_with_args(&entry, vec![]).expect("run");
         assert_eq!(res, Value::Quad(QuadVal::T));
     }
 
@@ -70,12 +70,8 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("negate_quad").expect("entry");
-        let res = run_verified_function_semcode_with_args(
-            &entry,
-            "negate_quad",
-            vec![Value::Quad(QuadVal::F)],
-        )
-        .expect("run");
+        let res = run_verified_function_semcode_with_args(&entry, vec![Value::Quad(QuadVal::F)])
+            .expect("run");
         assert_eq!(res, Value::Quad(QuadVal::T));
     }
 
@@ -89,7 +85,7 @@ mod tests {
             type_name: "Pair".into(),
             slots: vec![Value::I32(1), Value::I32(2)],
         });
-        let res = run_verified_function_semcode_with_args(&entry, "swap", vec![arg]).expect("run");
+        let res = run_verified_function_semcode_with_args(&entry, vec![arg]).expect("run");
         assert_eq!(
             res,
             Value::Record(RecordCarrier {
@@ -105,7 +101,7 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("need_two").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "need_two", vec![Value::I32(5)]);
+        let res = run_verified_function_semcode_with_args(&entry, vec![Value::I32(5)]);
         assert!(res.is_ok() || res.is_err());
     }
 
@@ -115,11 +111,7 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("add_one").expect("entry");
-        let res = run_verified_function_semcode_with_args(
-            &entry,
-            "add_one",
-            vec![Value::Quad(QuadVal::T)],
-        );
+        let res = run_verified_function_semcode_with_args(&entry, vec![Value::Quad(QuadVal::T)]);
         assert!(res.is_err());
     }
 
@@ -148,8 +140,7 @@ mod tests {
 
         for i in 1..=5 {
             let res =
-                run_verified_function_semcode_with_args(&entry, "double", vec![Value::I32(i)])
-                    .expect("run");
+                run_verified_function_semcode_with_args(&entry, vec![Value::I32(i)]).expect("run");
             assert_eq!(res, Value::I32(i * 2));
         }
     }
@@ -174,8 +165,8 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("caller").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "caller", vec![Value::F64(5.0)])
-            .expect("run");
+        let res =
+            run_verified_function_semcode_with_args(&entry, vec![Value::F64(5.0)]).expect("run");
         assert_eq!(
             res,
             Value::F64(1005.0),
@@ -191,8 +182,8 @@ mod tests {
         let bytes = compile_program_to_semcode(src).expect("compile");
         let token = verify_semcode_token(&bytes).expect("verify");
         let entry = token.require_entry("caller").expect("entry");
-        let res = run_verified_function_semcode_with_args(&entry, "caller", vec![Value::F64(16.0)])
-            .expect("run");
+        let res =
+            run_verified_function_semcode_with_args(&entry, vec![Value::F64(16.0)]).expect("run");
         assert_eq!(
             res,
             Value::F64(2016.0),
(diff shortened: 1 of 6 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1854, opened 2026-08-30
fix(calls): preserve source argument evaluation order and validate FnSig metadata (#1722 #1665)

### LINKED ISSUE #1665: FA-02-033: malformed public FnSig metadata can panic call argument reordering instead of failing closed
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — public API invariant violation can panic**

## Finding

`FnSig` is public and exposes `params`, optional `param_names`, and optional `param_defaults` independently. `type_check_function_with_table()` accepts a caller-supplied public `FnTable`.

`reorder_call_args()` / `finalize_ordered_call_args()` assume metadata lengths are coherent. Examples:

- if `param_names` contains more entries than `params`, a named argument may resolve to `param_index >= ordered.len()` and index `ordered[param_index]`;
- if `param_names` is shorter than `params`, a missing argument path indexes `param_names[idx]` while iterating `ordered.len() == params.len()`.

There is no canonical upfront FnSig shape validation for public tables.

## Risk

Malformed external/public frontend metadata can turn a semantic error into a Rust panic instead of a deterministic `FrontendError`.

## Required repair

Validate `FnSig` metadata invariants before call reordering/typechecking, or make inconsistent construction impossible through a checked constructor/private fields. Required invariant at minimum:

```text
param_names.len == params.len   when Some
param_defaults.len == params.len when Some
```

## Regression

- short/long `param_names` returns FrontendError, never panic;
- short/long defaults return FrontendError;
- canonical tables continue working;
- public `type_check_function_with_table` fails closed on malformed signatures.

### LINKED ISSUE #1722: FA-04-016: named-argument lowering evaluates in parameter order instead of source order
Umbrella: #1617
Module: 04 — sm-ir
Severity: P1
Classification: SEMANTIC-DRIFT · FALSE-READY

Finding: ordinary call lowering first calls `reorder_call_args(...)`, which rearranges named arguments into the callee's declared parameter order, and only then iterates that reordered vector and lowers/evaluates each argument. Thus source `f(b: expr_b, a: expr_a)` for declaration `f(a, b)` evaluates `expr_a` before `expr_b`.

`docs/spec/source_semantics.md` separately freezes deterministic evaluation order as “call arguments are evaluated left-to-right”. The difference is observable when argument expressions can trap or perform admitted effects/calls.

The named-argument section says names are reordered to declared parameter order before lowering, but that conflicts with the general source evaluation-order guarantee. The implementation currently chooses parameter-order evaluation rather than preserving source-order evaluation plus parameter-slot mapping.

Cross-phase impact:
source argument order → frontend accepts named call → sm-ir reorders ExprIds → evaluation/trap/effect order changes.

Evidence:
- crates/sm-ir/src/legacy_lowering.rs call expression and call statement paths using `reorder_call_args` before `lower_expr_with_expected`
- crates/sm-front/src/lib.rs::reorder_call_args
- docs/spec/source_semantics.md Deterministic Evaluation Order and named-argument sections

skulmakov-oss:
SSF-07: #1578

Fixes #1722
Fixes #1665

Baseline SHA: cca4aa57baf0a2d8ba4e68dc4ad50470a1147801

## Problem

**#1722 (FA-04-016):** ordinary call lowering called `reorder_call_args(...)` to
rearrange named arguments into declared-parameter order, then lowered/evaluated
the *reordered* vector directly. So `f(b: expr_b, a: expr_a)` for `fn f(a, b)`
evaluated `expr_a` before `expr_b` — contradicting the frozen
"call arguments are evaluated left-to-right" rule in
`docs/spec/source_semantics.md`. Both independent lowering paths
(call-expression, call-statement) in `crates/sm-ir/src/legacy_lowering.rs` had
this bug.

**#1665 (FA-02-033):** the public `FnSig` struct exposes `params`,
`param_names`, and `param_defaults` as independently constructible fields, and
`type_check_function_with_table(...)` accepts a caller-supplied public
`FnTable`. `reorder_call_args`/`finalize_ordered_call_args` indexed
`param_names[idx]` / `ordered[param_index]` without validating their lengths
against `params`, so a malformed externally-constructed `FnSig` could panic
instead of failing closed with a `FrontendError`.

- **#1722**, real pipeline (source → frontend → IR → emit → verify → VM): a
  call `add(b = ten / zero, a = min_val % neg_one)` — where `b`'s expression
  traps `DivisionByZero` and `a`'s expression traps `ArithmeticOverflow` — is
  expected to trap `DivisionByZero` first (b is written first in source).
  Pre-fix, it trapped `ArithmeticOverflow` instead (a evaluated first, matching
  parameter order, not source order). Same result for both the call-expression
  and call-statement paths. At the IR level, `scale(factor = 3.0, x = 2.0)`
  emitted `LoadF64(2.0)` before `LoadF64(3.0)` pre-fix (parameter order);
  should be `LoadF64(3.0)` before `LoadF64(2.0)` (source order).
- **#1665**: with `param_names` shortened below `params.len()` and a call
  missing an argument, `finalize_ordered_call_args` panicked
  `index out of bounds: the len is 1 but the index is 1` indexing
  `param_names[idx]`. With `param_names` lengthened beyond `params.len()` and
  a named call referencing the extra name, `reorder_call_args` panicked
  `index out of bounds: the len is 2 but the index is 2` indexing
  `ordered[param_index]`.

Both were verified to fail exactly this way by temporarily reverting the fix
in place, rerunning the new regression tests, and confirming the specific
failure/panic, then restoring the fix and confirming green.

## Root cause

Both defects share one root cause: `reorder_call_args`'s public contract
conflated two independent concerns — *which parameter slot an argument value
lands in* (named-argument resolution) and *evaluation order* (source position)
— and separately, never validated that its two independently-`pub` optional
metadata vectors (`param_names`, `param_defaults`) were consistent in length
with `params` before indexing them.

## Frozen invariants

- Call argument expressions are evaluated left-to-right in **source** order
  (`docs/spec/source_semantics.md`, Deterministic Evaluation Order). Named
  arguments determine destination parameter slot only; they must never reorder
  expression evaluation.
- `FnSig.param_names.len() == FnSig.params.len()` and
  `FnSig.param_defaults.len() == FnSig.params.len()` whenever those optional
  fields are `Some`, enforced before any indexing depends on them.

## Repair design

- `reorder_call_args` now returns `OrderedCallArgs { slots: Vec<ExprId>,
  eval_order: Vec<usize> }` instead of a bare `Vec<ExprId>`. `slots` keeps the
  existing per-parameter-slot semantics (defaults substituted). `eval_order`
  lists parameter-slot indices in the order their expressions must be
  *evaluated*: explicitly written arguments in source order, then defaulted
  slots (not source-written; const-safe by existing contract) appended last in
  parameter order.
- Both `crates/sm-ir/src/legacy_lowering.rs` call-lowering paths now iterate
  `eval_order` to decide lowering/evaluation order, writing each resulting
  register into its `slots` index, then flatten into the final parameter-order
  register list for `IrInstr::Call`.
- `crates/sm-front/src/typecheck.rs`'s three (type-checking only, no
  evaluation-order concern) call sites were updated to read
  `ordered_args.slots` instead of the old bare vector — no behavior change.
- New `validate_fn_sig_call_metadata` runs first, unconditionally, at the top
  of `reorder_call_args` (the single choke point every caller of a `FnSig`
  reaches before any `param_names`/`param_defaults` indexing), rejecting any
  length mismatch with a `FrontendError` before `finalize_ordered_call_args`
  can index either vector.
- `docs/spec/source_semantics.md`'s named-argument section was corrected: it
  previously described reordering as happening "before ordinary argument
  type-checking and lowering" in a way that read as authorizing parameter-order
  evaluation. It now explicitly states named arguments reorder slot assignment
  only, per the frozen Deterministic Evaluation Order rule above it, and that
  default initializers (not source-written) evaluate after explicit arguments.

## Why this is fail-closed

- Positional-only calls are unaffected: `eval_order` reduces to `0..params.len()`
  when there are no named arguments, byte-for-byte the same order as before.
- The malformed-`FnSig` fix rejects with a structured `FrontendError` before
  any indexing occurs — there is no code path left where a length-mismatched
  `param_names`/`param_defaults` reaches `finalize_ordered_call_args`'s
  indexing logic.
- No new default, guess, or plausible-substitute behavior was introduced for
  any unknown/malformed state — malformed metadata is rejected, never
  silently truncated, extended, or fabricated.

## Fallback/bypass audit

Reviewed the full changed control-flow neighborhood in
`crates/sm-front/src/lib.rs`, `crates/sm-front/src/typecheck.rs`, and
`crates/sm-ir/src/legacy_lowering.rs`:

- No `unwrap()`/`expect()`/`unwrap_or*` introduced. The existing
  `.get(idx)...flatten()` pattern for default lookup is unchanged (already
  bounds-safe; kept as-is).
- The `ordered.into_iter().flatten()` / new `regs.into_iter().flatten()`
  patterns rely on `eval_order` being a permutation of `0..slots.len()` by
  construction (every slot is filled by construction: explicit args plus
  finalize's default-fill loop), not on masking an error — matches the
  existing `finalize_ordered_call_args` idiom already used in this file.
- No `if let Ok/Some` silent-continuation branches added.
- No catch-all match arms added.
- The one observable behavior change (malformed-`FnSig` positional calls that
  previously silently succeeded using inconsistent metadata now reject) is the
  intended fail-closed repair, not a new fallback.

## Tests added

- `crates/sm-ir/src/legacy_lowering.rs`:
  `lower_named_arguments_evaluates_in_source_order_not_parameter_order`
  (call-expression path, IR instruction order),
  `lower_named_arguments_evaluates_in_source_order_for_call_statement`
  (call-statement path, IR instruction order),
  `lower_mixed_named_arguments_and_trailing_default_evaluates_in_source_order`
  (named reordering + trailing default interaction). Strengthened the
  existing `lower_named_arguments_to_ordinary_call_order` test's coverage
  scope is unchanged (kept, not weakened); the new tests assert actual order.
- `crates/sm-vm/src/semcode_vm.rs`:
  `vm_evaluates_named_call_expression_arguments_in_source_order`,
  `vm_evaluates_named_call_statement_arguments_in_source_order` — full
  pipeline (source → frontend → IR → emit → verify → VM), asserting the
  specific `RuntimeTrap` kind that fires, which is only possible if evaluation
  order is source order.
- `crates/sm-front/src/typecheck.rs`:
  `malformed_fn_sig_param_names_shorter_than_params_fails_closed`,
  `malformed_fn_sig_param_names_longer_than_params_fails_closed`,
  `malformed_fn_sig_param_defaults_shorter_than_params_fails_closed`,
  `malformed_fn_sig_param_defaults_longer_than_params_fails_closed`, and
  `canonical_fn_sig_continues_to_typecheck_via_public_table_api` (well-formed
  `FnSig` via the public `FnTable` API is unaffected), all exercising
  `type_check_function_with_table`.

All new tests were confirmed to fail (two of them via literal panic,
reproducing the exact issue text) against the pre-fix code by temporarily
reverting the production changes in place, then confirmed to pass after
restoring the fix.

```
cargo test -p sm-front            -> 488 passed; 0 failed
cargo test -p sm-ir               -> 116 passed; 0 failed
cargo test -p sm-vm               -> 116 passed; 0 failed
cargo test --test call_shape_surface_qualification -> 1 passed
cargo test --test public_api_contracts             -> 88 passed (incl. public_api_inventory_matches_checked_in_contract_snapshots — no golden-snapshot update needed)
cargo clippy -p sm-front -p sm-ir --all-targets -- -D warnings -> clean
cargo fmt -p sm-front -p sm-ir -p sm-vm --check                -> clean
```

### Environment/pre-existing baseline caveats (not fixed by this PR, by design)

- Workspace-wide `cargo fmt --check` and `cargo clippy --workspace
  --all-targets -- -D warnings` fail in this local environment with a Windows
  path-length OS error (`Имя файла или его расширение имеет слишком большую
  длину`, i.e. "file name or extension too long") unrelated to any file this
  PR touches — a local/environment limitation, not a code issue.
- `cargo clippy -p sm-vm --all-targets -- -D warnings` fails on exactly one
  pre-existing `dead_code` warning: `VmProgramView.header` is never read
  (`crates/sm-vm/src/semcode_vm.rs:1017`). Confirmed via `git diff origin/main
  -- crates/sm-vm/src/semcode_vm.rs` that this PR's only change to that file
  is the two new tests appended after line 4806 — nowhere near the flagged
  struct — and the field is identical, unused, on `origin/main` before this
  PR. Left unfixed per scope discipline; not registered as a new finding here
  since it is pre-existing platform debt outside #1722/#1665.

CTF touched: none

## Out of scope

- The `VmProgramView.header` dead-code warning above.
- The Windows path-length limitation on workspace-wide `fmt`/`clippy`
  invocations in this environment.
- Any other SSF-07 abstraction (numeric/text/collection/closure/generic/
  trait/pattern policy) — this PR is Repair Slice 1 only.
- No global AST/IR redesign, no VM changes, no SemCode format changes, no
  unrelated renaming or cleanup.

## Remaining SSF-07 work

This PR does **not** close #1578. After this slice merges, SSF-07 continues
with the live closure matrix (numeric/text/collection/closure/generic/trait/
pattern policy) per #1578's acceptance criteria, selected by severity →
semantic/trust impact → dependency order → smallest coherent repair radius.

### DIFF
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -977,12 +977,68 @@ pub fn builtin_sig(name: &str) -> Option<FnSig> {
 }
 
 #[cfg(any(feature = "alloc", feature = "std"))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OrderedCallArgs {
+    /// Final argument expression for each declared parameter slot, in
+    /// parameter order (defaults substituted). Used for per-slot type
+    /// checking and for the final argument-register order handed to the
+    /// callee.
+    pub slots: Vec<ExprId>,
+    /// Parameter slot indices in the order their expressions must be
+    /// *evaluated*: explicitly supplied arguments in source (left-to-right)
+    /// order, followed by defaulted slots in parameter order.
+    ///
+    /// Named-argument slot assignment and source evaluation order are
+    /// separate concerns: named arguments only ever reorder `slots`, never
+    /// `eval_order`'s relative order of explicitly written argument
+    /// expressions.
+    pub eval_order: Vec<usize>,
+}
+
+#[cfg(any(feature = "alloc", feature = "std"))]
+fn validate_fn_sig_call_metadata(
+    call_name: SymbolId,
+    sig: &FnSig,
+    arena: &AstArena,
+) -> Result<(), FrontendError> {
+    if let Some(param_names) = sig.param_names.as_ref() {
+        if param_names.len() != sig.params.len() {
+            return Err(FrontendError {
+                pos: 0,
+                message: format!(
+                    "function '{}' has malformed signature: {} parameter name(s) declared for {} parameter(s)",
+                    resolve_symbol_name(arena, call_name)?,
+                    param_names.len(),
+                    sig.params.len(),
+                ),
+            });
+        }
+    }
+    if let Some(param_defaults) = sig.param_defaults.as_ref() {
+        if param_defaults.len() != sig.params.len() {
+            return Err(FrontendError {
+                pos: 0,
+                message: format!(
+                    "function '{}' has malformed signature: {} parameter default(s) declared for {} parameter(s)",
+                    resolve_symbol_name(arena, call_name)?,
+                    param_defaults.len(),
+                    sig.params.len(),
+                ),
+            });
+        }
+    }
+    Ok(())
+}
+
+#[cfg(any(feature = "alloc", feature = "std"))]
 pub fn reorder_call_args(
     call_name: SymbolId,
     args: &[CallArg],
     sig: &FnSig,
     arena: &AstArena,
-) -> Result<Vec<ExprId>, FrontendError> {
+) -> Result<OrderedCallArgs, FrontendError> {
+    validate_fn_sig_call_metadata(call_name, sig, arena)?;
+
     let has_named = args.iter().any(|arg| arg.name.is_some());
     if !has_named {
         if args.len() > sig.params.len() {
@@ -997,10 +1053,12 @@ pub fn reorder_call_args(
             });
         }
         let mut ordered = vec![None; sig.params.len()];
+        let mut eval_order = Vec::with_capacity(args.len());
         for (idx, arg) in args.iter().enumerate() {
             ordered[idx] = Some(arg.value);
+            eval_order.push(idx);
         }
-        return finalize_ordered_call_args(call_name, ordered, sig, arena, args.len());
+        return finalize_ordered_call_args(call_name, ordered, sig, arena, args.len(), eval_order);
     }
 
     let Some(param_names) = sig.param_names.as_ref() else {
@@ -1014,6 +1072,7 @@ pub fn reorder_call_args(
     };
 
     let mut ordered = vec![None; sig.params.len()];
+    let mut eval_order = Vec::with_capacity(args.len());
     let mut positional_index = 0usize;
     let mut named_seen = false;
     for arg in args {
@@ -1040,6 +1099,7 @@ pub fn reorder_call_args(
                 });
             }
             ordered[param_index] = Some(arg.value);
+            eval_order.push(param_index);
         } else {
             if named_seen {
                 return Err(FrontendError {
@@ -1059,11 +1119,12 @@ pub fn reorder_call_args(
                 });
             }
             ordered[positional_index] = Some(arg.value);
+            eval_order.push(positional_index);
             positional_index += 1;
         }
     }
 
-    finalize_ordered_call_args(call_name, ordered, sig, arena, args.len())
+    finalize_ordered_call_args(call_name, ordered, sig, arena, args.len(), eval_order)
 }
 
 #[cfg(any(feature = "alloc", feature = "std"))]
@@ -1073,7 +1134,8 @@ fn finalize_ordered_call_args(
     sig: &FnSig,
     arena: &AstArena,
     provided_count: usize,
-) -> Result<Vec<ExprId>, FrontendError> {
+    mut eval_order: Vec<usize>,
+) -> Result<OrderedCallArgs, FrontendError> {
     let param_names = sig.param_names.as_ref();
     let param_defaults = sig.param_defaults.as_ref();
     for idx in 0..ordered.len() {
@@ -1086,6 +1148,7 @@ fn finalize_ordered_call_args(
             .flatten();
         if let Some(default_expr) = default_expr {
             ordered[idx] = Some(default_expr);
+            eval_order.push(idx);
             continue;
         }
         if let Some(param_names) = param_names {
@@ -1108,7 +1171,10 @@ fn finalize_ordered_call_args(
             ),
         });
     }
-    Ok(ordered.into_iter().flatten().collect())
+    Ok(OrderedCallArgs {
+        slots: ordered.into_iter().flatten().collect(),
+        eval_order,
+    })
 }
 
 #[cfg(any(feature = "alloc", feature = "std"))]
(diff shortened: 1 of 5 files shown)


skulmakov-oss:
@codex review

Issue #1667, opened 2026-08-14
FA-02-035: trait/impl Self is hard-coded as Record and misrepresents enum impl targets


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P1 — trait Self can acquire the wrong nominal runtime/source type**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

The current static trait surface admits impl targets by nominal `TypeName`, and bound matching already recognizes both `Type::Record(id)` and `Type::Adt(id)` as concrete impl targets.

However two frontend paths hard-code `Self` to record form:

1. `parse_impl_decl()` enters impl method parsing with:

```rust
with_self_type_scope(Type::Record(for_type), ...)
```

2. `substitute_trait_self_type()` replaces trait-side `Self` with:

```rust
Type::Record(concrete_self)
```

regardless of whether the actual declared target is a record or enum/ADT.

## Risk

For `impl Trait for SomeEnum`, `Self` in method parameters/returns is represented and compared as `Record(SomeEnum)`. Conformance can therefore reject valid enum-Self signatures, accept a wrong record-shaped interpretation, or create later phase disagreement.

## Required repair

Resolve the impl target through the canonical record/ADT tables first and carry the actual concrete `Type` into parser/typecheck Self substitution. Do not reconstruct target kind from a bare SymbolId with a Record fallback.

This repair should compose with FA-02-019 (target existence proof) rather than create a second nominal-type authority.

## Regression

- `Self` for record impl resolves to `Type::Record`;
- `Self` for enum/ADT impl resolves to `Type::Adt`;
- unknown target rejects;
- nested `Self` in tuple/sequence/Option/Result/closure signatures substitutes recursively with the correct nominal kind.

Issue #1654, opened 2026-08-14
FA-02-022: qtruth_* user functions can be admitted then replaced by intrinsic lowering


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P1 — admitted user function can lower to different semantics**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

The frontend function table does not reserve `qtruth_and`, `qtruth_or`, `qtruth_not`, or `qtruth_impl` as user-function collisions.

The IR lowering path resolves the call name and, when it matches one of those strings, emits dedicated `IrInstr::QTruth*` instructions directly instead of lowering an ordinary function call.

Therefore a user function can be admitted under one of these names while call lowering selects the intrinsic and never calls the user body.

## Risk

```text
user fn qtruth_and(...) admitted
→ call name reaches lowering
→ dedicated QTruthAnd emitted
→ user function semantics disappear
```

This is silent wrong-code at the frontend/lowering ownership boundary.

## Required repair

Make intrinsic namespace ownership explicit. Either reserve all qtruth intrinsic names during frontend function-table construction or implement a name-resolution model where an admitted user function unambiguously prevents intrinsic selection.

## Regression

- qtruth intrinsic calls still lower to explicit QTruth IR when intended;
- same-name user declarations cannot be silently shadowed;
- frontend name admission and IR intrinsic resolution share one authoritative collision policy.

Issue #1652, opened 2026-08-14
FA-02-020: untyped `let _ = expr` can skip RHS typechecking entirely


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P1 — invalid expression can bypass frontend checking**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

The source contract states that `let _ = expr;` evaluates its right-hand side and discards the value.

In `check_stmt`, the `Stmt::Discard` path typechecks the RHS only when an explicit type annotation is present. For the ordinary untyped form (`ty == None`), the branch can return success without calling `infer_expr_type` on `value`.

## Risk

Wrapping an otherwise invalid expression in an untyped discard can bypass frontend expression validation:

```text
invalid expr
→ `let _ = invalid_expr;`
→ no infer/typecheck
→ statement accepted
```

That violates the statement semantics and creates a phase-consistency hole.

## Required repair

Always infer/check the RHS exactly once for discard binds. If an explicit type annotation exists, additionally enforce that expected type; without one, still require the expression itself to be valid.

## Regression

- valid `let _ = expr;` remains admitted;
- an invalid/unknown/wrong-arity RHS rejects identically whether discarded or used;
- annotated discard keeps its type constraint;
- RHS is not evaluated/checked twice.

PR #1856, opened 2026-08-31
fix(front): reject duplicate match default arms (#1637 #1638)

### LINKED ISSUE #1637: FA-02-005: statement match accepts duplicate default arms and silently keeps the last one
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P1 — silent wrong-code source resolution**

## Finding

Statement `match` parsing stores the wildcard arm as `default = Some(block)` with no duplicate check. A second `_ => { ... }` simply replaces the first.

## Risk

Contradictory source is converted into a valid AST by last-write-wins semantics. The earlier default body disappears with no diagnostic.

## Required repair

Reject a second statement wildcard/default arm at parse time.

## Regression

- one wildcard arm passes;
- second wildcard arm rejects;
- no accepted match silently loses an arm.

### LINKED ISSUE #1638: FA-02-006: match expression accepts duplicate default arms and silently keeps the last one
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P1 — silent wrong-code source resolution**

## Finding

Value-producing `match` expression parsing also assigns each `_ => { ... }` to `default = Some(block)` without detecting that a default already exists. Later defaults overwrite earlier ones.

## Risk

Accepted source silently loses an earlier branch and executes a different default body than written first.

## Required repair

Reject duplicate wildcard/default arms for match expressions.

## Regression

- exactly one default passes;
- duplicate default rejects deterministically;
- statement and expression forms follow the same conflict rule.

skulmakov-oss:
SSF-07: #1578

Fixes #1637
Fixes #1638

Baseline SHA: 4ba2a89e96e587e24fecac4067990c1cf74f7ea1 (main, includes the #1722/#1665 and #1710 slices)

## Problem

Both the statement-form and expression-form `match` parsers in
`crates/sm-front/src/parser.rs` accept an unbounded number of `_` default
arms and silently keep only the last one:

- `parse_match_stmt_after_kw_match` (statement match, #1637 / FA-02-005):
  `default: Option<Vec<StmtId>>`, unconditionally does `default =
  Some(block);` for every `_` arm encountered.
- `parse_match_expr_after_kw_match` (expression match, #1638 / FA-02-006):
  `default: Option<BlockExpr>`, same unconditional overwrite.

Neither site checked whether `default` was already `Some(...)` before
overwriting it, so contradictory source (two `_` arms) silently resolved to
"keep the last one" instead of being rejected — a fail-open parsing gap on
directly conflicting syntax.

## Shared root cause

Both parsing functions are structurally identical for this concern: each
loop iteration that matches `_` unconditionally assigns into a local
`Option<...>` accumulator with no prior-state check. This is the same root
pattern in two independent (but nearly identical) functions — not one shared
helper's misuse — so this is one combined repair slice per the shared
invariant, not a coincidence.

Verified directly against baseline `4ba2a89e` by temporarily reverting both
checks and rerunning the exact regression tests below (`cargo test -p
sm-front`), then restoring the fix. Concrete AST evidence:

**#1638 (expression match)** — `match T { _ => { 1.0 } _ => { 2.0 } }` inside
a `let value: f64 = ...;` parses successfully pre-fix. The panic output from
the reverted `rustlike_parser_rejects_duplicate_default_match_expression_arm`
test dumps the full parsed `Program`, showing:

```
Match(MatchExpr { scrutinee: ExprId(0), arms: [],
  default: Some(BlockExpr { statements: [], tail: ExprId(2) }) })
```

`ExprId(2)` is the *second* literal (`2.0`); `ExprId(1)` (the first
literal, `1.0`) never appears anywhere in `default` — it is fully discarded,
not merely shadowed.

**#1637 (statement match)** — using distinguishable arm bodies
(`_ => { continue; }` then `_ => { break; }`, verified with a temporary probe
test not included in this PR) parses successfully pre-fix, and the resulting
`Stmt::Match.default` is exactly `[Break(None)]` — the second arm's body.
The first arm's `continue;` is never reachable in the accepted AST.

Both confirm: **the parser successfully accepts duplicate default arms, and
the resulting AST contains only the second default body** — the two
strongest of the three acceptable proof forms named in the issues.

## Frozen invariant

A `match` construct (statement or expression form) may contain at most one
`_` wildcard/default arm. A second `_` arm is a deterministic
`FrontendError` at parse time. The first parsed default arm is authoritative
and is never overwritten, replaced, merged with, or silently discarded.

## Repair design

Added the identical check, with the identical diagnostic
(`"match cannot have more than one default '_' arm"`), at both parsing
sites, immediately before each site's own `default = Some(block)`
assignment:

```rust
if default.is_some() {
    return Err(FrontendError {
        pos: self.pos(),
        message: "match cannot have more than one default '_' arm".to_string(),
    });
}
```

Kept as two local checks rather than one shared helper: the check itself is
a two-line `if`, and the two call sites hold different `Option<T>` inner
types (`Vec<StmtId>` vs `BlockExpr`) with different surrounding control flow
(the `default.unwrap_or_default()` shape at the statement site vs. the
`Option<BlockExpr>` kept as-is at the expression site) — a generic helper
would add an abstraction layer to remove two lines of genuinely
straightforward, self-explanatory duplication. The shared, load-bearing
piece — the exact wording of the diagnostic — is kept textually identical by
construction, satisfying "statement and expression forms follow one rule"
without a forced shared function.

No other parser profile or syntax variant reaches this code:
`grep -rln "MatchArm\|MatchExprArm"` across `crates/sm-front/src` returns
only `parser.rs` (the implementation, now fixed), `types.rs` (AST
definitions), and `lib.rs` (re-exports) — `match` has no separate Logos-DSL
parsing path, so there is no sibling exposure to patch.

## Why this is fail-closed

- The only new behavior is rejection with a structured `FrontendError`;
  nothing is silently coerced, defaulted, or merged.
- Zero default arms and exactly one default arm are untouched — same parse
  result as before this change (proven by the positive-control tests below).
- The rejection happens at parse time, strictly before typecheck, lowering,
  emission, verification, or VM execution ever see the program.

## Fallback/bypass audit

Searched the changed neighborhood and the wider `parser.rs` for the same
"assign into `Option`/map with no prior-state check" shape
(`grep -n "= Some("`):

- `default = Some(block)` (both match-arm sites) — **SAME ROOT**, this PR.
- `self.self_type_scope = Some(self_ty)` — scope-stack push (paired with a
  restore later in the same function), not a duplicate-conflict pattern —
  **SAFE**.
- `out.system = Some(system)` (Logos `System` declaration parsing) — same
  root *pattern* (unconditional `Option` overwrite on a duplicate
  declaration) but a different construct entirely, and already tracked as
  its own finding, **FA-02-012 / #1644** ("duplicate Logos System
  declarations silently use last-write-wins semantics") — **NEW FINDING,
  already filed, intentionally left untouched**, per this PR's scope
  discipline.
- All other `= Some(...)`/`== Some(...)` occurrences in `parser.rs` are
  lookahead comparisons (`peek_next_kind() == Some(...)`), unrelated to
  overwrite/duplication concerns — **SAFE**.

One representational note surfaced while writing the positive-control tests,
not a defect: `Stmt::Match.default` is stored as a bare `Vec<StmtId>` (not
`Option<Vec<StmtId>>`), so an *explicitly present but empty* default block
(`_ => { }`) is indistinguishable from *no default arm at all* by inspecting
the stored AST alone. This has no observable behavioral consequence (both
mean "no default statements execute") and is unrelated to the duplicate-arm
defect this PR fixes, so it is not addressed here; documented as a comment
on the affected test (`rustlike_parser_accepts_match_statement_with_one_default_arm`)
for future readers.

## Tests added

`crates/sm-front/src/parser.rs` (parser-level, direct AST evidence):

- `rustlike_parser_accepts_match_expression_with_no_default_arm` (positive control)
- `rustlike_parser_rejects_duplicate_default_match_expression_arm`
- `rustlike_parser_accepts_match_statement_with_no_default_arm` (positive control)
- `rustlike_parser_accepts_match_statement_with_one_default_arm` (positive control)
- `rustlike_parser_rejects_duplicate_default_match_statement_arm`

(`rustlike_parser_accepts_match_expression` and
`rustlike_parser_accepts_adt_match_expression_surface`, already in the
suite, cover "exactly one default arm" for the expression form and continue
to pass unchanged.)

`tests/match_surface_qualification.rs` (higher-level regression, full
`smc check` pipeline) plus two new fixtures:

- `examples/qualification/match_surface/negative_duplicate_default_expression_arm_rejected/src/main.sm`
- `examples/qualification/match_surface/negative_duplicate_default_statement_arm_rejected/src/main.sm`

registered in `match_surface_negative_fixtures_reject_deterministically`,
asserting the CLI's `check` command surfaces the exact diagnostic.

## Pre-fix failure evidence

All five new parser-level tests were run against the pre-fix code
(production checks temporarily reverted in place, tests unchanged) and
confirmed to fail exactly as expected:

- Both `rustlike_parser_rejects_duplicate_default_match_*_arm` tests failed
  because parsing unexpectedly **succeeded** (`expect_err` panicked with the
  full parsed `Program` dumped — see the AST evidence quoted above).
- The two positive-control "no default arm" tests were unaffected (parsing
  behavior for zero defaults was never buggy).

Then the fix was restored and the full suite reconfirmed green.

## Post-fix qualification results (exact commands, this HEAD)

```
cargo check --workspace --all-targets                       -> clean
cargo test -p sm-front                                        -> 493 passed; 0 failed
cargo test --test match_surface_qualification                -> 3 passed; 0 failed
cargo test --test public_api_contracts                        -> 88 passed (no signature/snapshot drift; both fixed functions are private)
cargo test --test canonical_source_style --test ssf_language_contract_drift -> 21 passed
cargo fmt -p sm-front --check                                 -> clean
cargo clippy -p sm-front --all-targets -- -D warnings         -> clean
cargo clippy --workspace --all-targets -- -D warnings         -> clean
```

`cargo fmt --check` (workspace-wide, no `-p`) still fails in this local
environment on the same unrelated Windows path-length OS error noted in the
prior two SSF-07 slices — confirmed unrelated (this PR touches no path
anywhere near whatever triggers it). Note that this run,
`cargo clippy --workspace --all-targets -- -D warnings` itself passed clean
(feature unification across the full workspace build graph apparently
enables whatever gates the previously-reported pre-existing
`VmProgramView.header` dead-code warning in `sm-vm`); running
`cargo clippy -p sm-vm --all-targets -- -D warnings` in isolation still
surfaces that same pre-existing, untouched-by-this-PR warning. This PR does
not touch `crates/sm-vm` at all.

Changed files: `crates/sm-front/src/parser.rs`,
`docs/spec/source_semantics.md`, `tests/match_surface_qualification.rs`,
`examples/qualification/match_surface/negative_duplicate_default_expression_arm_rejected/src/main.sm`
(new),
`examples/qualification/match_surface/negative_duplicate_default_statement_arm_rejected/src/main.sm`
(new).

CTF touched: none

## Out of scope

Per the assigned slice boundary — none of the following are touched:
#1636 (duplicate function parameters), #1652 (discard RHS typechecking),
#1654/#1721 (intrinsic namespace collisions), #1667 (trait `Self`), #1710
(CrystalFold — already merged in a prior slice), ownership fixes,
module/import fixes, verifier/runtime fixes, exhaustiveness redesign, new
pattern syntax/forms, async/macros, broad AST cleanup, or unrelated
warnings/refactors. Also explicitly out of scope: FA-02-012 / #1644 (the
same-root Logos `System` sibling identified in the fallback/bypass audit
above) and the `Stmt::Match.default` empty-vs-absent representational
ambiguity noted above — both are independent from this fix and are reported
rather than repaired here.

## Remaining SSF-07 work

This PR does **not** close #1578 and does not complete the wider SSF-07 /
Phase-B repair program. Do not begin SSF-08 work as a result of this PR.

### DIFF
--- a/crates/sm-front/src/parser.rs
+++ b/crates/sm-front/src/parser.rs
@@ -2418,6 +2418,12 @@ impl<'a> Parser<'a> {
                         message: "default '_' arm in match currently cannot have guard".to_string(),
                     });
                 }
+                if default.is_some() {
+                    return Err(FrontendError {
+                        pos: self.pos(),
+                        message: "match cannot have more than one default '_' arm".to_string(),
+                    });
+                }
                 self.expect(TokenKind::FatArrow, "expected '=>' after '_'")?;
                 let block = self.parse_block()?;
                 default = Some(block);
@@ -2448,6 +2454,12 @@ impl<'a> Parser<'a> {
                     return Err(FrontendError {
                         pos: self.pos(),
                         message: "default '_' arm in match currently cannot have guard".to_string(),
+                    });
+                }
+                if default.is_some() {
+                    return Err(FrontendError {
+                        pos: self.pos(),
+                        message: "match cannot have more than one default '_' arm".to_string(),
                     });
                 }
                 self.expect(TokenKind::FatArrow, "expected '=>' after '_'")?;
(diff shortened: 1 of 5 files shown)


skulmakov-oss:
@codex review

Issue #1729, opened 2026-08-15
FA-04-023: CrystalFold annihilator rules can erase runtime type-mismatch traps on admitted raw IR


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P1
Classification: SEMANTIC-DRIFT · FAIL-OPEN · FALSE-READY

Finding: CrystalFold applies one-sided annihilator rules without proving the other operand's runtime type. Examples:

- `BoolAnd`: if either tracked operand is constant `false`, replace with `LoadBool false`;
- `BoolOr`: if either tracked operand is constant `true`, replace with `LoadBool true`;
- `QAnd` / `QOr` have analogous one-sided quad annihilator rewrites.

A verifier-admitted raw IR sequence such as:

```text
LoadBool r1, false
LoadI32  r2, 1
BoolAnd  r3, r1, r2
```

would reach a VM runtime type mismatch without O1, because `BoolAnd` requires bool operands. CrystalFold instead replaces the operation with a constant bool result and removes that failure edge.

The public `validate_ir`/SemCode verifier validate structural/register shape, not dynamic register value families, so this cannot be justified as an unreachable invalid artifact at the public IR/pass boundary.

This is separate from FA-04-004 signed-zero drift: here an explicit runtime rejection becomes successful execution.

Evidence:
- `crates/sm-ir/src/passes/crystalfold.rs` BoolAnd/BoolOr/QAnd/QOr annihilator branches
- sm-vm typed opcode execution (`as_bool` / quad value expectations)
- sm-verify structural operand checks without register-family proof
- `docs/spec/ir.md` CrystalFold semantics-preserving pass contract

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
Additional same-root proof: `const_eq()` returns `false` for any mixed `ConstVal` families, while VM `value_eq()` returns `TypeMismatchRuntime("CmpEq/CmpNe operands must have same runtime type")` for mixed runtime types. Thus constant `CmpEq/CmpNe` on structurally admitted mixed scalar IR is another CrystalFold rejection→success path. Kept under FA-04-023.

Issue #1717, opened 2026-08-15
FA-04-011: completed generics track claims IR monomorphisation but sm-ir contains no TypeVar/monomorphisation path


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P2
Classification: FALSE-READY · INERT-CONTRACT · QUALIFICATION-GAP

Finding: `docs/roadmap/language_maturity/generics_full_scope.md` is marked completed and its Wave 3/acceptance boundary explicitly requires an IR monomorphisation pass plus lowering of generic definitions to concrete SemCode paths. The audited sm-ir production sources contain no monomorphisation pass and no TypeVar handling. The compile path lowers every program function directly after type_check_program.

The sm-front typechecker itself still comments that generic body checking is deferred until Wave 3 monomorphisation, so this is not merely a naming difference.

Gap: a completed roadmap acceptance claim names a lowering phase that does not exist in its canonical owner crate.

Evidence:
- docs/roadmap/language_maturity/generics_full_scope.md
- crates/sm-ir/src/legacy_lowering.rs (no TypeVar/monomorphisation path)
- crates/sm-ir/src/passes/* (only StructuralCleanup and CrystalFold)
- sm-front typecheck generic deferral comments

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
Lifecycle correction, no new code: this finding is already repaired by merged PR #1873 (squash/merge commit \`17d4b7dc255bf3cf96685dc33c8e4b48118e7a19\`), part of the SSF-07 residual-FA reconciliation. Never closed at the time.

**Original finding**: `docs/roadmap/language_maturity/generics_full_scope.md` claimed a completed IR monomorphisation pass (Wave 3), but `crates/sm-ir/src/legacy_lowering.rs` had zero `type_params` handling — worse, a generic function with an *unused* type parameter (`fn marker<T>(x: i32) -> i32`) silently lowered as an ordinary, non-generic `IrFunction`, discarding `type_params` entirely (genuine type erasure, not partial admission).

[...]

Issue #1669, opened 2026-08-14
FA-02-037: trait method signature types are not independently canonicalized or validated


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — invalid trait contract can enter TraitTable**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

`build_trait_table()` validates duplicate trait names and then clones each `TraitDecl` into the table. `type_check_program()` runs trait coherence and impl conformance, but there is no standalone pass that canonicalizes/resolves every parameter/return type in trait method signatures against the admitted record/ADT/type surface.

An unknown nominal type in a trait method signature can therefore remain in a trait declaration when no impl forces a later comparison/body check to expose it.

## Risk

The canonical trait contract table can contain unresolved type identities and still be treated as an admitted source-level abstraction contract.

## Required repair

Add owner-layer trait declaration validation using the same canonical nominal type authority as function/record/ADT declarations, with the explicit narrow `Self` placeholder exception. Unknown or unsupported method-signature types must reject even if no impl exists.

## Regression

- trait signature using admitted scalar/record/ADT types passes;
- reserved `Self` works only in the admitted trait positions;
- unknown nominal parameter/return type rejects without requiring an impl;
- TraitTable never contains an unresolved successful contract.

Issue #1668, opened 2026-08-14
FA-02-036: generic impl parameters are admitted although first-wave ImplDecl requires a concrete non-generic target


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — unsupported generic impl surface accepted**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

`parse_impl_decl()` accepts an optional type-parameter list and stores it in `ImplDecl.type_params`.

The owner-layer contract on `ImplDecl` states that `type_params` are empty in first-wave canonical form, and the completed trait scope lists blanket/generic impl forms outside the current admitted surface.

Current coherence/conformance keys remain the concrete `(trait_name, for_type)` pair and do not implement generic impl semantics.

## Risk

Unsupported generic impl syntax can enter the canonical AST and be processed by concrete-impl validation as though the type parameters had no semantic effect.

## Required repair

Reject non-empty impl type parameters in the current trait surface, or explicitly widen and implement generic/blanket impl semantics before admission. Do not parse/store future-only syntax and then run concrete-impl rules over it.

## Regression

- canonical `impl Trait for ConcreteType` passes;
- `impl<T> ...` rejects in the current contract;
- coherence/bound satisfaction never treats a generic impl as a concrete one accidentally.

skulmakov-oss:
Reassessed against current `main` after SSF-07 Slice 6:

Current main: `4efc6ec6c198cb9e4acbdaba8d7585077def4efb`

This finding is already enforced by the existing frontend owner-layer logic; no new repair PR is required.

Evidence:

* `validate_trait_coherence()` rejects every `ImplDecl` with non-empty `type_params` unconditionally.
* The rejection occurs before the concrete `(trait_name, for_type)` coherence key is computed or inserted, so a generic/blanket impl cannot be treated as a concrete impl accidentally.
* Existing regression `blanket_impl_with_type_params_is_rejected` directly proves `impl<T> ...` rejection.
* Existing positive impl/conformance coverage proves canonical concrete impls remain admitted.

Therefore all three required regression conditions from this issue are satisfied:

1. canonical `impl Trait for ConcreteType` passes;
2. `impl<T> ...` rejects deterministically;
3. generic impls cannot reach concrete coherence/conformance bookkeeping.

The original Phase-A finding remains valid historically for its audit context, but current `main` already satisfies the required first-wave enforcement.

[...]

PR #1857, opened 2026-08-31
fix(front): resolve impl targets before Self conformance (#1667 #1651)

### LINKED ISSUE #1651: FA-02-019: impl conformance does not prove that the impl target type exists
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — incomplete trait admission proof**

## Finding

`type_check_program()` validates trait coherence/conformance using only `ImplDecl`, `TraitTable`, and `AstArena`. `validate_impl_conformance()` proves that the trait exists and that impl methods match it, but it has no `RecordTable`/`AdtTable` input and therefore cannot prove that `imp.for_type` names an actually declared nominal type.

If a trait's method signatures do not otherwise force the target through a resolved `Self` type position, an impl can reference a missing target name without this conformance boundary rejecting it.

## Risk

An `impl Trait for MissingType` shape can be admitted as a coherent/conforming impl even though the target type does not exist. Later phases then inherit an invalid abstraction relation.

## Required repair

Validate every impl target against the canonical admitted nominal-type tables before conformance succeeds. Preserve the current direct-record/ADT scope explicitly; unsupported target families should reject.

## Regression

- impl for a declared admitted target passes;
- impl for an unknown target rejects before body/lowering;
- trait existence and target existence are both mandatory admission proofs.

### LINKED ISSUE #1667: FA-02-035: trait/impl Self is hard-coded as Record and misrepresents enum impl targets

skulmakov-oss:
SSF-07: #1578

Fixes #1667
Fixes #1651

Baseline SHA: 9ddcf93e8e5ed2b349f3b3e129d836f0bcf19c63 (main, includes the #1722/#1665, #1710, and #1637/#1638 slices)

## Ground-truth reproduction

Both issues were reproduced directly against baseline before any code changed.

**#1651** (`impl_target_that_does_not_exist_rejects_even_without_any_self_reference`, and an earlier throwaway probe): a trait whose methods never reference `Self` at all —

```
trait Marker { fn ping() -> i32; }
impl Marker for MissingType { fn ping() -> i32 { return 1; } }
```

— **typechecked successfully** (`Ok(())`) on baseline `main`, even though `MissingType` is never declared anywhere as a record or enum. `validate_impl_conformance` had no `RecordTable`/`AdtTable` input at all, and nothing else in `type_check_program`'s pipeline independently checks that `imp.for_type` names a real declared type when no method signature forces that resolution.

**#1667** (`impl_self_conformance_mismatch_reports_the_correct_adt_type_not_record`, and an earlier throwaway direct-call probe): calling `substitute_trait_self_type(TypeVar(Self), Some(self_id), color_id)` where `Color` is declared as an `enum` returned `Record(SymbolId(N))` — the function unconditionally reconstructs `Type::Record(concrete_self)` for `Self`, regardless of the target's real nominal family. This surfaces as a wrong diagnostic: a genuine conformance mismatch on a `Self`-typed return for an enum impl reported `"...expected Record(SymbolId(3)) from trait 'Make'"` on baseline `main`, where the correct resolved type is `Adt(SymbolId(3))`.

Both reverts were done in place (production code only, tests unchanged) and confirmed via `cargo test -p sm-front`, then restored — see "Post-fix evidence" below for the exact commands and observed pre-fix failures.

## Why #1667 and #1651 are one repair slice

They compose through exactly one canonical impl-target-resolution boundary, with no second nominal-type authority, no parser change, and no trait/AST/generic-system redesign:

- #1651's fix *is* calling the resolver at all, unconditionally, before conformance succeeds.
- #1667's fix *is* using the resolver's actual output (a concrete `Type::Record`/`Type::Adt`) instead of blindly guessing `Record`, in the one place (`substitute_trait_self_type`) that reconstructs a concrete type for `Self`.

Investigation (see "Parser-phase decision for Self" below) showed **no parser change was needed** — impl-side `Self` already parsed through this codebase's general, pre-existing "unresolved nominal name" convention, and an already-canonical resolver (`canonicalize_declared_type_generic`) already existed and already handled the exact ambiguity/unknown-target failure modes both issues need. The entire repair is two functions in `crates/sm-front/src/typecheck.rs`.

## Root cause

One conflated assumption in two places:
1. `validate_impl_conformance` assumed the impl target's existence would always be incidentally proven by *something else* (a `Self`-typed method signature forcing canonicalization, or the separate `validate_top_level_name_collisions` pass) — untrue whenever no method signature mentions `Self`.
2. `substitute_trait_self_type` assumed "the impl target" and "a record" were the same concept, because in the frontend's original, blanket-impl-free surface, most/all impl targets in test coverage happened to be records.

## Canonical authority used

`canonicalize_declared_type_generic` (`crates/sm-front/src/lib.rs:726`), the same function every other declared-type position in the frontend (function params, return types, record fields, etc.) already goes through to disambiguate a parsed `Type::Record(name)` placeholder into the real `Type::Record`/`Type::Adt`, or reject it. Its existing behavior, reused verbatim (no new resolver written):

```rust
Type::Record(name) => match (record_table.contains_key(name), adt_table.contains_key(name)) {
    (true, false) => Ok(Type::Record(*name)),
    (false, true) => Ok(Type::Adt(*name)),
    (true, true)  => Err("top-level name '...' is ambiguously declared as both record and enum"),
    (false, false) => Err("unknown nominal type '...'"),
}
```

`validate_impl_conformance` now calls `canonicalize_declared_type_generic(&Type::Record(imp.for_type), record_table, adt_table, arena, &[])` once per impl, immediately after confirming the trait exists, propagating its error (wrapped with impl/trait context) via `?`. This single call is simultaneously the #1651 fix (fails closed on unknown/ambiguous target, unconditionally, before any method-level check) and the source of the one resolved concrete `Type` used for #1667's substitution.

**Validation-order note:** `validate_impl_conformance` still runs before `validate_top_level_name_collisions` in `type_check_program` (unchanged global ordering — no reordering was needed). A record/enum name collision can therefore still reach the resolver mid-audit; `canonicalize_declared_type_generic`'s existing `(true, true)` branch already fails closed on that ambiguity rather than guessing, so this is safe without touching global ordering. Confirmed via `impl_target_ambiguous_between_record_and_enum_rejects_deterministically`.

## Parser-phase decision for Self

No parser change. Investigated whether `parse_impl_decl`'s `with_self_type_scope(Type::Record(for_type), ...)` (line 402) needed to change to a neutral placeholder (mirroring the trait side's `Type::TypeVar(intern("Self"))`, line 336) — it does not: `Type::Record(name)` is this codebase's general, pre-existing convention for *any* not-yet-canonicalized nominal type reference (confirmed by reading `canonicalize_declared_type_generic`'s own handling of ordinary `Type::Record(name)` parameters, used for every other declared type in the language, not just impl `Self`). Writing `Self` in an impl method and writing the target type's name directly are — correctly — indistinguishable at the parsed-`Type` level; both denote the same concrete type for a given impl.

## Impl-target resolution invariant

```
Given impl.for_type: SymbolId,
resolve_impl_target = canonicalize_declared_type_generic(&Type::Record(for_type), record_table, adt_table, arena, &[])
  => Ok(Type::Record(id))   -- exactly one of Record/Adt admitted
  => Ok(Type::Adt(id))
  => Err(..)                -- ambiguous (both) or unknown (neither), always
```

Never `unknown target → assume Record → continue`. Never a second resolver competing with this one.

## Self substitution behavior

`substitute_trait_self_type(ty, self_type_var, for_type, concrete_self: &Type)` now recognizes **both** parsed forms of `Self` and substitutes both to the identical resolved `concrete_self`:

- `Type::TypeVar(name)` where `name == self_type_var` — the trait-side placeholder.
- `Type::Record(name)` where `name == for_type` — the impl-side "unresolved nominal name" form (see above).

It is applied to **both sides** of every conformance comparison (the trait's declared param/return type *and* the impl's own written param/return type), not only the trait side as before — necessary because the impl side's own `Self` occurrences are, by the point of comparison, still in their unresolved `Record(for_type)` form and must be resolved identically to compare correctly against an enum target. The existing recursive walk (tuple, `Sequence`, `Map`, `Closure` param/return, `Measured`, `Option`, `Result`) is unchanged in structure — only the base case and the threaded parameters changed; every pre-existing composite case still recurses correctly with no new gaps.

## Record-target behavior

Unchanged and reconfirmed: `trait_self_contract_allows_multiple_impl_targets`, `trait_self_contract_still_rejects_wrong_concrete_impl_parameter`, and the new `substitute_trait_self_type_covers_nested_positions_for_both_nominal_families` (record half of the table) all pass — `Self` for a record target resolves to `Type::Record`.

## ADT-target behavior

New: `Self` for an enum/ADT target resolves to `Type::Adt`, proven directly (unit-level, `substitute_trait_self_type_covers_nested_positions_for_both_nominal_families`) and end-to-end (`impl_self_for_adt_target_typechecks_body_as_the_correct_adt_not_a_guessed_record`, which pattern-matches on `self` inside the impl method body — this only typechecks if `self`'s resolved scope type is genuinely `Adt`, not a guessed `Record`).

## Unknown-target behavior

New: rejects deterministically even when no method signature references `Self` at all (`impl_target_that_does_not_exist_rejects_even_without_any_self_reference`), and even when the ambiguity is a record/enum name collision rather than an outright-missing name (`impl_target_ambiguous_between_record_and_enum_rejects_deterministically`). A real, unambiguous target with no `Self` reference still typechecks (`impl_target_that_exists_without_self_reference_still_typechecks`, positive control).

## Nested-Self substitution coverage

Table-driven unit test (`substitute_trait_self_type_covers_nested_positions_for_both_nominal_families`) exercises, for **both** a record and an ADT concrete target, and for **both** parsed `Self` forms: direct `Self`, tuple containing `Self`, `Sequence(Self)`, `Option(Self)`, `Result(Self, ..)`/`Result(.., Self)`, closure parameter containing `Self`, closure return containing `Self` — plus a check that an unrelated nominal type reference (not `Self`, not the impl target) is never substituted. Complemented by one source-level end-to-end test (`impl_self_nested_in_option_typechecks_end_to_end_for_enum_target`, `Option(Self)` through the real pipeline for an enum target).

## Validation-order implications

None beyond what's noted above under "Canonical authority used" — no reordering of `type_check_program`'s validator sequence.

## Pre-fix evidence

Verified by temporarily reverting `validate_impl_conformance`/`substitute_trait_self_type` to their exact baseline form in place (production code only; the one new unit test whose signature depends on the fix was disabled for that window via `#[cfg(any())]`, then re-enabled — no test logic was altered) and rerunning:

```
impl_target_that_does_not_exist_rejects_even_without_any_self_reference
  -> FAILED: typecheck_source returned Ok(()) instead of erroring (#1651, confirmed)

impl_target_ambiguous_between_record_and_enum_rejects_deterministically
  -> FAILED: got a DIFFERENT error ("top-level name 'Dup' cannot be used for
     both record and enum", from the later validate_top_level_name_collisions
     pass) instead of validate_impl_conformance's own ambiguity rejection --
     confirming the resolver itself did not yet fail closed on ambiguity and
     was relying on incidental coverage from an unrelated, later validator

impl_self_conformance_mismatch_reports_the_correct_adt_type_not_record
  -> FAILED: message was "...expected Record(SymbolId(3)) from trait 'Make'"
     (#1667, confirmed -- the exact historical wrong-value bug)
```

The other three new tests (`impl_self_for_adt_target_typechecks_body_as_the_correct_adt_not_a_guessed_record`, `impl_self_nested_in_option_typechecks_end_to_end_for_enum_target`, `impl_target_that_exists_without_self_reference_still_typechecks`) already passed pre-fix — expected, since they don't hit a comparison where the two independently-computed sides diverge; they remain in the suite as positive-control/characterization coverage for the fixed code path.

## Post-fix evidence

Fix restored (verified via clean `git diff main -- crates/sm-front/src/typecheck.rs`, no residual markers). Exact qualification below.

## Tests added

All in `crates/sm-front/src/typecheck.rs`:

- `substitute_trait_self_type_covers_nested_positions_for_both_nominal_families`
- `impl_self_for_adt_target_typechecks_body_as_the_correct_adt_not_a_guessed_record`
- `impl_self_conformance_mismatch_reports_the_correct_adt_type_not_record`
- `impl_target_that_does_not_exist_rejects_even_without_any_self_reference`
- `impl_target_ambiguous_between_record_and_enum_rejects_deterministically`
- `impl_target_that_exists_without_self_reference_still_typechecks`
- `impl_self_nested_in_option_typechecks_end_to_end_for_enum_target`

## Fallback/bypass audit

Reviewed the changed control-flow neighborhood in `validate_impl_conformance` and `substitute_trait_self_type`:

- `canonicalize_declared_type_generic(...).map_err(...)?` — **FAIL-CLOSED REPAIR**, this is the #1651 fix itself; propagates via `?`, no swallowed error.
- `_ => ty.clone()` catch-all in `substitute_trait_self_type` — **KEEP**, pre-existing and unchanged in spirit: the correct base case for a substitution walker (any type that is neither form of `Self` nor a composite containing it is legitimately left untouched).
- One self-review finding during drafting: I initially wrote `resolve_symbol_name(...).unwrap_or("<unknown>")` inside the new `map_err` closure (there is a pre-existing, unrelated precedent for this exact pattern elsewhere in the file at line ~2715, used only for a display string, not for admission logic). Rather than rely on that precedent, I refactored to resolve both symbol names via `?` *before* the closure, matching this file's dominant convention with zero `unwrap_or` in the new code. **FAIL-CLOSED REPAIR** (self-corrected before commit).
- No new `unwrap()`/`expect()`/`unwrap_or_default()`/`if let Ok(...)`/silent catch-all admission path anywhere in the diff.

## Sibling exposure audit

- **#1646** (`Self` outside trait/impl context silently reinterpreted as `Record`) — **UNAFFECTED**. Untouched code path (`parse_type`'s fallback when `self_type_scope` is `None`); `self_type_outside_trait_or_impl_positions_is_not_admitted` still passes unchanged.
- **#1668** (generic impl parameters admitted beyond first-wave contract) — **UNAFFECTED**. `validate_trait_coherence`'s rejection of non-empty `imp.type_params` is untouched; this PR passes `&[]` for `type_params` to `canonicalize_declared_type_generic` deliberately, matching that existing constraint, not loosening it.
- **#1669** (trait method signature types not independently canonicalized/validated) — **INTERSECTS BUT DOES NOT FIX**. Confirmed by construction: a trait parameter that is neither `Self` nor the impl's own `for_type` falls through `substitute_trait_self_type`'s unchanged `_ => ty.clone()` arm on both sides, so an unknown, non-`Self` type referenced only in a trait signature still passes conformance unvalidated. Genuinely the same class of gap as #1651 but a different surface (trait declarations, not impl targets) — left for its own issue rather than folded in here.
- **#1648** (recursive generic call inference/substitution) — **UNAFFECTED**. Entirely separate machinery (`subst_apply`, generic call-site substitution); not touched.
- **#1650** (applied generic nominal type model) — **UNAFFECTED**. No generic nominal type model exists yet for impls (type params on impls remain rejected); not touched.
- **#1717** (missing IR monomorphisation / completed-generics claim) — **UNAFFECTED**. This PR is entirely within `sm-front` frontend typecheck; impl methods still aren't registered in `FnTable` and remain uncallable from ordinary code, so no monomorphisation-adjacent claim is made or implied.

No sibling turned out to share the same root invariant such that it would remain independently broken after this repair — none were folded in.

```
cargo check --workspace --all-targets                                       -> clean
cargo test -p sm-front                                                       -> 500 passed; 0 failed
cargo test --test public_api_contracts                                       -> 88 passed (no signature/snapshot drift; both changed functions are private)
cargo test --test ssf_language_contract_drift --test canonical_source_style  -> 21 passed
cargo test --test frontend_parser_qualification --test frontend_lexer_qualification -> 17 passed
cargo test -p sm-ir -p sm-vm                                                 -> unaffected, unchanged pass counts
cargo fmt -p sm-front --check                                                -> clean
cargo clippy -p sm-front --all-targets -- -D warnings                        -> clean
cargo clippy --workspace --all-targets -- -D warnings                        -> clean
```

`cargo fmt --check` (workspace-wide, no `-p`) still fails in this local environment on the same unrelated Windows path-length OS error noted in every prior SSF-07 slice in this sequence — this PR touches nothing near whatever triggers it.

Changed files: `crates/sm-front/src/typecheck.rs`, `docs/spec/foundation_source_profile_v1.md`.

CTF touched: none

## Out of scope

Per the assigned slice boundary — none of the following are touched: generic traits, generic/blanket impls, specialization, trait objects/dynamic dispatch, monomorphisation, broader runtime trait dispatch, `Self` parser policy outside impl/trait (#1646), global `Type`-system redesign, #1636, #1652, #1654/#1721, #1729 (already merged separately), ownership fixes, module/import fixes, verifier/runtime fixes, exhaustiveness redesign, new pattern syntax/forms, async/macros, or unrelated warnings/refactors. Also explicitly out of scope and reported rather than repaired: #1669 (see sibling audit above).

## Remaining SSF-07 work

This PR does **not** close #1578 and does **not** authorize SSF-08. Candidate next slices from the earlier triage remain open (e.g. #1644 duplicate Logos `System` declarations, and the broader numeric/text/collection/closure/generic/pattern closure matrix).

### DIFF
--- a/crates/sm-front/src/typecheck.rs
+++ b/crates/sm-front/src/typecheck.rs
@@ -201,7 +201,7 @@ pub fn type_check_program(p: &Program) -> Result<(), FrontendError> {
     // M9.2 Wave 3: trait coherence and impl conformance.
     let trait_table = build_trait_table(p)?;
     validate_trait_coherence(&p.impls, &p.arena)?;
-    validate_impl_conformance(&p.impls, &trait_table, &p.arena)?;
+    validate_impl_conformance(&p.impls, &trait_table, &record_table, &adt_table, &p.arena)?;
     validate_top_level_name_collisions(p, &table, &record_table, &adt_table, &schema_table)?;
     validate_record_declarations(p, &record_table, &adt_table)?;
     validate_adt_declarations(p, &record_table, &adt_table)?;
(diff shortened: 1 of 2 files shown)


skulmakov-oss:
@codex review

Issue #1650, opened 2026-08-14
FA-02-018: generic record/ADT declarations have no faithful applied nominal type path for source instantiation


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — completed-surface claim exceeds executable type model**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

The completed M9.1 generics scope includes generic record and ADT definitions and monomorphisation. `RecordDecl` / `AdtDecl` retain `type_params`, but the current nominal `Type` representation is only `Record(SymbolId)` / `Adt(SymbolId)` with no applied type arguments, and the source type parser has no user-nominal application form analogous to `Foo(T)`/`Foo<i32>` that carries concrete arguments.

Additionally, record/ADT field validation paths still encounter declaration `TypeVar` through non-instantiated nominal declarations.

## Risk

The frontend can accept generic record/ADT definitions while lacking a faithful type representation for concrete source instantiation. This leaves the declared completed surface phase-inconsistent: definition metadata exists, but applied nominal semantics cannot be preserved end-to-end.

## Required repair

Either implement the narrow applied-nominal type/instantiation model required by the documented first-wave generic record/ADT contract, or narrow the active contract and reject these definitions until that model exists.

Do not retain a definition surface whose type arguments disappear at use sites.

## Regression

- a documented generic record can be instantiated with a concrete type and its fields typecheck with that substitution;
- same for a generic ADT payload;
- wrong type arguments reject;
- no generic nominal argument is silently erased.

Issue #1648, opened 2026-08-14
FA-02-016: generic call inference only binds direct TypeVar parameters and does not recurse through compound types


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — incomplete generic semantics / false admission boundary**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

Generic call-site inference builds substitutions only when the entire expected parameter is exactly `Type::TypeVar(T)`:

```rust
if let Type::TypeVar(tid) = &sig.params[i] { ... }
```

and `subst_apply()` substitutes only a top-level `TypeVar`, otherwise returning the original type unchanged.

Therefore generic occurrences nested inside `Sequence(T)`, `Option(T)`, `Result(T,E)`, tuples, maps, closures, measured/composite forms, etc. neither constrain T nor receive recursive substitution.

The completed generics scope explicitly motivates and admits generic use across compound families such as Sequence/Option/Result.

## Risk

Generic source can be syntactically/owner-layer admitted while ordinary compound parameter shapes cannot be instantiated according to the declared contract, producing phase inconsistency or false `cannot infer`/type mismatch behavior.

## Required repair

Implement one recursive structural unification/substitution model for the admitted first-wave type families, or narrow the public generic contract to direct-TypeVar-only until that exists.

## Regression

- `fn id<T>(x:T)->T` still works;
- admitted `Sequence(T)`, `Option(T)`, `Result(T,E)` and tuple cases infer/substitute recursively;
- conflicting structural constraints reject deterministically;
- no unresolved TypeVar survives successful call admission.

Issue #1646, opened 2026-08-14
FA-02-014: Self outside trait/impl context is silently reinterpreted as a nominal record type


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — unsupported type context silently reinterpreted**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`

## Finding

`parse_type()` recognizes the identifier `Self`. When a trait/impl self-type scope exists it returns that type, but outside such a scope it falls back to:

```rust
Type::Record(self.arena.intern_symbol("Self"))
```

The current source contract states that `Self` denotes the impl-anchored receiver contract only inside the narrow trait/impl type positions and does not widen the general executable type surface.

## Risk

Unsupported `Self` usage is not rejected at the owning parser boundary; it is silently reinterpreted as an ordinary nominal record name. This can produce confusing later errors or accidentally bind to a declaration named `Self`.

## Required repair

Outside the admitted trait/impl self-type scope, reject `Self` deterministically instead of constructing `Record("Self")`.

## Regression

- `Self` in admitted trait/impl positions preserves the intended concrete/self placeholder semantics;
- `Self` in ordinary function/local/type positions rejects immediately;
- a nominal record cannot accidentally stand in for the reserved `Self` meaning.

PR #1858, opened 2026-08-31
fix(front): validate trait signatures and executable type admission (#1669 #1647)

### LINKED ISSUE #1647: FA-02-015: QVec falls through executable type validation as supported
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — unsupported type family accepted by catch-all**

## Finding

`ensure_executable_type_supported()` explicitly recurses through known executable composite families and then ends with a broad `_ => Ok(())` arm. `Type::QVec(_)` is not rejected by a dedicated arm, so it falls through as executable-supported.

Current type documentation describes `qvec(N)` as a reserved/parser-level family rather than a current executable Foundation family.

## Risk

```text
unsupported/reserved type
→ catch-all
→ Ok(())
```

A future or parser-only type can silently cross an executable-admission check simply because it was not enumerated in the rejection logic.

## Required repair

Make executable type admission exhaustive and fail closed for non-admitted families. `QVec` must reject until explicitly promoted and qualified.

## Regression

- every currently admitted executable type passes;
- QVec rejects at frontend executable-type admission;
- adding a new Type variant cannot silently become executable through a catch-all.

### LINKED ISSUE #1669: FA-02-037: trait method signature types are not independently canonicalized or validated

skulmakov-oss:
SSF-07: #1578

Fixes #1669
Fixes #1647

Baseline SHA: 30b5fa7e0f0472db50141a5a57b670047258ddf0 (main, includes the #1722/#1665, #1710, #1637/#1638, and #1667/#1651 slices)

> **Post-review correction:** a contract audit after the first review round
> (reviewed commit `a7887ef9dec7bc622e3a64fc425c935753c81c0f`) found that this
> PR's original claim — "#1669 is fully covered and independent of #1647" —
> was incorrect. #1669 requires that *unknown OR unsupported* trait
> method-signature types reject; the original commit only closed the nominal
> (Record/Adt/unknown-name) half. Its own regression,
> `build_trait_table_admits_qvec_signatures_independent_of_1647`, actually
> *proved* the remaining gap: a `qvec` trait signature was admitted into
> `TraitTable`. The slice was therefore extended narrowly, on a new commit
> (`34c3a4a1978f6d006fb20d7ad07a33cd752d09ba`), to repair #1647 at its owning
> boundary (`ensure_executable_type_supported`) and consume that
> authoritative check from `TraitTable` construction. The original Codex
> review of `a7887ef9...` is historical evidence only; a fresh review was
> requested on the corrective HEAD.

### #1669 (nominal half, first commit `a7887ef9de`)

```
build_trait_table_rejects_unknown_parameter_type:
  trait Broken { fn f(x: MissingType) -> i32; }
  -> Ok, stored TraitMethodSig { params: [(_, Record(SymbolId(4)))], ret: I32 }

build_trait_table_rejects_unknown_return_type:
  trait Broken { fn f() -> MissingType; }
  -> Ok, stored TraitMethodSig { params: [], ret: Record(SymbolId(3)) }

build_trait_table_rejects_nested_unknown_type:
  trait Broken { fn f(x: Option(MissingType)) -> i32; }
  -> Ok, stored params: [(_, Option(Record(SymbolId(4))))]

build_trait_table_rejects_ambiguous_nominal_identity_with_no_impl:
  record Dup { n: i32 }  enum Dup { A, B }  trait T { fn a(x: Dup) -> i32; }
  -> Ok, stored params: [(_, Record(SymbolId(0)))] -- picked Record silently

build_trait_table_canonicalizes_adt_signature_to_adt_type:
  enum E { A, B }  trait T { fn a(x: E) -> E; }
  -> assertion failed: left: Record(SymbolId(0)), right: Adt(SymbolId(0))
```

### #1647 (executable-admission half, corrective commit `34c3a4a197`)

```
build_trait_table_rejects_direct_qvec_signature (pre-fix, executable check
neutralized in place):
  trait Marker { fn f(x: qvec[8]) -> i32; }
  -> panicked: "a reserved qvec trait signature must reject at admission:
     {SymbolId(0): TraitDecl { .. methods: [TraitMethodSig { ..
     params: [(SymbolId(3), QVec(8))], ret: I32 }] }}"

build_trait_table_rejects_nested_qvec_signature (same neutralization):
  trait Marker { fn f(x: Option(qvec[8])) -> i32; }
  -> panicked: "qvec nested inside Option must reject at admission:
     {.. params: [(SymbolId(3), Option(QVec(8)))] .. }"
```

Both proofs were obtained by temporarily reverting the relevant production
behavior in place (new tests unchanged), rerunning, capturing the failure,
then restoring — see "Pre-fix / post-fix evidence" below for the exact
technique used for each commit.

## Root cause

Two distinct, independently-reachable gaps:

1. **#1669's original gap** (nominal): `build_trait_table` stored the raw
   parsed `TraitDecl` verbatim, with method signature types never
   canonicalized or existence-checked against `RecordTable`/`AdtTable`.
2. **#1647's gap** (executable admission, exposed by closing #1669's first
   half): `ensure_executable_type_supported` — the function that decides
   whether a *canonical* type is on the current executable-admitted
   surface — ended in `_ => Ok(())`. Every `Type` variant it didn't
   explicitly enumerate (`QVec`, `TypeVar`, `Closure`, `RangeI32`, every
   scalar) silently passed. `build_trait_table`'s first commit correctly
   canonicalized signature types but never called this function at all, so
   a *canonically resolved but not executable* type (`QVec`) sailed through
   both checks.

## TraitTable architectural role

Unchanged from the first commit: `TraitTable` is the owner-layer canonical
trait contract (confirmed **(B)** — `validate_impl_conformance` trusts its
entries directly). This corrective commit strengthens what "canonical" means
for admission purposes: canonical identity is necessary but not sufficient.

## Canonical authority reused

No new authority introduced. `ensure_executable_type_supported` already
existed and was already the sole function this concern is owned by (used by
`type_check_function_with_tables` for ordinary function signatures);
`build_trait_table` now consumes it, mirroring exactly how it already
consumes `canonicalize_declared_type_generic`. No trait-only QVec check and
no second executable-type whitelist were added.

## Self exception model

Unchanged in spirit, extended to the executable check: `ensure_executable_
type_supported` now takes an `admitted_type_vars: &[SymbolId]` parameter —
the same narrow, caller-scoped admission list `canonicalize_declared_type_
generic` already uses for the identical type. A `Type::TypeVar` is
executable-admitted only when its name is in that list. `build_trait_table`
passes only the reserved `Self` symbol; the two existing function-signature
call sites (`type_check_function_with_tables`) pass `&func.type_params`,
preserving their exact existing behavior for both top-level and (previously
unchecked-here, now correctly checked) nested generic type-parameter
positions. TypeVar admission was **not** globalized — it remains as narrowly
scoped as before, just explicit instead of falling through a catch-all.

## Stored TraitTable representation

Unchanged from the first commit for the nominal half. Now additionally:
a successful `TraitTable` entry's stored types are guaranteed to be both
canonical *and* executable-admitted — `qvec` (or any future reserved family)
can never reach storage.

## Validation ordering

No change. `build_trait_table` still fails closed entirely on its own,
before `type_check_program` reaches `validate_trait_coherence`/
`validate_impl_conformance`.

## Pre-fix / post-fix evidence

**#1647 / `ensure_executable_type_supported`:** temporarily replaced the new
`Type::TypeVar`/`Type::QVec` match arms with the exact pre-fix `Ok(())`
behavior (function signature and every other arm left untouched), reran:

- `build_trait_table_rejects_direct_qvec_signature` → FAILED (`Ok`
  returned, exact stored value quoted above)
- `build_trait_table_rejects_nested_qvec_signature` → FAILED (same)

Restored the fix (verified via `git diff` against the exact reverted
region), reran both → pass. The two new direct unit tests on
`ensure_executable_type_supported` itself
(`ensure_executable_type_supported_rejects_qvec_directly_and_when_nested`,
`ensure_executable_type_supported_admits_ordinary_families`) needed their
new 4-argument call signature to exist to compile at all, so they were
temporarily disabled via `#[cfg(any())]` for the revert window only (not
logic-altered) and re-enabled immediately after restoring the fix — no
disabled test remains in the final commit (verified: `grep -n
"TEMP-DISABLED"` returns nothing).

**#1669 (nominal half):** unchanged from the first commit's evidence — see
above.

## Tests added (this corrective commit)

`crates/sm-front/src/lib.rs`:

- `build_trait_table_rejects_direct_qvec_signature` (replaces the
  incorrect `..._admits_qvec_signatures_independent_of_1647`)
- `build_trait_table_rejects_nested_qvec_signature`

`crates/sm-front/src/typecheck.rs`:

- `ensure_executable_type_supported_rejects_qvec_directly_and_when_nested`
  (table-driven: direct, `Option`, `Sequence`, `Result(qvec,i32)`,
  `Result(i32,qvec)`, tuple)
- `ensure_executable_type_supported_admits_ordinary_families` (positive
  control: every scalar, `RangeI32`, `Unit`, `Record`, `Adt`, `Tuple`,
  `Option`, `Result` still admits)

## Full `Type`-variant executable-admission classification

All 20 `Type` variants, exhaustively matched (compiler-enforced — no `_`
arm compiles):

| Variant | Classification | Handling |
|---|---|---|
| `Quad`, `Bool`, `Text`, `I32`, `U32`, `Fx`, `F64`, `Unit`, `RangeI32` | ADMITTED EXECUTABLE (leaf) | `Ok(())` directly |
| `Tuple`, `Sequence`, `Map`, `Option`, `Result`, `Measured` | ADMITTED EXECUTABLE (composite) | recurse into every nested `Type` (unchanged from before, `Closure` newly added — see below) |
| `Closure` | ADMITTED EXECUTABLE (composite) | **newly** recurses into `param`/`ret`; previously fell through the catch-all and was never recursed into at all |
| `Record`, `Adt` | ADMITTED EXECUTABLE (nominal) | unchanged — resolves the symbol name for diagnostics |
| `TypeVar` | STRUCTURAL / PLACEHOLDER WITH SPECIAL CONTEXT | admitted only if the name is in the caller-scoped `admitted_type_vars` list; otherwise rejects |
| `QVec` | RESERVED / NON-EXECUTABLE | rejects deterministically (`#1647`'s fix) |

`Closure` and `RangeI32` were **not** explicitly matched before this commit
(both fell through `_ => Ok(())`); `RangeI32` is a leaf so this had no
practical effect, but `Closure`'s param/return were previously never
recursively checked at all by this function (unlike
`canonicalize_declared_type_generic`, which already recursed into it) — this
commit closes that latent gap too, since making the match exhaustive
required deciding every variant explicitly.

## Exact change to `ensure_executable_type_supported`

- Added `admitted_type_vars: &[SymbolId]` parameter.
- Removed the `_ => Ok(())` catch-all entirely — replaced with 20 explicit
  arms (grouped where behavior is identical).
- `Type::TypeVar(name)`: admits iff `admitted_type_vars.contains(name)`,
  else `Err`.
- `Type::QVec(_)`: unconditional `Err` ("qvec is a reserved type and is not
  yet admitted as an executable type in {context}").
- `Type::Closure`: new explicit arm, recurses into `param`/`ret`.
- Both existing call sites (`type_check_function_with_tables`, ordinary
  function param/return checks) updated to pass `&func.type_params` —
  their pre-existing top-level `TypeVar`-skip-when-generic guards are
  unchanged and still correct; the new parameter only affects *nested*
  generic-`TypeVar` positions, which were previously unchecked by this
  function (though already checked earlier by `canonicalize_declared_type_
  generic`) and are now explicitly, correctly admitted via the same list.
- `ensure_executable_type_supported` changed from module-private to
  `pub(crate)` so `build_trait_table` (in `lib.rs`) can call it — a
  crate-internal visibility change only; the crate's *external* public API
  is unaffected (confirmed: `public_api_contracts` golden snapshot test
  unchanged).

## Was the broad `_ => Ok(())` removed?

Yes, entirely, from `ensure_executable_type_supported`. `ensure_storage_
type_supported` — a separate function with the identical catch-all pattern,
but for record/ADT *field storage* types, not executable *signature*
types — was **not** touched; reported as a sibling finding below, not
folded into this repair.

## Direct QVec post-fix behavior

`ensure_executable_type_supported(&Type::QVec(8), ..)` → `Err("qvec is a
reserved type and is not yet admitted as an executable type in ..")`,
proven directly (`ensure_executable_type_supported_rejects_qvec_directly_
and_when_nested`).

## Nested QVec post-fix behavior

Same error, proven for `Option(qvec)`, `Sequence(qvec)`, `Result(qvec,i32)`,
`Result(i32,qvec)`, and a tuple containing `qvec` — all in the same test.

## TraitTable QVec post-fix behavior

`build_trait_table` rejects both a direct and an `Option`-nested `qvec`
trait method signature parameter, proven end-to-end through the real
canonicalize → executable-check → store pipeline
(`build_trait_table_rejects_direct_qvec_signature`,
`..._rejects_nested_qvec_signature`).

## Self behavior

Unaffected and reconfirmed: `build_trait_table_admits_reserved_self_in_
direct_and_nested_source_positions` and `canonicalize_declared_type_
generic_admits_self_in_closure_positions` still pass — `Self` now also
passes the new executable-admission check (via the same `admitted_type_
vars = [self_symbol]` list threaded through).

## Record/ADT canonicalization behavior

Unaffected and reconfirmed: `build_trait_table_canonicalizes_record_
signature_to_record_type` / `..._canonicalizes_adt_signature_to_adt_type`
still pass unchanged.

## Existing #1669 regressions

All reconfirmed green in the full suite run on this HEAD: unknown parameter/
return/nested-nominal rejection, ambiguous-nominal rejection, Record/Adt
canonicalization, Self admission (direct + nested source positions), no
accidental Self substitution, trait's-own-type-parameter rejection.

## Existing #1667/#1651 regressions

All reconfirmed green: `impl_self_*` (3), `impl_target_*` (4, incl. the
parser-level anchoring test), `substitute_trait_self_type_covers_nested_
positions_for_both_nominal_families`.

## Sibling issues deliberately left open

- **`ensure_storage_type_supported`** (not a filed issue number, discovered
  during this corrective pass) — structurally identical catch-all pattern,
  but a separate function owning record/ADT *field* storage-type admission,
  not executable *signature* types. Left untouched; worth its own finding
  if not already tracked.
- **#1635, #1646, #1668, #1648, #1650, #1717** — unaffected, unchanged from
  the first commit's sibling audit (see below).

### Full sibling-exposure classification (carried forward + updated)

- **#1635** (generic trait declarations admitted beyond first-wave
  contract) — **INCIDENTALLY BLOCKED AT OWNER ADMISSION** (unchanged).
- **#1646** (`Self` outside trait/impl context reinterpreted as `Record`) —
  **UNAFFECTED**.
- **#1647** — **FIXED BY THIS ROOT** (this corrective commit).
- **#1668** (generic impl parameters beyond first-wave contract) —
  **UNAFFECTED**.
- **#1667 / #1651** — previously fixed (prior merged slice); reconfirmed
  unaffected and still green.
- **#1648** (recursive generic call inference) — **UNAFFECTED**.
- **#1650** (applied generic nominal type representation) — **UNAFFECTED**.
- **#1717** (missing IR monomorphisation) — **UNAFFECTED**.

```
cargo check --workspace --all-targets                                       -> clean
cargo test -p sm-front                                                       -> 514 passed; 0 failed
cargo test --test public_api_contracts                                       -> 88 passed (no signature/snapshot drift)
cargo test --test ssf_language_contract_drift --test canonical_source_style  -> 21 passed
cargo fmt -p sm-front --check                                                -> clean
cargo clippy -p sm-front --all-targets -- -D warnings                        -> clean
cargo clippy --workspace --all-targets -- -D warnings                        -> clean
```

`cargo fmt --check` (workspace-wide, no `-p`) still fails in this local
environment on the same unrelated Windows path-length OS error noted in
every prior SSF-07 slice in this sequence — unchanged, confirmed unrelated
to this diff.

Changed files (this corrective commit): `crates/sm-front/src/lib.rs`,
`crates/sm-front/src/typecheck.rs`, `docs/spec/foundation_source_profile_v1.md`.

Changed files (cumulative, both commits): same three files.

CTF touched: none

## Out of scope

Per the assigned slice boundary — none of the following are touched:
`ensure_storage_type_supported`'s identical catch-all (reported above, not
fixed), QVec parser syntax (#1640/#1641/#1642/#1643), generic traits
(#1635), generic impls (#1668), `Self` outside trait/impl context (#1646),
IR monomorphisation (#1717), runtime trait dispatch, or any unrelated
warnings/refactors. Fixing executable admission of QVec is explicitly not
the same as fixing QVec parser syntax — those remain separate findings.

## Remaining SSF-07 work

This PR does **not** close #1578 and does **not** authorize SSF-08. Open
candidates from prior triage remain: #1644 (duplicate Logos `System`
declarations), #1635, #1646, #1668, `ensure_storage_type_supported`'s
catch-all (new, unfiled), and the broader numeric/text/collection/closure/
generic/pattern closure matrix.

### DIFF
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -608,16 +608,28 @@ pub fn build_trait_table(program: &Program) -> Result<TraitTable, FrontendError>
                     .params
                     .iter()
                     .map(|(name, ty)| {
-                        Ok((
-                            *name,
-                            canonicalize_declared_type_generic(
-                                ty,
-                                &record_table,
-                                &adt_table,
-                                &program.arena,
-                                &admitted_type_vars,
-                            )?,
-                        ))
+                        let canonical = canonicalize_declared_type_generic(
+                            ty,
+                            &record_table,
+                            &adt_table,
+                            &program.arena,
+                            &admitted_type_vars,
+                        )?;
+                        // FA-02-015 / #1647: TraitTable admission proves both
+                        // canonical nominal identity AND that the canonical
+                        // type is on the current executable-admitted
+                        // surface -- never store first and validate later.
+                        typecheck::ensure_executable_type_supported(
+                            &canonical,
+                            &program.arena,
+                            &admitted_type_vars,
+                            format!(
+                                "parameter '{}' of trait method '{}'",
+                                resolve_symbol_name(&program.arena, *name)?,
+                                resolve_symbol_name(&program.arena, m.name)?
+                            ),
+                        )?;
+                        Ok((*name, canonical))
                     })
                     .collect::<Result<Vec<_>, FrontendError>>()?;
                 let ret = canonicalize_declared_type_generic(
@@ -626,6 +638,15 @@ pub fn build_trait_table(program: &Program) -> Result<TraitTable, FrontendError>
                     &adt_table,
                     &program.arena,
                     &admitted_type_vars,
+                )?;
+                typecheck::ensure_executable_type_supported(
+                    &ret,
+                    &program.arena,
+                    &admitted_type_vars,
+                    format!(
+                        "return type of trait method '{}'",
+                        resolve_symbol_name(&program.arena, m.name)?
+                    ),
                 )?;
                 Ok(TraitMethodSig {
                     name: m.name,
@@ -1718,16 +1739,14 @@ fn main() {
     }
 
     #[test]
-    fn build_trait_table_admits_qvec_signatures_independent_of_1647() {
-        // Sibling-exposure note for #1647 (QVec falls through executable
-        // type validation as supported): this repair calls only
-        // canonicalize_declared_type_generic, never
-        // ensure_executable_type_supported (the separate, already-tracked
-        // function #1647 is about), so #1669's closure is independent of
-        // #1647's status either way. QVec is not a Record/Adt nominal
-        // reference, so canonicalize's unrelated catch-all arm returns it
-        // unchanged -- proving this repair neither depends on nor
-        // absorbs #1647's defect.
+    fn build_trait_table_rejects_direct_qvec_signature() {
+        // FA-02-015 / #1647, corrected: a post-review contract audit found
+        // that #1669's "unsupported type must reject" requirement depends
+        // on ensure_executable_type_supported actually being exhaustive.
+        // build_trait_table now consumes that authoritative check after
+        // canonicalizing, so a reserved, not-yet-promoted-to-executable
+        // QVec signature must reject at trait admission, exactly like an
+        // unknown nominal type does -- never silently stored.
         let src = r#"
             trait Marker {
                 fn f(x: qvec[8]) -> i32;
@@ -1737,7 +1756,35 @@ fn main() {
             }
         "#;
         let program = parse_program(src).expect("parse");
-        build_trait_table(&program).expect("qvec trait signature must admit, independent of #1647");
+        let err = build_trait_table(&program)
+            .expect_err("a reserved qvec trait signature must reject at admission");
+        assert!(
+            err.message.contains("qvec is a reserved type"),
+            "unexpected error: {}",
+            err.message
+        );
+    }
+
+    #[test]
+    fn build_trait_table_rejects_nested_qvec_signature() {
+        // Recursive enforcement: an unsupported family nested inside an
+        // otherwise-admitted composite must also reject.
+        let src = r#"
+            trait Marker {
+                fn f(x: Option(qvec[8])) -> i32;
+            }
+            fn main() {
+                return;
+            }
+        "#;
+        let program = parse_program(src).expect("parse");
+        let err = build_trait_table(&program)
+            .expect_err("qvec nested inside Option must reject at admission");
+        assert!(
+            err.message.contains("qvec is a reserved type"),
+            "unexpected error: {}",
+            err.message
+        );
     }
 
     #[test]
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -567,6 +567,28 @@ pub fn build_adt_table(program: &Program) -> Result<AdtTable, FrontendError> {
 
 #[cfg(any(feature = "alloc", feature = "std"))]
 pub fn build_trait_table(program: &Program) -> Result<TraitTable, FrontendError> {
+    // FA-02-037 / #1669: TraitTable is the owner-layer canonical trait
+    // contract (the same architectural role FnSig/FnTable plays for
+    // functions), not an unchecked cache of raw parsed declarations. Every
+    // non-Self method-signature type must resolve against the same
+    // canonical nominal authority every other declared-type position in the
+    // frontend uses (see build_fn_table above), before this trait is
+    // admitted -- independent of whether any impl ever exists. Built
+    // locally (mirroring build_fn_table) rather than accepting
+    // caller-supplied tables, so this invariant holds for every public
+    // caller of build_trait_table, not only type_check_program's.
+    let record_table = build_record_table(program)?;
+    let adt_table = build_adt_table(program)?;
+    // The sole admitted type variable in a trait method signature is the
+    // trait-side `Self` placeholder (Type::TypeVar, interned once per
+    // parsed trait). A trait's own declared type_params are deliberately
+    // NOT admitted here: generic traits are not part of the first-wave
+    // canonical contract (#1635), so a signature that references its own
+    // trait type parameter is rejected by the same "type variable is not
+    // in scope" path canonicalize_declared_type_generic already uses for
+    // any other out-of-scope type variable, rather than silently admitted.
+    let self_type_var = program.arena.symbol_to_id.get("Self").copied();
+    let admitted_type_vars: Vec<SymbolId> = self_type_var.into_iter().collect();
     let mut out = BTreeMap::new();
     for t in &program.traits {
         if out.contains_key(&t.name) {
@@ -578,7 +600,48 @@ pub fn build_trait_table(program: &Program) -> Result<TraitTable, FrontendError>
                 ),
             });
         }
-        out.insert(t.name, t.clone());
+        let methods = t
+            .methods
+            .iter()
+            .map(|m| {
+                let params = m
+                    .params
+                    .iter()
+                    .map(|(name, ty)| {
+                        Ok((
+                            *name,
+                            canonicalize_declared_type_generic(
+                                ty,
+                                &record_table,
+                                &adt_table,
+                                &program.arena,
+                                &admitted_type_vars,
+                            )?,
+                        ))
+                    })
+                    .collect::<Result<Vec<_>, FrontendError>>()?;
+                let ret = canonicalize_declared_type_generic(
+                    &m.ret,
+                    &record_table,
+                    &adt_table,
+                    &program.arena,
+                    &admitted_type_vars,
+                )?;
+                Ok(TraitMethodSig {
+                    name: m.name,
+                    params,
+                    ret,
+                })
+            })
+            .collect::<Result<Vec<_>, FrontendError>>()?;
+        out.insert(
+            t.name,
+            TraitDecl {
+                name: t.name,
+                type_params: t.type_params.clone(),
+                methods,
+            },
+        );
     }
     Ok(out)
 }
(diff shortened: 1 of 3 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

Issue #1773, opened 2026-08-16
FA-09-005: verified function invocation has no target arity/type contract and can accept invalid argument vectors


skulmakov-oss:
Umbrella: #1617
Module: **09 — `sm-vm`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **SEMANTIC-DRIFT · FAIL-OPEN · FALSE-READY · IDENTITY/TRUST-GAP**

## Trigger

Invoke a verified function with the wrong number or runtime types of arguments through:

```rust
run_verified_function_semcode_with_args(..., args: Vec<Value>)
```

or construct admitted SemCode whose internal `CALL` argument vector does not match the source-level target function signature.

## Claim / confidence source

Semantic source functions are statically typed and their declared parameter count/types are part of source meaning.

`sm-vm/src/lib.rs` contains tests explicitly named:

- `test_6_reject_wrong_argument_count`
- `test_7_reject_wrong_argument_type`

which create confidence that the public verified-function boundary enforces those contracts.

The public helper is documented as the canonical verified function invocation path with arguments and structured return value.

## Actual guarantee

`FunctionBytecode` contains no function parameter signature or arity metadata. Its runtime fields include name, strings/symbols, debug/ownership metadata, code and instruction offset, but no parameter count/types.

`push_frame()` therefore cannot validate a target signature. It simply chooses:

```rust
let initial_reg_count = 16usize.max(args.len());
```

and copies every supplied `Value` into the leading registers.

Consequences:

### Wrong argument count

A missing argument is not rejected at call entry. If the missing parameter is never read, execution can still succeed. Extra arguments likewise have no target-arity check and can be silently unused.

### Wrong argument type

The VM does not compare incoming `Value` kinds against declared parameter types. A mismatch fails only if some later opcode happens to demand another runtime shape. If a wrong-typed parameter is unused, the call can complete successfully.

The same structural limitation applies to internal `CALL`: the instruction carries an `argc`/argument register list, but the VM has no target signature against which to validate it.

## Gap

Typed function meaning is lost at the executable invocation boundary:

```text
source: fn f(a: i32, b: bool)
→ lowering / SemCode function body
→ FunctionBytecode has no (i32,bool) signature
→ VM receives arbitrary Vec<Value>
→ no arity/type proof at frame creation
```

[...]

PR #1873, opened 2026-08-31
fix(ir): reject unspecialized generic functions at executable boundary (#1717)

skulmakov-oss:
## Baseline

Fetched fresh `origin/main`; branched from `0a9976ad253b0ca24d045281068b43f96b8f7f9e` (post-#1871/#1650 merge) in an isolated `EnterWorktree` worktree, per the shared-worktree collision hazard discovered in an earlier slice.

## Pre-fix compile pipeline

`compile_program_to_ir_with_options_and_profile` — the central entrypoint every public `compile_program_to_ir*`/`compile_program_to_semcode*` variant funnels through — ran:

```rust
let mut program = parse_program_with_profile(input, parser_profile)?;
let fn_table = build_fn_table(&program)?;
let record_table = build_record_table(&program)?;
let adt_table = build_adt_table(&program)?;
type_check_program(&program)?;
for f in &program.functions {
    let lowered = lower_function_to_ir_with_tables(f, &program.arena, &fn_table, &record_table, &adt_table, &program.impls)?;
    ...
}
```

`grep -n "type_params" crates/sm-ir/src/legacy_lowering.rs` returned **zero matches** before this fix. No deliberate check on a function's declared type parameters existed anywhere at the IR boundary.

## Public lowering-entrypoint inventory

`lower_function_to_ir`, `lower_function_to_ir_with_tables` (private, shared choke point), `compile_program_to_ir`, `compile_program_to_immutable_ir`, `compile_program_to_ir_with_options`, `compile_program_to_ir_with_profile`, `compile_program_to_ir_with_options_and_profile` (the entrypoint all `compile_program_to_ir*` variants funnel through), `compile_program_to_ir_optimized`, `compile_program_to_semcode`, `compile_program_to_semcode_with_options`, `compile_program_to_semcode_with_options_debug` (→ `compile_program_to_immutable_ir` → same central entrypoint). `emit_ir_to_semcode` operates on already-lowered `&[IrFunction]`, not source level — out of scope.

## Reproductions (pre-fix, empirically observed)

| Case | Source | Pre-fix result |
|---|---|---|
| Direct-T | `fn id<T>(x: T) -> T` | Rejected — but with the stale, misleading `"...deferred to M9.1 Wave 2"` diagnostic (accidental hit via `canonicalize_declared_type`'s `TypeVar` rejection), not a deliberate IR boundary |
| Nested-T | `fn keep<T>(x: Option(T)) -> Option(T)` | Same accidental rejection, same stale diagnostic |
| **Unused-T** | `fn marker<T>(x: i32) -> i32 { return x; }` | **`Ok(...)`** — lowered as an ordinary `IrFunction { name: "marker", params: [I32], ... }`. `type_params` silently discarded. This is genuine type erasure, not a partial admission gap. |
| Concrete-call | `id(1)` | Same accidental rejection as direct-T (declaration itself already fails) |
| Multi-instantiation | `id(1)` and `id("x")` both called | Same accidental rejection, both no-ops relative to instantiation count — no specialization identity exists |
| Generic-to-generic | `fn outer<T>(x: T) -> T { return id(x); }` calling `fn id<T>(y: T) -> T` | `outer` itself already fails via the accidental path (its own `x: T` parameter) |
| Trait-bound | `fn wrap<T: Zeroable>(x: Option(T)) -> i32` with satisfying `impl Zeroable for ZeroInt` | Frontend `type_check_program` succeeds (trait bound proven satisfiable); IR lowering still hits the accidental rejection via `Option(T)` |

The unused-T case is the strongest evidence: it proves the pre-fix "rejection" seen in the other six cases was never a deliberate admission boundary — it was an incidental side effect of which structural positions a `TypeVar` happened to occupy. Whether a generic function's declaration survived to execute depended on where its own type parameter was written, not on whether the declaration was generic at all.

## TypeVar handling inventory (IR layer)

- **ACCIDENTAL FAILURE**: `canonicalize_declared_type` (non-generic, used in `lower_function_to_ir_with_tables` for params/ret canonicalization) unconditionally rejects `Type::TypeVar` with the stale `"...deferred to M9.1 Wave 2"` message. Not deliberate here — this function is the shared non-generic canonicalizer, reused incidentally in a context that can receive generic declarations. This exact stale message was already documented as an "accidental gate" in #1650's record/ADT slice; it recurs here for the same structural reason and remains out of scope for this slice (tracked separately, see Sibling classifications below).
- **UNREACHABLE CLAIM**: `callable_family_for_type(Type::TypeVar(_))` returns an error, with its own doc comment stating it "can never reach this function as a resolved, written parameter type... exists only for exhaustiveness." Confirmed via tracing: `canonicalize_declared_type` already rejects `TypeVar` before this function is ever called with one, in every case except unused-T — and unused-T never produces a `TypeVar` parameter type at all, so this arm is genuinely unreachable given the codebase's actual call order, not just by claim.
- **EARLY ADMISSION (new, this PR)**: `ensure_function_is_ir_concrete`, called as the first line of `lower_function_to_ir_with_tables`, before any canonicalization or lowering work begins.

## IR pass inventory

`crates/sm-ir/src/passes/` contains only `cleanup.rs` (`StructuralCleanup`) and `crystalfold.rs` (`CrystalFold`). No specialization or monomorphisation pass exists.

## SemCode identity / Verifier / VM audit

`compile_program_to_semcode` calls `compile_program_to_immutable_ir`, which funnels through the same central gated entrypoint — the new regression `compile_program_to_semcode_rejects_generic_function_before_any_emission` proves rejection happens before `emit_semcode` is ever reached, so no partial or malformed SemCode artifact can be produced for a generic function. `sm-verify`/`sm-vm` are never reached for a generic function under either the old accidental path or the new deliberate one; `cargo test -p sm-vm --offline` (116 passed, 0 failed) confirms sm-vm is otherwise entirely unaffected.

## Architecture decision: Model B (deterministic rejection), not Model A (IR monomorphisation)

**Model A** (implement a real specialization pass: call-graph specialization discovery, deterministic specialization keys, deduplication, call-target rewriting, recursive-specialization termination, trait-bound-respecting substitution) was rejected for this slice. It is a large, multi-part undertaking with its own architecture-decision surface (specialization key scheme, generic-to-generic recursion termination proof, verifier/VM changes for specialized output) that cannot be proven complete within one repair slice, and the task's acceptance criteria for Model A were not met by anything in this diff.

**Model B** (a single, narrow, deliberate rejection boundary) is implemented. It closes the false-ready surface — no generic function can silently execute with erased type parameters — without claiming monomorphisation exists. This mirrors the #1650 precedent (narrow the admitted contract to what is actually implemented, rather than build the larger missing feature under time pressure).

## The exact executable generic invariant

A generic function declaration (`type_params` non-empty) is deterministically rejected at the IR/SemCode boundary — regardless of whether its type parameter is referenced in its own signature, regardless of call-site count (zero, one, or many), regardless of trait bounds, and regardless of which other public entrypoint is used to reach IR lowering. No generic function can produce an `IrFunction`, ever, under the current contract.

## Frontend-vs-executable contract

Frontend generic function **declaration admission** (`build_fn_table`, at most one type parameter, #1634) and **call-site type inference/substitution** (`collect_generic_constraints`/`subst_apply`, #1648, #1649) remain fully closed and unaffected — `cargo test -p sm-front --offline` (565 passed, 0 failed) confirms this. What #1717 closes is a separate, previously ungated layer: whether a frontend-admitted generic declaration is allowed to reach and execute as IR. Frontend admission answers "does this parse and type-check"; this fix answers "can this run" — and the answer is now uniformly no, deterministically, not "sometimes, depending on where T appears."

## Sibling classifications

- **#1634** (function arity ≤ 1) — unaffected, frontend-layer, already closed.
- **#1648** (recursive call-site inference) — unaffected, frontend-layer, already closed.
- **#1649** (canonical `FnSig` authority) — unaffected, frontend-layer, already closed.
- **#1650** (generic record/ADT declarations) — sibling false-ready pattern at a different layer (frontend construction vs. this slice's IR execution); independently closed, unaffected by this diff.
- **#1722** — UNAFFECTED. Live #1722 is `FA-04-016: named-argument lowering evaluates in parameter order instead of source order` (closed); it concerns named/default-argument evaluation order, not generics or IR admission, and this diff does not touch it. (Corrected: an earlier revision of this PR body misattributed the stale TypeVar diagnostic below to #1722 — that was wrong; #1722 is a real, separate, already-closed issue with no connection to this slice.)
- **Pre-existing stale TypeVar diagnostic** — the `"...deferred to M9.1 Wave 2"` message in `canonicalize_declared_type` is a known, pre-existing issue (documented as an "accidental gate" during #1650's slice and again here), but it is not tracked by any filed issue number found in this repository at time of writing, so none is cited here. It is not the mechanism this PR relies on or weakens; `ensure_function_is_ir_concrete` is a new, independent, and now primary admission check that fires before that message's accidental path is ever reached for the direct-T/nested-T cases, and the unused-T case never reached it at all pre-fix. Out of scope for this slice.
- **#1665, #1773** — unrelated surfaces (not touched, not affected by this diff).

## Fallback/bypass audit

`git diff | grep -E "^\+" | grep -viE "^\+\+\+" | grep -iE "unwrap_or|unimplemented!\(\)|todo!\(\)|allow\(dead_code|continue;|\.first\(\)|truncate\(|clear\(\)|match _ =>|_ =>"` — zero matches in the diff.

## Pre-fix → post-fix proof

Saved `git diff -- crates/sm-ir/src/legacy_lowering.rs` to a local patch file. Temporarily neutralized the check (`if false && !func.type_params.is_empty() { // TEMP-DISABLED-FOR-PREFIX-PROOF`). Ran the 8 central regressions — **all 8 failed** as predicted: 7 with the old accidental `"...deferred to M9.1 Wave 2"` message (or simply the wrong diagnostic shape), and `compile_program_to_ir_rejects_unused_generic_type_param` failed with concrete evidence of type erasure — `Ok([IrFunction { name: "marker", params: [I32], ... }, IrFunction { name: "main", ... }])`. Restored the check by reverting the neutralization. Confirmed zero residual `TEMP-DISABLED` markers via grep. Confirmed the restored file's diff was byte-identical to the saved patch via `diff`. Deleted the patch files.

## Docs / status correction

- `docs/roadmap/language_maturity/generics_full_scope.md`: added a highly-visible current-status correction directly under the historical "Status: completed" line (left verbatim per the existing "historical record, not retroactively edited" precedent set by #1650's own addendum); re-scoped the #1650 addendum's "genuinely closed" wording for generic functions to explicitly mean frontend admission/inference only, not IR execution; appended a new "Post-Close-Out Correction (SSF-07 #1717)" section documenting this fix in the same style/rigor as the existing #1650 section.
- `docs/spec/foundation_source_profile_v1.md`: added normative execution-boundary wording immediately after the existing records/ADT correction paragraph, explaining the IR admission gap and its fix, without promoting generics to "Included."
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md`: **corrective commit** — the original single "First-wave generics / monomorphisation" row kept the status label **Landed but unqualified** while its own rewritten Evidence text said "No IR monomorphisation or specialization pass exists anywhere in `sm-ir`" — the status and evidence directly contradicted each other, reproducing exactly the FALSE-READY pattern this issue exists to eliminate, just relocated into the matrix instead of fixed. Per review on this PR, split into two rows so each carries one accurate status, honoring the matrix's own "exactly one status per family" rule: **First-wave generics: frontend declaration admission and call-site inference** (`F`, **Landed and qualified on `main`**) and **First-wave generics: IR monomorphisation / execution** (`I/V/R`, **Roadmap**). `cargo test --test ssf_status_drift` (1 passed) confirms the required status vocabulary and base-SHA authority are still present.

## Workspace-wide stale-claim search

Searched for `monomorphi[sz]ation|generic lowering|generic functions? (is|are|complete)|generics? complete|Wave 3` across `*.md`/`*.rs`. All matches outside the three docs above were either: (a) unrelated tracks' own "Wave 3" labels (UI backend, collections, closures, etc. — different subtracks, not generics), (b) pre-existing frontend code comments already correctly scoped as "distinct from IR monomorphisation (#1717, unaffected)" (added proactively in #1648/#1650, verified accurate and left unchanged), or (c) this PR's own new code/docs. No additional stale false-ready claims found.

## Qualification

- `cargo build -p sm-ir --tests --offline` — clean
- `cargo test -p sm-ir --offline` — 131 passed, 0 failed (121 baseline + 10 new)
- `cargo test -p sm-front --offline` — 565 passed, 0 failed (unaffected)
- `cargo test -p sm-vm --offline` — 116 passed, 0 failed (unaffected)
- `cargo test --test public_api_contracts` — 88 passed, 0 failed
- `cargo fmt -p sm-ir --check` — clean
- `cargo test --test ssf_status_drift --offline` — 1 passed, 0 failed (corrective commit: matrix row split)

## Changed files

- `crates/sm-ir/src/legacy_lowering.rs` — `ensure_function_is_ir_concrete` + call site + 10 new regression tests
- `docs/roadmap/language_maturity/generics_full_scope.md`
- `docs/spec/foundation_source_profile_v1.md`
- `docs/roadmap/stable_foundation/semantic_stable_foundation_matrix.md`

## Out of scope

This PR does **not**: implement IR monomorphisation or any specialization pass; implement multi-parameter generics; restore generic record/ADT declarations or their construction; introduce runtime generic dispatch; weaken verifier or VM admission in any way; fix the pre-existing stale `"...deferred to M9.1 Wave 2"` diagnostic string (out of scope here); close #1578 (the SSF-07 parent umbrella); or authorize SSF-08. It closes exactly one gap: a generic function declaration can no longer silently reach and execute as IR with its type parameters erased. Generic function *execution* remains not implemented — it is now deterministically and uniformly rejected instead of inconsistently rejected-or-erased depending on incidental structure.


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1879, opened 2026-09-01
docs(ssf08): freeze bounded ownership model position (#1579)

skulmakov-oss:
SSF-08: #1579
Architecture decision only.
No ownership implementation repair in this PR.

## Corrections (owner review before merge)

Four corrections applied across three owner-review rounds, all recorded in the decision record's "Correction Log" section. Current HEAD `715efde5b95354718f6a87ef6649d252bb0e40b7` (prior HEADs: `3ac20fe71f6066c4bf8d1e16dda5f089e9175bd7` -> `f3e172bc416a187a9c11c208081987bfa4d84156` -> `3ad8507f0e75912b2788bfab508b5bd6b163e3af` -> `715efde5b95354718f6a87ef6649d252bb0e40b7`):

1. **Quota/taxonomy current-facing contradiction, fixed.** `semantic_stable_foundation_matrix.md`'s "Quotas/fuel and trap taxonomy" row read **Landed and qualified on `main`**, directly contradicted by this same PR's own evidence (`#1759`-`#1763`: most quota kinds inert, taxonomy drifted). Corrected to **Landed but unqualified** (existing status vocabulary, nothing invented).
2. **Scalar/direct-root classification, resolved empirically.** The prior revision called scalar/direct-root "already qualified" while filing a Known Blocking Finding against the same invariant, without first proving whether Semantic's actual contract is copy or move. Resolved by full-pipeline empirical reproduction (compile -> verify -> VM run) for `i32`, `bool`, `quad`, `f64`, `text`, plus a non-Copy `record`: plain root rebinding is copy-by-value by design, matching the existing `use_after_move_rejects` test's own "i32 is Copy" comment. Durable regression added: `tests/scalar_root_rebind_copy_semantics_e2e.rs` (6 tests).
3. **Residual echoes of the pre-Correction-2 model, and a maturity-status overclaim Correction 2's caveat didn't reach.** Removed stale `consumed`-as-required-repair references from the Decision Matrix and Repair Dependency Order. Separately, `semantic_stable_foundation_matrix.md`'s "OWN0 tuple/direct-record paths" row still read **Landed and qualified on `main`** even though this record proves `#1709` is a false-negative-direction bug inside that exact frozen contract and `#1725`/`#1726` violate the record's own normative Path-identity/Event-timing invariants for tuple/record too. Corrected to **Landed but unqualified**, same fix shape as Correction 1.
[...]

Issue #1778, opened 2026-08-16
FA-11-001: prom-abi publicly admits non-canonical Quad(u8) values with no ABI-level validation


skulmakov-oss:
Umbrella: #1617
Module: **11 — `prom-abi`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FALSE-READY · IDENTITY/TRUST-GAP · QUALIFICATION-GAP**
Related cross-module finding: #1775 / `FA-09-007`

## Trigger

Any public `PrometheusHostAbi` implementation can construct and return:

```rust
AbiValue::Quad(4)
AbiValue::Quad(0xff)
```

because the ABI carrier is declared as:

```rust
pub enum AbiValue {
    Quad(u8),
    ...
}
```

with no checked constructor, closed quad carrier, `TryFrom<u8>`, or ABI-boundary validator.

## Claim / confidence source

`prom-abi` is the canonical host ABI boundary. Its public value vocabulary is a trust-boundary contract, and the platform audit explicitly treats externally constructible public states as readiness claims.

The canonical Semantic quad domain contains exactly four values (`N/F/T/S`), not 256 raw byte states.

The public API snapshot also pins `AbiValue` as part of the exported ABI surface.

## Actual guarantee

The ABI type system itself cannot distinguish canonical quad values from malformed bytes. Every `u8` is representable as `AbiValue::Quad` and `PrometheusHostAbi::gate_read/state_query` can return it as `Ok(...)`.

No test in `prom-abi` exercises rejection of an out-of-domain quad value because the crate currently has no rejection path for one.

## Gap

The host trust boundary admits a larger state space than the Semantic value domain it claims to carry:

```text
host-produced raw byte 0..=255
→ AbiValue::Quad(raw)
→ ABI says Ok(value)
→ downstream must guess/validate/normalize
```

That makes canonical quad validity an undocumented downstream responsibility rather than an invariant of the ABI value itself.

Today the downstream VM does not reject this state; #1775 / FA-09-007 proves that `sm-vm` masks high bits and silently converts every malformed byte into `N/F/T/S`. The two findings are intentionally separate root causes:

- **FA-11-001:** the ABI publicly represents invalid quad states at all;
- **FA-09-007:** the VM turns those invalid states into successful Semantic values instead of rejecting them.

Fixing either one alone should not erase the other Phase-A root cause.

## Cross-phase impact

```text
[...]

Issue #1763, opened 2026-08-16
FA-08-005: RuntimeTrap is documented as shared frozen failure vocabulary but most variants are not emitted by the VM


skulmakov-oss:
Umbrella: #1617
Module: **08 — `sm-runtime-core`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **SEMANTIC-DRIFT · INERT-CONTRACT · FALSE-READY · QUALIFICATION-GAP**

## Trigger

Consume `sm_runtime_core::RuntimeTrap` as the shared runtime failure taxonomy.

The enum contains variants including:

```text
StackOverflow
StackUnderflow
TypeMismatch
InvalidOpcode
InvalidJump
CapabilityDenied
AbiViolation
VerifierRejected
QuotaExceeded(...)
```

However `sm-vm` has parallel top-level `RuntimeError` variants for these failure families and does not construct the corresponding `RuntimeTrap` variants on the inspected production paths.

At the audit SHA, direct `RuntimeError::Trap(...)` construction was confirmed only for the narrower runtime trap subset such as assertion failure, borrow/write conflict, division by zero and arithmetic overflow.

## Claim / confidence source

- `sm-runtime-core` is the shared runtime vocabulary owner.
- `docs/roadmap/language_maturity/core_trust_freeze/trap_taxonomy.md` says the current execution-failure surface is frozen for PCC core readiness and cites `RuntimeTrap` as shared evidence.
- root diagnostic code contains an exhaustive `RuntimeTrap` text mapping, reinforcing the appearance that every variant is part of one authoritative runtime trap channel.

## Actual guarantee

The public failure model is split:

```text
RuntimeTrap::<some families>
RuntimeError::<parallel families>
```

For many public `RuntimeTrap` variants, the VM's actual failure path uses the parallel `RuntimeError` variant instead.

## Gap

The shared vocabulary enum overstates its authority: presence of a `RuntimeTrap` variant does not mean callers can actually receive that family through `RuntimeError::Trap`. Documentation and diagnostic consumers can therefore treat an enum shape as stronger evidence than the production execution path supports.

This is taxonomy/contract drift, not a new runtime safety bypass, hence P2.

## Evidence

- `crates/sm-runtime-core/src/lib.rs::RuntimeTrap`
- `crates/sm-vm/src/semcode_vm.rs::RuntimeError`
- VM construction sites for `RuntimeError::Trap`
[...]

Issue #1759, opened 2026-08-16
FA-08-001: max_steps and max_calls are published bounded-execution quotas but are never enforced by sm-vm


skulmakov-oss:
Umbrella: #1617
Module: **08 — `sm-runtime-core`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P1**
Classification: **FAIL-OPEN · FALSE-READY · QUALIFICATION-GAP**

## Trigger

Run verified SemCode whose execution exceeds either published dynamic budget:

- `RuntimeQuotas::max_steps`
- `RuntimeQuotas::max_calls`

The model owner exposes both as first-class quota kinds:

```rust
QuotaKind::Steps
QuotaKind::Calls
```

and all baseline profiles assign finite values (`100_000/250_000` steps, `16_384/32_768` calls).

But the VM has no step counter, no call counter, and no `enforce_quota(..., QuotaKind::Steps, ...)` or `QuotaKind::Calls` path. The instruction loop records opcodes only for optional profiling; profiling is not quota enforcement.

## Claim / confidence source

`docs/spec/quotas.md` says runtime quotas define the **bounded execution contract** and that `sm-vm` enforces quotas during execution. `docs/spec/vm.md` likewise lists runtime quota enforcement as a VM responsibility.

## Actual guarantee

At the audit SHA only these quota kinds are dynamically enforced by VM code:

- `Frames`
- `StackDepth`
- `Registers`
- `EffectCalls`

`Steps` and `Calls` are never charged.

A verified backward loop can therefore continue beyond `max_steps`; repeated successful calls can exceed `max_calls` without that quota ever firing (subject only to other unrelated limits such as frame depth).

## Gap

The shared runtime vocabulary advertises finite execution budgets that do not participate in execution. For `max_steps`, this breaks the core bounded-execution promise itself: the configured finite step budget cannot stop an otherwise admissible non-terminating loop.

This is not a missing diagnostic label. It is a published safety/resource boundary that is absent from the consumer.

## Evidence

- `crates/sm-runtime-core/src/lib.rs::{QuotaKind,RuntimeQuotas}`
- `crates/sm-vm/src/semcode_vm.rs` — VM state and quota enforcement paths
- `docs/spec/quotas.md`
- `docs/spec/vm.md`

Phase A evidence only. **No repair during Phase A.**

PR #1880, opened 2026-09-02
fix(front): make ScopeEnv ownership state survive control flow

### LINKED ISSUE #1656: FA-02-024: statement `if` discards ownership state changes made in branch bodies
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — ownership state fail-open across control flow**

## Finding

Statement `if` checking clones the incoming `ScopeEnv` independently for the then/else branches, checks each branch, pops the branch scope, and returns `Ok(())` without joining moved/borrowed path state back into the parent environment.

Therefore ownership effects on outer bindings performed inside a branch disappear after the `if`.

## Risk

A value/path moved in a reachable branch can appear available again after the conditional because the owner-layer state change existed only in a discarded clone.

## Required repair

Define and implement conservative ownership-state join semantics for conditional control flow. Any post-`if` availability must be valid for every reachable branch; uncertainty must not restore availability.

## Regression

- move in one branch is reflected conservatively after the if;
- move in both branches remains moved;
- branch-local bindings do not leak;
- borrow/path-state joins preserve existing M9.7/M9.8 conflict guarantees.

### LINKED ISSUE #1657: FA-02-025: loop statement checking discards ownership changes to outer bindings
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — ownership state fail-open across loops**

## Finding

The statement-level loop forms (`while`, range/foreach/loop paths in the current checker) typecheck their bodies against cloned/scoped environments. Ownership changes to bindings that originated outside the loop are not conservatively propagated back to the enclosing `ScopeEnv` after body checking.

## Risk

A move/borrow of an outer value inside a loop body can be forgotten after the loop, making the value appear available again. Iteration also means the same invalid state may be re-entered on later iterations without owner-layer proof.

## Required repair

Define conservative loop ownership semantics, including loop-carried state/fixed-point handling or a deliberately narrower rejection policy. Do not discard outer-binding ownership mutations simply because body scope is temporary.

## Regression

- moving an outer binding in a loop cannot make it available after the loop without proof;
- repeated-iteration move/borrow conflicts are caught;
- loop-local bindings remain scoped;
- zero-iteration cases are handled conservatively and explicitly.

### LINKED ISSUE #1658: FA-02-026: statement match discards ownership effects on non-scrutinee outer variables
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — partial ownership propagation across match**

## Finding

Statement `match` builds an environment clone per arm/default. Pattern capture plans are later applied back to the original scrutinee through `apply_plans_to_scrutinee`, but ownership mutations performed by arm bodies on *other outer bindings* live only in the cloned arm environments and are discarded.

## Risk

The checker tracks one class of match ownership effect (scrutinee pattern capture) while silently forgetting another (moves/borrows of surrounding values inside branch bodies). A value moved inside an arm can appear available after the match.

## Required repair

Join ownership state from all reachable match arm/default bodies into the enclosing environment, in addition to the existing scrutinee-plan accounting. The post-match state must be conservative across alternatives.

## Regression

- pattern capture state remains enforced;
- move/borrow of a separate outer variable in one arm is reflected after match;
- all-arm and default combinations are joined conservatively;
- arm-local bindings do not leak.

### LINKED ISSUE #1659: FA-02-027: value-producing if expressions cannot propagate ownership effects from branch bodies
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — ownership state lost in expression control flow**

## Finding

`infer_expr_type()` receives `&ScopeEnv`. For `Expr::If`, both value-producing branch blocks are inferred from the same immutable parent environment. Any ownership changes to outer bindings occurring while checking branch statements cannot be propagated to the caller after the expression.

## Risk

A move/borrow performed inside a value-producing `if` branch can disappear once the expression finishes, restoring apparent availability of an outer value.

## Required repair

Introduce an ownership-aware expression checking result/state transition, or explicitly restrict ownership-affecting operations inside value-producing branches until conservative join semantics exist. Do not infer type while silently dropping semantic state changes.

## Regression

- branch result type checking remains intact;
- outer-value moves/borrows in expression branches are reflected conservatively afterward;
- local bindings remain scoped;
- statement and expression `if` ownership semantics agree.

### LINKED ISSUE #1660: FA-02-028: loop expressions cannot propagate ownership state changes to the enclosing scope
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — ownership state lost in value-producing loops**

## Finding

`infer_loop_expr_type()` is reached through `infer_expr_type(..., env: &ScopeEnv, ...)`. It can determine the loop expression's value type, but the ownership state of outer bindings mutated/moved while checking the loop body cannot be returned to the enclosing scope through this immutable interface.

## Risk

Ownership effects inside a value-producing loop can disappear after the expression, and loop-carried move/borrow conflicts cannot be represented faithfully at the caller boundary.

## Required repair

Make loop-expression checking ownership-state-aware or reject ownership-affecting loop bodies until conservative loop state semantics are implemented. Type inference alone must not erase move/borrow state.

## Regression

- valid value-producing loops retain their result typing;
- outer moves/borrows cannot disappear after the loop;
- repeated-iteration ownership conflicts reject;
- loop-local bindings remain local.

### LINKED ISSUE #1661: FA-02-029: match expressions skip scrutinee consumed-state propagation
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — pattern move state lost after expression match**

## Finding

The expression `match` path validates pattern binding/capture plans but runs through `infer_expr_type`, which receives an immutable `&ScopeEnv`. The implementation explicitly notes that consumed-state marking cannot be applied at this expression layer.

As a result, a pattern that captures by move can be accepted and typed while the enclosing scope is not updated to reflect that the matched scrutinee/path was consumed.

## Risk

A value moved through a match expression can appear usable again afterward.

## Required repair

Make match-expression checking return/apply ownership state transitions, or prohibit ownership-consuming patterns in expression matches until that state can be propagated. Do not validate a move and then forget it.

## Regression

- borrow-only match expression preserves availability where allowed;
- move capture makes the relevant scrutinee/path unavailable afterward;
- mixed arms are conservatively joined;
- expression and statement match ownership rules agree.

### LINKED ISSUE #1662: FA-02-030: if-let expressions skip scrutinee consumed-state propagation
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — pattern ownership state lost after if-let expression**

## Finding

The `Expr::IfLet` checker builds and validates a `BindingPlan` and applies bindings to a cloned then-environment, but the implementation explicitly notes:

```text
scrutinee consumed-state is enforced at statement level only
(infer_expr_type receives &ScopeEnv, which is immutable)
```

Thus a move capture in an if-let expression is checked locally but not propagated to the enclosing scrutinee state.

## Risk

A value/path consumed by an admitted if-let expression can appear available again afterward.

## Required repair

Make if-let expression checking ownership-state-aware, or reject consuming patterns in this expression form until propagation exists.

## Regression

- borrow pattern behavior remains valid;
- move capture marks the relevant scrutinee/path unavailable after the expression;
- then/else branch state joins conservatively;
- statement and expression ownership semantics do not diverge.

### LINKED ISSUE #1663: FA-02-031: ownership plan validation/application silently bypasses projected scrutinees
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — projected-path ownership bypass**

## Finding

The ownership helpers that validate/apply pattern plans to a scrutinee only operate when the scrutinee expression is a plain variable:

```rust
let Expr::Var(name) = arena.expr(scrutinee_expr) else {
    return Ok(()); // validation
}
```

and the state-application path similarly does nothing for non-`Var` scrutinees.

However the frontend already has path extraction for projected expressions such as record fields / indexed sequence paths (`expr_access_path`).

## Risk

Matching/destructuring a projected path can bypass prior ownership conflict checks and can fail to record move/borrow state on the actual root/path. Unsupported expression shape becomes “no ownership work needed.”

## Required repair

Use the canonical access-path extraction for admitted projected scrutinees, and fail closed for a consuming pattern whose ownership target cannot be represented. Do not treat non-Var as automatically safe.

## Regression

- plain-variable scrutinee behavior remains intact;
- record-field / admitted index projections receive exact path-state validation and application;
- conflicting prior move/borrow rejects;
- unrepresentable consuming scrutinee rejects rather than returning Ok/no-op.

### LINKED ISSUE #1664: FA-02-032: ScopeEnv ownership APIs fail open when the requested binding is missing
Umbrella: #1617
Module: **02 — `sm-front`**
Severity: **P2 — internal ownership invariant loss becomes allowed/no-op**

## Finding

Several `ScopeEnv` operations treat a missing binding as success/no state:

- `mark_consumed(name)` → no-op if not found;
- `is_consumed(name)` → `false` if not found;
- `mark_path_state(...)` → no-op if not found;
- `check_path_available(...)` → `Ok(())` if not found;
- `check_capture_allowed(...)` → `Ok(())` if not found;
- `is_const(name)` → `false` if not found;
- `is_mutable(name)` → `false` if not found.

These methods participate in semantic/ownership state, not presentation-only behavior.

## Risk

If a caller loses/misroutes a binding identity, the ownership layer can silently interpret “binding missing” as “not consumed / available / capture allowed / not const,” hiding the invariant failure instead of surfacing it.

## Required repair

Separate query APIs whose absence is genuinely expected from semantic mutation/validation APIs that require an existing binding. Required-binding operations should return a deterministic frontend/internal semantic error when the binding cannot be found. Where a query legitimately needs tri-state behavior, return `Option`/`Result` rather than collapsing absence to `false`.

Do not replace a lost owner-state object with an implicit Available/non-const state.

## Regression

- required ownership operation on missing binding rejects;
- const/mutability checks cannot mistake missing binding for an existing non-const/non-mutable binding;
- existing binding paths retain current behavior;
- lookup APIs that intentionally expose absence remain explicit;
- source-level unknown variable diagnostics are not duplicated or weakened.

skulmakov-oss:
## Summary

SSF-08 Lane 1: a single canonical `ScopeEnv` ownership-state join model, replacing the nine independent state-loss bugs (#1656-#1664) that all shared one root cause — ownership-state transitions computed in a cloned/temporary environment (branch arm, loop body, match arm, value-block) were discarded instead of joined back into the enclosing environment.

This is **not** nine patches. Every control-flow construct (`if`/`match`/`if-let`, in both statement and expression form, plus loops) now routes through one of two shared primitives:

- `ScopeEnv::join_ownership_from(&mut self, successors: &[ScopeEnv])` — the conservative join: for each pre-existing binding, `consumed = OR` over every successor, and `path_state` is the union of every successor's recorded restrictions (never removed, only accumulated). A path is only readable afterward if no successor recorded it as moved.
- `run_loop_body_to_fixed_point(...)` — repeatedly runs the loop body against a `candidate` env and joins the result back with the pre-loop candidate (soundly modeling "0 more iterations" as one more join input) until the state stops changing (`PartialEq`-derived fixed point), capped at 64 iterations with a **fail-closed internal error** (not a silent fallback) if that cap is ever hit.

`if`/`match` dispatch is shared between statement-position (`check_stmt`) and loop-body statement-position (`check_loop_expr_stmt`) via a `BodyStmtChecker` fn-pointer, and value-producing `if`/`match`/`if-let`/`loop` expressions share the identical join primitives through `build_pattern_arm_env` / `infer_value_block_type`. There is one semantic authority, not per-construct duplicates.

**Governing constraints honored:** Position A (bounded deterministic VM, no borrow-checker-general claims) untouched. Plain root rebinding (`let y = x;`) stays copy-by-value — `tests/scalar_root_rebind_copy_semantics_e2e.rs` passes unmodified, and zero new call sites to `mark_consumed`/`.consumed` were added anywhere (production or test). #1724 (lexical binding identity) is not touched. File scope: `crates/sm-front/src/lib.rs`, `crates/sm-front/src/typecheck.rs`, `crates/sm-front/src/types.rs` only — no changes to `sm-ir`, `sm-format`, `sm-verify`, `sm-vm`, `sm-runtime-core`.

**Do not merge.** This is a draft PR for review, matching the discipline established on the prior SSF-08 architecture PR (#1879).

## Per-issue evidence

| Issue | Pre-fix reproduction | Repair mechanism | Regression test(s) | Result |
|---|---|---|---|---|
| #1656 (`if` statement) | A move inside one/both branches of a statement `if` was checked against a cloned `then_env`/`else_env` that was discarded after the branch; the outer `env` never saw the move | `check_if_branches_joined` clones `env` into `then_env`/`else_env`, runs each branch through the shared `BodyStmtChecker`, then unconditionally `env.join_ownership_from(&[then_env, else_env])` — then and else successor states are always joined; when the source has no `else`, the empty `else_block` represents the unchanged pre-`if` successor (zero statements run against `else_env`, so it stays a plain clone of the pre-branch `env`), not a special-cased omission from the join | `ssf08_1656_one_branch_move_persists_after_if`, `ssf08_1656_both_branches_move_rejects_after`, `ssf08_1656_sibling_paths_across_branches_both_retained`, `ssf08_1656_branch_local_binding_does_not_leak`, `ssf08_1656_no_else_move_still_persists` | PASS — outer effect survives; branch-local bindings confirmed not to leak |
| #1657 (loops) | A move inside a loop body was checked against a cloned `body_env`, discarded each iteration; a move on iteration 1 was forgotten by iteration 2 | `run_loop_body_to_fixed_point`: a real fixed-point over `candidate`, re-joining the body's effect each pass until convergence (soundness argument: joining a spurious "0 iterations" baseline can only ever be a subset of the true reachable restrictions, never remove one) | `ssf08_1657_while_loop_move_persists_after_loop`, `ssf08_1657_while_loop_repeated_move_of_same_path_rejects`, `ssf08_1657_for_range_loop_variable_does_not_leak`, `ssf08_statement_loop_break_move_persists_after_loop` | PASS — outer effect survives across iterations; loop-local variables (`for i in ...`) confirmed not to leak; repeated re-move across iterations correctly rejects |
| #1658 (`match` statement) | Each arm's ownership effect was computed against a cloned/temporary env, discarded per-arm | `check_match_arms_joined`: every arm (plus `default` if present) builds its own env via `build_pattern_arm_env`, runs to completion, and all reachable successors are joined via `join_ownership_from` | `ssf08_1658_arm_moves_unrelated_outer_binding_persists_after_match`, `ssf08_1658_default_arm_included_in_join`, `ssf08_1658_arm_local_binding_does_not_leak`, `ssf08_1658_sibling_arms_do_not_contaminate_each_other` | PASS — outer effect survives; arm-local pattern bindings don't leak; sibling arms verified isolated from each other pre-join |
| #1659 (`if` expression) | `infer_expr_type`'s `Expr::If` case discarded both branch envs after computing the tail type | `Expr::If` now collects `then_result_env`/`else_result_env` from `infer_value_block_type` and joins both into `env` before returning the tail type | `ssf08_1659_expr_if_ownership_effect_survives`, `ssf08_1659_expr_if_result_type_still_correct_when_state_preserved` | PASS — ownership effect reaches the caller; result-type inference unaffected by the fix |
| #1660 (`loop` expression) | `infer_loop_expr_type` had no loop-carried analysis at all (only statement loops did) | Generalized `run_loop_body_to_fixed_point` to accept a `BodyStmtChecker`, reused identically for `infer_loop_expr_type` (this gap was self-caught during implementation, before it could become a second silent defect) | `ssf08_1660_loop_expr_outer_effect_survives`, `ssf08_1660_loop_expr_break_type_still_correct` | PASS — outer effect survives; `break`-type inference unaffected |
| #1661 (`match` expression) | `infer_match_expr_type` discarded every arm's env after computing `arm_ty` | Rewritten to mirror `check_match_arms_joined`'s structure via `build_pattern_arm_env` per arm, collecting all successors (arms + default) and joining once at the end | `ssf08_1661_match_expr_move_pattern_state_reaches_caller`, `ssf08_1661_match_expr_alternatives_join_conservatively` | PASS — arm effects reach the caller; conservative join across alternatives confirmed (a path moved in only one alternative is rejected afterward) |
| #1662 (`if let` expression) | `Expr::IfLet`'s then/else envs were discarded after tail-type inference | Then-branch uses `build_pattern_arm_env` (pattern-bound scope popped before joining); else-branch is a plain `env` clone-through; both joined via `join_ownership_from` | `ssf08_1662_if_let_expr_success_move_reflected_after`, `ssf08_1662_if_let_expr_else_branch_included_in_join`, `ssf08_1662_if_let_expr_pattern_binding_does_not_leak` | PASS — success-path move reaches the caller; else-branch included in the join; pattern-bound name confirmed scoped to the then-branch only |
| #1663 (projected scrutinee) | A move captured through a projected scrutinee (`c.pair` in `match c.pair { Pair{a,b} => ... }`) was either silently untracked or the whole pattern-capture machinery skipped ownership entirely for non-`Var` scrutinees | `apply_arm_pattern_capture` uses `expr_access_path` to resolve the scrutinee to `(root, base_path)` and composes `full_path = base_path.extend(&item.path)` (new `PatternPath::extend`) before calling `check_capture_allowed`/`mark_path_state`; an unresolvable-but-projection-shaped scrutinee (e.g. dynamic sequence index) is a hard `Err`, never a silent skip | `ssf08_1663_projected_record_field_scrutinee_move_is_tracked`, `ssf08_1663_projected_record_field_scrutinee_sibling_field_still_usable`, `ssf08_1663_dynamically_indexed_sequence_scrutinee_with_capture_rejects` | PASS — move on the exact projected path is tracked; a genuine sibling field remains usable; a dynamically-indexed scrutinee with a capturing pattern deterministically rejects instead of silently skipping the check |
| #1664 (fail-open/fail-closed API classification) | All seven `ScopeEnv` ownership APIs (`mark_consumed`, `is_consumed`, `mark_path_state`, `check_path_available`, `check_capture_allowed`, `is_const`, `is_mutable`) collapsed a missing binding into a plausible success/default state | **See the dedicated "#1664 corrective round" section below** — 6 of 7 APIs are now genuinely fail-closed (`Result`-returning via `require_binding`/`require_binding_mut`); `is_const` could not be changed in place without editing a Lane 2 crate, so it keeps its fail-open shape for that one external dependency and gained a fail-closed sibling (`is_const_checked`) for every in-crate use instead | See table in the corrective-round section (10 tests total: 3 carried over + 7 new) | **PARTIAL** — see below for why `Closes #1664` is *not* included |

### Cross-product tests (proving one shared model, not per-construct duplicates)

- `ssf08_cross_product_match_arm_modifies_outer_tuple_inside_value_block`
- `ssf08_cross_product_if_let_on_projected_record_scrutinee_with_move_then_sibling`
- `ssf08_cross_product_loop_body_conditional_move_then_second_pass_conflict`

All three combine two-or-more of the above constructs in a single program and pass — the join model composes correctly rather than needing construct-specific fixups.

## #1664 corrective round: full seven-API audit

A review pass on the first version of this PR correctly identified that only 2 of the 7 APIs #1664 names (`mark_path_state`, `check_capture_allowed`) had actually been made fail-closed; the other five (`mark_consumed`, `is_consumed`, `check_path_available`, `is_const`, `is_mutable`) still collapsed a missing binding into a plausible default. This section documents the full audit and repair.

Every production call site of all seven APIs was located and classified. All are **REQUIRED_BINDING** — none of these queries has a legitimate "the binding may genuinely not exist" caller; every call site either (a) already independently established the binding's existence immediately beforehand, or (b) is now restructured to do so.

| API | Production call sites found | Classification | Fix |
|---|---|---|---|
| `mark_path_state` | `crates/sm-front/src/typecheck.rs` (many, via `ScopeEnv` join/capture machinery) | REQUIRED_BINDING | Already fixed in the prior round: `Result`-returning via `require_binding_mut` |
| `check_capture_allowed` | `crates/sm-front/src/typecheck.rs` (`apply_arm_pattern_capture`) | REQUIRED_BINDING | Already fixed in the prior round: `Result`-returning via `require_binding` |
| `check_path_available` | One call site: top of `infer_expr_type` (`typecheck.rs`) | REQUIRED_BINDING | Internals changed from `if let Some(binding) = self.binding(name) { .. } else { Ok(()) }` to `let binding = self.require_binding(name)?;` — now genuinely fails closed. The canonical "unknown variable" diagnostic is preserved through **call ordering**, not by the validation API absorbing the missing case: the call site now does `if env.get(name).is_some() { env.check_path_available(name, &path)?; }`, so a genuinely-absent root binding skips the ownership check entirely and falls through to `Expr::Var`'s own `env.get(*v).ok_or(..."unknown variable"...)`, which raises the one, undivided, correct diagnostic. No duplicated diagnostic paths. |
| `mark_consumed` | Zero anywhere in the workspace (only one pre-existing *test* call site) | REQUIRED_BINDING | `Result`-returning via `require_binding_mut`. Zero external dependents, so this was a free change. No new production call sites added — whole-root "consumed" remains inert under Position A's frozen copy-by-value root-rebinding semantics (PR #1879); `let y = x;` stays copy-by-value, proven by `tests/scalar_root_rebind_copy_semantics_e2e.rs` (still 6/6). |
| `is_consumed` | Zero anywhere in the workspace | REQUIRED_BINDING | `Result<bool, FrontendError>`-returning via `require_binding`. Zero dependents, free change. |
| `is_mutable` | Zero anywhere in the workspace | REQUIRED_BINDING | `Result<bool, FrontendError>`-returning via `require_binding`. Zero dependents, free change. |
| `is_const` | Zero in `sm-front`'s own production code *(both call sites already independently resolve the binding via `env.get(*name).ok_or(...)` immediately beforehand)* — **but three live production call sites in `crates/sm-ir/src/legacy_lowering.rs`** (lines ~2467, ~5011, ~6107 at the time of audit), a **Lane 2 crate this PR cannot modify** | REQUIRED_BINDING in every context it's actually used, **but frozen at the API-signature level** | See "Why `is_const` blocks full #1664 closure" below |

### Why `is_const` blocks full #1664 closure

`is_const`'s current signature is `pub fn is_const(&self, name: SymbolId) -> bool`. Changing it to `Result<bool, FrontendError>` (or `Option<bool>`) — the only way to make a *missing binding* distinguishable from *an existing non-const binding* — is a breaking signature change. `crates/sm-ir/src/legacy_lowering.rs` calls `env.is_const(*capture)` / `env.is_const(*name)` at three live production sites, all as a plain `if env.is_const(...) { .. }` boolean condition. Changing the signature without also updating those three call sites will not compile; updating them means editing `sm-ir`, a Lane 2 crate — explicitly out of scope for SSF-08 Lane 1 (file scope: `crates/sm-front` only; "do not alter Lane 2").

This is confirmed, not assumed: `cargo check --workspace --all-targets` passes with `is_const` left exactly as-is (proving Lane 2 is genuinely unaffected by this PR) and would not compile if its signature changed.

**What was still done, within scope:**
- Added `pub(crate) fn is_const_checked(&self, name: SymbolId) -> Result<bool, FrontendError>`, the fail-closed counterpart, implemented via `require_binding`.
- Migrated all three `sm-front`-internal call sites (`Stmt::Assign`, `Stmt::AssignTuple` in `check_stmt`, and `ensure_const_initializer_safe`'s `Expr::Var` arm) to `is_const_checked`. The first two were already safe (existence proven by a preceding `env.get(*name).ok_or(...)?` in the same block) — the migration is a pure hardening, not a behavior change, there. The third was a **genuine pre-existing bug**: `ensure_const_initializer_safe` runs *before* `infer_expr_type_with_expected` in the `Stmt::Const` path (see `check_stmt`), so its `Expr::Var` arm had no prior existence check to rely on — `const X: i32 = totally_unknown_name;` used to report `"'totally_unknown_name' is not const"` (wrong: it doesn't exist at all) instead of the canonical `"unknown variable 'totally_unknown_name'"`. Fixed by resolving existence first inside that arm itself. Regression: `ssf08_1664_const_initializer_unknown_variable_diagnostic_remains_canonical`.
- `is_const` itself is left with its exact original signature and behavior, with a doc comment explaining precisely why (quoted above), and an explicit instruction that it must gain no new `sm-front` call sites. `ssf08_1664_is_const_legacy_bool_api_remains_fail_open_for_lane2_compat` pins this as *deliberate*, not an oversight.

### Regressions covering all seven APIs (`crates/sm-front/src/typecheck.rs`; 7 new this round + 3 carried over from the prior round)

| Proof required | Test |
|---|---|
| `mark_consumed(missing)` → explicit failure | `ssf08_1664_mark_consumed_fails_closed_on_missing_binding` |
| `is_consumed(missing)` → explicit failure | `ssf08_1664_is_consumed_fails_closed_on_missing_binding` |
| `mark_path_state(missing)` → explicit failure | (carried over from prior round) `ssf08_1664_mark_path_state_fails_closed_on_missing_binding` |
| `check_path_available(missing)` → explicit failure as a required check | `ssf08_1664_check_path_available_fails_closed_on_missing_binding` |
| `check_capture_allowed(missing)` → explicit failure | (carried over from prior round) `ssf08_1664_check_capture_allowed_fails_closed_on_missing_binding` |
| `is_const` (fail-closed form) `(missing)` → explicit failure | `ssf08_1664_is_const_checked_fails_closed_on_missing_binding` |
| `is_const` legacy symbol `(missing)` → documented, deliberate exception | `ssf08_1664_is_const_legacy_bool_api_remains_fail_open_for_lane2_compat` |
| `is_mutable(missing)` → explicit failure | `ssf08_1664_is_mutable_fails_closed_on_missing_binding` |
| source-level unknown variable → unchanged canonical diagnostic (assignment/read path) | (carried over) `ssf08_1664_unknown_variable_diagnostic_remains_canonical` |
| source-level unknown variable in a const initializer → canonical diagnostic, not "is not const" | `ssf08_1664_const_initializer_unknown_variable_diagnostic_remains_canonical` |

All 634 `sm-front` lib tests pass (627 after the original Lane 1 round + 7 net-new this round).

### Consistency check: `Closes #1664` removed

Per the explicit stop condition for this PR: *if #1664 cannot be completely repaired without breaking an actually frozen public API, remove `Closes #1664` and report why, rather than silently narrowing the issue.* `is_const`'s bool-returning signature is exactly that case — a real, external, Lane-2-crate dependency this PR is not permitted to touch. `Closes #1664` has been **removed** from this PR's closing keywords below. #1664 remains open, now scoped precisely to: *`is_const`'s signature cannot be changed without a Lane 2 (sm-ir) edit, which requires separate, explicit authorization out of SSF-08 Lane 1's scope.* Every other API #1664 names is fully repaired and regression-proven in this PR.

## Self-correction found during this PR: or-pattern rejection regression

While running the full workspace qualification suite (`cargo test --workspace`), `match_surface_negative_fixtures_reject_deterministically` failed: the `negative_quad_or_pattern_lowering_rejected` fixture, which must be rejected with `"or-pattern match arms ('A | B') are not supported"`, was unexpectedly passing type-checking.

**Root cause:** the or-pattern rejection check (SSF-07) lived inline inside `check_match_arms_joined` (the statement-`match` path), but was never carried into `build_pattern_arm_env` — the function actually shared between statement-`match`, expression-`match`, and `if-let`. Statement-form `match` kept rejecting or-patterns; expression-form `match` (`let x = match ... { A | B => ... }`) silently stopped rejecting them. This was exactly the class of bug this PR exists to eliminate: two code paths that were supposed to share one authority quietly diverged.

**Fix:** moved the check into `build_pattern_arm_env` itself (removing the now-redundant inline copy from `check_match_arms_joined`), so every caller — statement `match`, expression `match`, statement `if-let`, expression `if-let` — enforces it identically through the one shared function. Verified against the pre-Lane-1 baseline via a temporary stash (the fixture passes cleanly on `origin/main`, confirming this was a Lane-1-introduced regression, not pre-existing) before fixing, and confirmed `cargo test --test match_surface_qualification` and the full `cargo test --workspace` pass afterward.

This is disclosed here rather than silently folded into the diff because it is exactly the kind of gap the task's no-fallback/no-hidden-state-loss audits exist to catch, and it was only caught by running qualification fixtures beyond `sm-front`'s own unit tests.

## Determinism hardening

`join_ownership_from` iterates a scope's bindings (a `BTreeMap`, so already ordered) via `.keys()` before applying the join. The only diagnostic reachable from that loop (`"internal ownership state: required binding {} is missing from a branch successor"`) is a should-be-unreachable internal-invariant defense (every successor is cloned from the same predecessor at the same scope depth, so a name present in the predecessor is guaranteed present in every successor) — but to remove any doubt per the "no iteration-order-dependent diagnostics" requirement, and to not depend on the underlying collection type staying `BTreeMap`, the collected name list is explicitly sorted (`names.sort_unstable()`, `SymbolId: Ord`) before iteration.

## Out-of-scope defect found, filed, and documented — not fixed here

While building the #1663 regression tests, a **pre-existing, previously-undiscovered** frontend defect surfaced: reading a two-level-nested field projection (`c.pair.b`) spuriously rejects if a *sibling* field of the one-level-intermediate base was moved (`c.pair.a`), because the intermediate base (`c.pair`) is independently re-checked as a whole-value read. Root cause: `infer_expr_type_no_check` only skips its own path-availability re-check when the base expression is a bare `Expr::Var`; a `RecordField` base (as in `c.pair.b`, whose own base `c.pair` is itself a `RecordField` access) falls through to a second, independent, shorter-path check that the outer expression's own check has already subsumed.

Confirmed **100% independent of this PR's changes** via a straight-line program with zero `if`/`loop`/`match`, reproducing identically — this is a read-path-checking gap orthogonal to control-flow/branch-joining (this PR's actual scope), not a Lane 1 regression.

Filed as **#1881 (FA-02-039: "nested projection reads re-check an intermediate base as a whole-value access")**, umbrella #1617, Module 02 (`sm-front`), Classification SEMANTIC-DRIFT · FALSE-READY, with the exact reproduction, straight-line independence proof, root-cause citation, and expected/current-behavior contrast. Routed as a frontend read-path-validation defect (base-expression resolution), explicitly *not* Lane 2 despite concerning paths — left for the SSF-08 owner to sequence as a Lane 1 follow-up or its own tracked item.

Pinned in this PR as a permanent `expect_err` regression, `fa02_039_known_defect_nested_projection_rereads_intermediate_base_as_whole_value`, documenting *current* (buggy) behavior — the test and issue both state explicitly that when FA-02-039 is repaired, this test must be replaced/inverted to assert success, not preserved as desired semantics. The #1663 sibling-field test (`ssf08_1663_projected_record_field_scrutinee_sibling_field_still_usable`) was written against a one-level sibling (`c.other`) specifically to avoid exercising this unrelated defect while still proving path-precision (not a coarser `c` root).

## Explicitly out of scope (not touched, not opportunistically repaired)

#1709, #1718, #1724 (lexical binding identity — Lane 2b/IR concern), #1725, #1726, #1759-#1763, #1778.

## Fallback / hidden-state-loss audit

- No `unwrap_or`, `unwrap_or_default`, `unwrap_or_else`, silent `let Some(...) = ... else { return Ok(()) }`, or `Default::default()`-as-recovery introduced in production code for any of the seven #1664 APIs, **except** the one documented, evidenced exception: `is_const`'s legacy `bool` form keeps its `.unwrap_or(false)` solely because a Lane 2 crate (`sm-ir`) depends on that exact signature (see the corrective-round section above). Grepped the full diff to confirm this is the only remaining occurrence.
- The one `let Some(...) = ... else` in `apply_arm_pattern_capture` (unresolvable scrutinee, #1663) is deliberately branched: `Ok(())` only for a genuine non-path-shaped rvalue, hard `Err` for anything projection-shaped — never a silent skip for a capturing pattern.
- `mark_path_state`'s `Result` is never discarded: all 16 pre-existing test call sites updated to `.expect("test-constructed binding must exist")` rather than clippy's suggested `let _ = ...`. The new `mark_consumed` call site added this round follows the same convention.
- Every `env.clone()` / `push_scope()` / `pop_scope()` introduced has an accounted-for fate: `then_env`/`else_env`/`arm_env`/`def_env` are always either joined via `join_ownership_from` or (branch/arm/loop-local scopes) popped and then joined, never silently dropped. The fixed-point loop's `candidate`/`body_env`/`next_candidate` triple is fully traced in `run_loop_body_to_fixed_point` above.
- Zero new production call sites to `mark_consumed`/`.consumed` — grepped and confirmed; the only call site anywhere is the pre-existing `whole_var_consumed_still_blocks` test.
- `git diff --check`: clean, no whitespace errors.

## Local qualification (all commands run from workspace root)

| Command | Result |
|---|---|
| `cargo test -p sm-front --lib` | **634 passed**, 0 failed (594 pre-Lane-1 baseline + 33 Lane 1 + 7 net-new #1664 corrective-round tests) |
| `cargo check --workspace --all-targets` | clean — proves Lane 2 (`sm-ir`) is genuinely unaffected by leaving `is_const` unchanged |
| `cargo test --test scalar_root_rebind_copy_semantics_e2e` | 6/6 passed — frozen PR #1879 copy-semantics invariant intact |
| `cargo test --workspace` | **4428 passed, 0 failed** — full pass, not marked INCOMPLETE |

## Files changed

- `crates/sm-front/src/lib.rs` — `ScopeEnv` primitives: `require_binding`/`require_binding_mut`, `join_ownership_from`; all seven #1664 APIs now fail-closed except the documented `is_const` Lane 2 exception (which gained a fail-closed `is_const_checked` sibling)
- `crates/sm-front/src/typecheck.rs` — the shared join dispatch for `if`/`match`/`if-let`/loops in both statement and expression form, retirement of the old per-construct plan-application functions, the or-pattern-check unification, the `ensure_const_initializer_safe` existence-ordering fix, 40 new regression tests total (33 Lane 1 + 7 net-new in the #1664 corrective round)
- `crates/sm-front/src/types.rs` — `PatternPath::extend` for scrutinee-path composition

No changes outside `sm-front`.

Closes #1656
Closes #1657
Closes #1658
Closes #1659
Closes #1660
Closes #1661
Closes #1662
Closes #1663

#1664 is **not** closed by this PR — see "#1664 corrective round" above for the precise remaining scope (`is_const`'s signature, blocked on a Lane 2 crate edit this PR cannot make).

Related: filed #1881 (FA-02-039) for the newly-discovered, out-of-scope nested-projection read defect — not fixed here.

### DIFF
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -260,18 +260,24 @@ impl ScopeEnv {
     }
 
     /// Mark a variable as consumed (moved out). Subsequent reads will be rejected.
-    pub fn mark_consumed(&mut self, name: SymbolId) {
-        for scope in self.scopes.iter_mut().rev() {
-            if let Some(binding) = scope.get_mut(&name) {
-                binding.consumed = true;
-                return;
-            }
-        }
+    ///
+    /// SSF-08 Lane 1 (#1664): REQUIRED_BINDING -- fails closed if `name` is
+    /// not a known binding, rather than silently no-op'ing. Zero production
+    /// call sites exist anywhere in the workspace today (whole-root
+    /// "consumed" is inert under Position A's frozen copy-by-value plain
+    /// root-rebinding semantics, see PR #1879); this PR adds none.
+    pub fn mark_consumed(&mut self, name: SymbolId) -> Result<(), crate::types::FrontendError> {
+        self.require_binding_mut(name)?.consumed = true;
+        Ok(())
     }
 
     /// Returns true if the variable has been moved and is no longer available.
-    pub fn is_consumed(&self, name: SymbolId) -> bool {
-        self.binding(name).map(|b| b.consumed).unwrap_or(false)
+    ///
+    /// SSF-08 Lane 1 (#1664): REQUIRED_BINDING -- fails closed if `name` is
+    /// not a known binding, rather than treating "missing" as "not
+    /// consumed". Zero call sites exist anywhere in the workspace today.
+    pub fn is_consumed(&self, name: SymbolId) -> Result<bool, crate::types::FrontendError> {
+        Ok(self.require_binding(name)?.consumed)
     }
 
     /// SSF-08 Lane 1: `true` if `a` denotes the same path as, or a path
@@ -381,6 +387,15 @@ impl ScopeEnv {
     ///
     /// Rejects if any stored path overlaps `access_path` with state `Moved`.
     /// Conservative: borrows are not currently enforced as blocking reads.
+    ///
+    /// SSF-08 Lane 1 (#1664): REQUIRED_BINDING -- fails closed if `name` is
+    /// not a known binding (via `require_binding`), rather than treating
+    /// "missing" as "available". `name` must already be established to
+    /// exist by the caller; the canonical source-level "unknown variable"
+    /// diagnostic must come from that caller's own resolution, not from
+    /// this ownership check -- see the call site in `infer_expr_type`
+    /// (`crates/sm-front/src/typecheck.rs`), which only calls this once
+    /// `env.get(name)` has already confirmed the binding exists.
     pub fn check_path_available(
         &self,
         name: SymbolId,
@@ -398,38 +413,37 @@ impl ScopeEnv {
             path_is_prefix(a, b) || path_is_prefix(b, a)
         }
 
-        if let Some(binding) = self.binding(name) {
-            // Whole-variable consumed takes priority.
-            if binding.consumed {
-                return Err(crate::types::FrontendError {
-                    pos: 0,
-                    message: format!("use of moved value '{}'", name.0),
-                });
-            }
-            for (stored_path, avail) in &binding.path_state {
-                if paths_overlap(stored_path, access_path) {
-                    if *avail == PathAvailability::Moved {
-                        // M9.9 Wave D: more precise diagnostic.
-                        // Distinguish "accessing moved path" from "accessing parent of moved child".
-                        let msg = if path_is_prefix(stored_path, access_path) {
-                            // stored = root.0, access = root.0 or root.0.x → moved path
-                            format!(
-                                "use of moved value: path was moved earlier (moved path {:?})",
-                                stored_path.elems
-                            )
-                        } else {
-                            // stored = root.0, access = root → whole-var after partial move
-                            format!(
-                                "use of partially moved value: cannot use whole variable because \
-                                 child path {:?} was moved",
-                                stored_path.elems
-                            )
-                        };
-                        return Err(crate::types::FrontendError {
-                            pos: 0,
-                            message: msg,
-                        });
-                    }
+        let binding = self.require_binding(name)?;
+        // Whole-variable consumed takes priority.
+        if binding.consumed {
+            return Err(crate::types::FrontendError {
+                pos: 0,
+                message: format!("use of moved value '{}'", name.0),
+            });
+        }
+        for (stored_path, avail) in &binding.path_state {
+            if paths_overlap(stored_path, access_path) {
+                if *avail == PathAvailability::Moved {
+                    // M9.9 Wave D: more precise diagnostic.
+                    // Distinguish "accessing moved path" from "accessing parent of moved child".
+                    let msg = if path_is_prefix(stored_path, access_path) {
+                        // stored = root.0, access = root.0 or root.0.x → moved path
+                        format!(
+                            "use of moved value: path was moved earlier (moved path {:?})",
+                            stored_path.elems
+                        )
+                    } else {
+                        // stored = root.0, access = root → whole-var after partial move
+                        format!(
+                            "use of partially moved value: cannot use whole variable because \
+                             child path {:?} was moved",
+                            stored_path.elems
+                        )
+                    };
+                    return Err(crate::types::FrontendError {
+                        pos: 0,
+                        message: msg,
+                    });
                 }
             }
         }
@@ -507,16 +521,39 @@ impl ScopeEnv {
         self.binding(name).map(|binding| binding.ty.clone())
     }
 
+    /// SSF-08 Lane 1 (#1664): kept fail-open (`false` on a missing binding)
+    /// -- unlike every other #1664 API, this exact bool-returning shape has
+    /// live production call sites in `crates/sm-ir/src/legacy_lowering.rs`,
+    /// a Lane 2 crate this PR is not permitted to modify (SSF-08 Lane 1's
+    /// scope is `crates/sm-front` only). Changing this signature would
+    /// break that crate's build. Every in-crate (`sm-front`) production
+    /// call site now uses the fail-closed `is_const_checked` below instead;
+    /// this method is retained solely for that external, out-of-scope
+    /// dependency and must gain no new call sites within `sm-front`.
     pub fn is_const(&self, name: SymbolId) -> bool {
         self.binding(name)
             .map(|binding| binding.is_const)
             .unwrap_or(false)
     }
 
-    pub fn is_mutable(&self, name: SymbolId) -> bool {
-        self.binding(name)
-            .map(|binding| binding.is_mutable)
-            .unwrap_or(false)
+    /// SSF-08 Lane 1 (#1664): fail-closed counterpart of `is_const`, used by
+    /// every `sm-front`-internal call site. See `is_const`'s doc comment
+    /// for why that original bool-returning method could not simply be
+    /// changed in place.
+    pub(crate) fn is_const_checked(
+        &self,
+        name: SymbolId,
+    ) -> Result<bool, crate::types::FrontendError> {
+        Ok(self.require_binding(name)?.is_const)
+    }
+
+    /// SSF-08 Lane 1 (#1664): REQUIRED_BINDING -- fails closed if `name` is
+    /// not a known binding, rather than treating "missing" as "not
+    /// mutable". Zero call sites exist anywhere in the workspace today, so
+    /// unlike `is_const` this signature carries no external Lane 2
+    /// dependency and can change freely.
+    pub fn is_mutable(&self, name: SymbolId) -> Result<bool, crate::types::FrontendError> {
+        Ok(self.require_binding(name)?.is_mutable)
     }
 
     fn binding(&self, name: SymbolId) -> Option<&ScopeBinding> {
@@ -576,11 +613,13 @@ impl ScopeEnv {
             }
         }
         for depth in 0..self.scopes.len() {
-            // Sorted so the (should-be-unreachable, invariant-defense-only)
-            // "missing from a branch successor" error below always names
-            // the same binding first, independent of this HashMap's
-            // iteration order -- no new diagnostic here may be
-            // hash-order-dependent.
+            // `scopes` is a `BTreeMap`, so `.keys()` is already sorted and
+            // this is a no-op today -- kept explicit (rather than relying
+            // on the underlying collection type) so the (should-be-
+            // unreachable, invariant-defense-only) "missing from a branch
+            // successor" error below always names the same binding first
+            // even if that collection type ever changes; no diagnostic
+            // here may be iteration-order-dependent.
             let mut names: Vec<SymbolId> = self.scopes[depth].keys().copied().collect();
             names.sort_unstable();
             for name in names {
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -274,45 +274,107 @@ impl ScopeEnv {
         self.binding(name).map(|b| b.consumed).unwrap_or(false)
     }
 
-    /// M9.7: Record that `path` within variable `name` has been moved or borrowed.
+    /// SSF-08 Lane 1: `true` if `a` denotes the same path as, or a path
+    /// leading to, `b` (i.e. `a` is a prefix of `b`'s element list).
+    fn path_is_prefix(a: &crate::types::PatternPath, b: &crate::types::PatternPath) -> bool {
+        if a.elems.len() > b.elems.len() {
+            return false;
+        }
+        a.elems.iter().zip(&b.elems).all(|(x, y)| x == y)
+    }
+
+    /// SSF-08 Lane 1: shared normalisation core of `mark_path_state`, factored
+    /// out so the branch/loop/match join (`join_ownership_from`) can build a
+    /// joined `path_state` using the exact same compaction rules, rather than
+    /// re-deriving them.
+    ///
+    /// Rule 1 — new path subsumes longer existing entries of the same state:
+    ///   e.g. adding Moved(root) while Moved(root.0) exists → drop root.0.
+    /// Rule 2 — if an existing entry already covers the new path (same state,
+    ///   existing is a prefix of new path), the new entry is redundant.
+    ///
+    /// Deliberately does not merge across *different* states: `(p, Moved)`
+    /// and `(p, Borrowed)` may both be present at once. `check_path_available`
+    /// / `check_capture_allowed` treat that as "restricted under either
+    /// reading", which is exactly the conservative behavior a branch join
+    /// needs when one successor moved `p` and another only borrowed it.
+    fn push_path_state_normalized(
+        path_state: &mut Vec<(crate::types::PatternPath, crate::types::PathAvailability)>,
+        path: crate::types::PatternPath,
+        state: crate::types::PathAvailability,
+    ) {
+        path_state.retain(|(existing, existing_state)| {
+            if *existing_state != state {
+                return true;
+            }
+            !Self::path_is_prefix(&path, existing)
+        });
+        let redundant = path_state.iter().any(|(existing, existing_state)| {
+            *existing_state == state && Self::path_is_prefix(existing, &path)
+        });
+        if !redundant {
+            path_state.push((path, state));
+        }
+    }
+
+    /// SSF-08 Lane 1: look up a binding that the caller has already
+    /// established must exist (e.g. because an earlier, independently
+    /// successful lookup on the same `name` already proved it, or because
+    /// `name` is drawn from `self`'s own known binding set). Fails closed:
+    /// unlike `binding`, absence here is treated as an internal ownership
+    /// state invariant failure, not a legitimate "unknown variable" case.
+    /// This is deliberately a *different* error from the source-level
+    /// "unknown variable 'x'" diagnostic (see `Expr::Var` handling in
+    /// `typecheck.rs`) — callers that have not already independently
+    /// confirmed existence must not use this.
+    fn require_binding(
+        &self,
+        name: SymbolId,
+    ) -> Result<&ScopeBinding, crate::types::FrontendError> {
+        self.binding(name)
+            .ok_or_else(|| crate::types::FrontendError {
+                pos: 0,
+                message: format!(
+                    "internal ownership state: required binding {} is missing",
+                    name.0
+                ),
+            })
+    }
+
+    /// Mutable counterpart of `require_binding`. See its doc comment for the
+    /// fail-closed contract.
+    fn require_binding_mut(
+        &mut self,
+        name: SymbolId,
+    ) -> Result<&mut ScopeBinding, crate::types::FrontendError> {
+        for scope in self.scopes.iter_mut().rev() {
+            if scope.contains_key(&name) {
+                return Ok(scope.get_mut(&name).expect("just checked contains_key"));
+            }
+        }
+        Err(crate::types::FrontendError {
+            pos: 0,
+            message: format!(
+                "internal ownership state: required binding {} is missing",
+                name.0
+            ),
+        })
+    }
+
+    /// M9.7 / SSF-08 Lane 1: Record that `path` within variable `name` has
+    /// been moved or borrowed. `name` must already be a known binding —
+    /// callers that have not independently established this (e.g. a bare
+    /// source-level read where "unknown variable" must remain the reported
+    /// diagnostic) must not call this directly; see `require_binding_mut`.
     pub fn mark_path_state(
         &mut self,
         name: SymbolId,
         path: crate::types::PatternPath,
         state: crate::types::PathAvailability,
-    ) {
-        use crate::types::PatternPath;
-
-        fn path_is_prefix(a: &PatternPath, b: &PatternPath) -> bool {
-            if a.elems.len() > b.elems.len() {
-                return false;
-            }
-            a.elems.iter().zip(&b.elems).all(|(x, y)| x == y)
-        }
-
-        for scope in self.scopes.iter_mut().rev() {
-            if let Some(binding) = scope.get_mut(&name) {
-                // M9.9 Wave C: normalise path-state to keep it compact.
-                //
-                // Rule 1 — new path subsumes longer existing entries of the same state:
-                //   e.g. adding Moved(root) while Moved(root.0) exists → drop root.0.
-                binding.path_state.retain(|(existing, existing_state)| {
-                    if *existing_state != state {
-                        return true;
-                    }
-                    !path_is_prefix(&path, existing)
-                });
-                // Rule 2 — if an existing entry already covers the new path (same state,
-                //   existing is a prefix of new path), the new entry is redundant.
-                let redundant = binding.path_state.iter().any(|(existing, existing_state)| {
-                    *existing_state == state && path_is_prefix(existing, &path)
-                });
-                if !redundant {
-                    binding.path_state.push((path, state));
-                }
-                return;
-            }
-        }
+    ) -> Result<(), crate::types::FrontendError> {
+        let binding = self.require_binding_mut(name)?;
+        Self::push_path_state_normalized(&mut binding.path_state, path, state);
+        Ok(())
     }
 
     /// M9.7: Check that accessing `access_path` within `name` is allowed.
@@ -374,8 +436,13 @@ impl ScopeEnv {
         Ok(())
     }
 
-    /// M9.8: Check that a new capture of `path` with `capture` mode is compatible
-    /// with the existing path-state of variable `name`.
+    /// M9.8 / SSF-08 Lane 1: Check that a new capture of `path` with `capture`
+    /// mode is compatible with the existing path-state of variable `name`.
+    /// `name` must already be a known binding (see `require_binding`) — this
+    /// is a pattern-capture check, always run against a scrutinee whose own
+    /// existence was already independently confirmed by successfully
+    /// typechecking the scrutinee expression before any pattern plan is
+    /// built.
     ///
     /// Rules:
     ///   prior Borrowed + new Move   → error ("cannot move from borrowed value")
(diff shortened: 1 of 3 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

Issue #1726, opened 2026-08-15
FA-04-020: OWN0 loses Borrow/Write event timing and VM activates every borrow at frame entry


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P1
Classification: SEMANTIC-DRIFT · FALSE-READY · IDENTITY/TRUST-GAP

Finding: the frozen runtime ownership contract says the frontend decides where borrow capture exists, lowering preserves Borrow/Write event intent and deterministic event order, and an activated borrow then remains active for the frame.

`OWN0` serializes events in order, but decoding immediately splits them into two independent vectors (`borrowed_paths` and `write_paths`), losing the original Borrow/Write interleaving. VM `push_frame()` then initializes the frame with `borrowed_paths: f.borrowed_paths.clone()`, making every borrow active before the first instruction executes. There is no PC/event-position marker and no runtime transition that activates a borrow at the source/lowered point where it occurred.

Consequences:
- a write that is semantically before a later borrow can be rejected as if the borrow already existed;
- total event order promised by the transport contract is not preserved into execution;
- the artifact cannot faithfully represent “borrow begins here, remains until frame exit”.

This is independent of FA-04-019's SymbolId mismatch: even with perfect identity remapping, temporal ownership meaning is lost.

Evidence:
- crates/sm-ir/src/legacy_lowering.rs::OwnershipPathEvent and OWN0 emission
- crates/sm-format/src/semcode_decode.rs ownership decoder splitting events by kind
- sm-vm Frame / push_frame initialization and next_write_path handling
- docs/spec/runtime_ownership.md Frontend And Lowering Contract, SemCode Transport Contract, Borrow lifetime v0

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
### Module 07 (`sm-verify`) cross-boundary evidence

Frozen audit SHA: `4ac4d1a169902777e760e23eaee92aa233edc04d`.

The verifier checks OWN0 structure and capability presence but has no representation of ownership-event PC/timing, because the decoded envelope has already split the section into `borrowed_paths` and `write_paths`. It therefore issues `VerifiedSemCode` without being able to prove the documented event-order/timing contract.

This is the same transport root as FA-04-020, so no duplicate FA-07 issue was created.

Phase A evidence only; no repair.

skulmakov-oss:
[...]

Issue #1725, opened 2026-08-15
FA-04-019: OWN0 roots serialize frontend SymbolId but VM interprets them as function string-table indexes


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P1
Classification: IDENTITY/TRUST-GAP · SEMANTIC-DRIFT · FALSE-READY

Finding: `AccessPath.root` is a frontend `sm_front::SymbolId`. `emit_ownership_events()` serializes it directly with `write_u32_le(out, event.path.root.0)`.

The function string table is built independently by `StringInterner`, assigning dense local u16 ids in first textual-use order for `LoadVar`/`StoreVar` and other string operands. No mapping from frontend SymbolId to this local string-table id is constructed for OWN0.

During VM load, decoded ownership roots are interpreted as local string-table indexes:

```rust
let local_root = path.root_symbol_id as usize;
let root = symbol_ids
    .get(local_root)
    .copied()
    .unwrap_or(SymbolId(path.root_symbol_id));
```

where `symbol_ids` is built from the function string table. Thus the same numeric field has two incompatible meanings across the producer/consumer boundary.

Impact: Borrow/Write metadata can be attached to the wrong runtime local, or fall back to an unrelated raw runtime SymbolId, while the artifact remains structurally decodable/verifiable. Runtime borrow/write conflict enforcement can therefore fail to protect the source binding whose ownership event was emitted.

The same identity concern applies to field/variant SymbolId-bearing ownership components, which are also serialized as raw frontend ids rather than through a demonstrated artifact-local identity map.

Evidence:
- crates/sm-ir/src/legacy_lowering.rs::AccessPath
- crates/sm-ir/src/legacy_lowering.rs::emit_ownership_events
- crates/sm-ir/src/legacy_lowering.rs::StringInterner / emit_semcode_function
- sm-vm SemCode load path `symbol_ids` + `remap_paths`
- docs/spec/runtime_ownership.md ownership preservation/enforcement contract

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
### Module 07 (`sm-verify`) cross-boundary evidence

Frozen audit SHA: `4ac4d1a169902777e760e23eaee92aa233edc04d`.

[...]

Issue #1724, opened 2026-08-15
FA-04-018: lowering flattens lexical scopes into one string-local namespace and can leak shadowed bindings


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P1
Classification: SEMANTIC-DRIFT · FALSE-READY · IDENTITY/TRUST-GAP

Finding: source/typecheck uses nested `ScopeEnv` scopes for branches, match arms and loop bodies, but IR local transport is plain `StoreVar { name: String }` / `LoadVar { name: String }` into one function-frame local namespace. Lowering clones/pushes compile-time scopes but does not mangle lexical binding identity or restore an outer runtime binding after a nested scope exits.

Therefore an admitted nested shadow such as an outer `x` plus branch-local/arm-local/loop-local `let x = ...` can emit `StoreVar "x"` for the inner binding. After the nested scope exits, later `LoadVar "x"` resolves to the inner runtime value even though the compile-time environment again resolves the outer binding.

Claim source: `docs/spec/source_semantics.md` defines lexical block scoping and branch-local scopes. The frontend `ScopeEnv` models those scopes, so lowering must preserve that identity distinction.

Cross-phase impact:
frontend resolves outer/inner `SymbolId` by lexical scope → sm-ir erases scope identity to one textual local name → VM observes a different binding after scope exit.

Evidence:
- sm-front `ScopeEnv` push/pop/binding behavior
- crates/sm-ir/src/legacy_lowering.rs branch/match/loop lowering with scoped envs
- IrInstr::StoreVar / LoadVar textual names
- sm-vm frame locals keyed by symbolized string identity
- docs/spec/source_semantics.md Scope And Binding Rules

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
Reopened: the 2026-09-02 closure was not backed by an implementation repair. The associated `c48882a` merge is the SSF-08 Position A architecture decision and explicitly states that no production ownership repair was performed; the same commit routes #1724 into the SSF-08 repair DAG as Lane 2b ahead of #1725. Restoring the issue to OPEN so its state matches the accepted dependency record.

Issue #1718, opened 2026-08-15
FA-04-012: sm-ir silently emits ADT/sequence ownership paths under tuple-only SEMCOD11


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir
Audit HEAD: 4ac4d1a169902777e760e23eaee92aa233edc04d
Severity: P1
Classification: SEMANTIC-DRIFT · IDENTITY/TRUST-GAP · FALSE-READY

Finding: `IrFunction.ownership_events` can contain `PathComponent::AdtPayload` and `PathComponent::SequenceIndexStatic`, and `emit_ownership_events()` serializes both. Header selection, however, only distinguishes any ownership (`SEMCOD11`) from record Field ownership (`SEMCOD12`); there is no version/capability promotion for ADT or sequence path components. Therefore an artifact containing these newer component kinds can be emitted under `SEMCOD11`.

Current verifier decoding structurally accepts both ADT payload and static sequence-index ownership components, and sm-vm converts the decoded ADT path into runtime `AccessPath`, so this is not merely dead vocabulary.

Claim source: `docs/spec/semcode.md` freezes SEMCOD11 as tuple-only ownership transport and the No Silent Mutation Rule forbids changing an existing header's section interpretation without a documented version change. `docs/spec/runtime_ownership.md` likewise limits its frozen current slice to tuple+record.

Gap: the meaning/admission surface of an existing header family has been widened without a matching header/capability/version contract.

Evidence:
- crates/sm-ir/src/legacy_lowering.rs::has_v11_ownership_events
- has_v12_record_field_ownership_events
- emit_ownership_events
- sm-verify ownership component admission
- sm-vm decoded ownership path conversion
- docs/spec/semcode.md
- docs/spec/runtime_ownership.md

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
Module 05 (`sm-format`) Phase A adds direct decoder-test evidence for FA-04-012 at frozen audit HEAD `4ac4d1a169902777e760e23eaee92aa233edc04d`.

`crates/sm-format/src/semcode_decode.rs` has an in-crate helper/test that deliberately constructs `MAGIC11` + `OWN0` + `OWNERSHIP_PATH_COMPONENT_SEQUENCE_INDEX` and asserts that `decode_semcode_envelope()` accepts it as `SequenceIndexStatic(7)`.

[...]

PR #1882, opened 2026-09-02
fix(front): preserve exact ownership path for nested projection reads

### LINKED ISSUE #1881: FA-02-039: nested projection reads re-check an intermediate base as a whole-value access
Umbrella: #1617
Module: **02 — `sm-front`**
Classification: **SEMANTIC-DRIFT · FALSE-READY**
Severity: **P2 — false rejection of valid, ownership-safe reads (over-restrictive, not a soundness hole)**
Audit HEAD: `c48882a28b0302209f571f05fd743ca6647c491d`

## Finding

Reading a two-level-nested field projection spuriously rejects when only a *sibling* field of the one-level-intermediate base was moved:

```rust
record Pair { a: i32, b: i32 }
record Container { pair: Pair, other: i32 }

fn main() {
    let c: Container = Container { pair: Pair { a: 1, b: 2 }, other: 9 };
    let Pair { a: moved, b: _ } = c.pair;   // moves only c.pair.a
    let _ = moved;
    let z: i32 = c.pair.b;                  // c.pair.a and c.pair.b are siblings -- must be fine
    let _ = z;
    return;
}
```

This rejects with `use of partially moved value: cannot use whole variable because child path ["pair", "a"] was moved` even though the read is of `c.pair.b`, a sibling of the moved path, not of `c.pair.a` or of `c.pair` as a whole value.

**Root cause** — `crates/sm-front/src/typecheck.rs`:

- `infer_record_field_access_type` (~line 15156) resolves a field access's base via `infer_expr_type_no_check(field_expr.base, ...)` (~line 15168), on the stated assumption ("use no-check variant for the base; caller already verified full path") that no further ownership check is needed for the base.
- `infer_expr_type_no_check` (~line 16635) only honors that assumption when the base is a bare `Expr::Var` (~line 16647, "No path check here; the outer infer_expr_type call for the full field/index expression already checked the correct sub-path"). Any other base expression kind falls through to its `_ =>` arm, which calls the **full** `infer_expr_type` (~line 16655) — reintroducing the top-of-function `check_path_available` gate (`crates/sm-front/src/typecheck.rs`, currently at ~line 2425 in `infer_expr_type`; the `check_path_available` method itself is defined in `crates/sm-front/src/lib.rs`) for that intermediate base.
- For `c.pair.b`, the base of the outer `RecordField` access is itself `c.pair` — an `Expr::RecordField`, not an `Expr::Var`. So resolving its type re-triggers `check_path_available(c, [pair])` as an independent, shorter-path check, which overlaps with the stored `[pair, a]` = Moved restriction and is misreported as "reading `c.pair` as a whole value," when it is actually only being resolved for its type en route to projecting through to `.b`.

The outer expression's own top-level check (`check_path_available(c, [pair, b])`) would correctly find no conflict with `[pair, a]` (true siblings) — the bug is that a second, independent, incorrectly-scoped check on the intermediate base runs first and rejects before the outer check is ever reached.

## Independence from SSF-08 Lane 1

This is a pre-existing, previously-undiscovered defect, not introduced by Lane 1's control-flow join model (skulmakov-oss/Semantic#1880). It reproduces on a **straight-line program with zero `if`/`loop`/`match`** — the isolation test below has no control flow at all — proving it is orthogonal to Lane 1's actual scope (branch/loop/match ownership-state joining). Lane 1 does not modify `infer_expr_type_no_check`, `infer_record_field_access_type`, or the restriction-matching logic inside `check_path_available`; it only changed how existence of the root binding is established before that logic runs (skulmakov-oss/Semantic#1880, #1664 corrective round), which is unrelated to this bug.

## Expected sibling-path semantics

Reading `c.pair.b` after only `c.pair.a` was moved must succeed: `pair.a` and `pair.b` are sibling paths under `pair`, and neither is a prefix of the other. Only a read that actually overlaps a moved path (the moved path itself, or an ancestor/descendant of it) should be rejected.

## Current (buggy) behavior

The read is rejected with a "partially moved" diagnostic, as if `c.pair` itself were being read as a whole value, even though only `.b` is ultimately being projected out.

## Permanent regression test

`crates/sm-front/src/typecheck.rs`, `typecheck::tests::fa02_039_known_defect_nested_projection_rereads_intermediate_base_as_whole_value` (added in skulmakov-oss/Semantic#1880, named after this issue) pins the *current* (buggy) behavior via `expect_err` with an explicit comment marking it as documentation of this pre-existing defect, not desired language semantics. **When this issue is repaired, that test must be replaced or inverted (asserting success), not preserved as-is** — its continued `expect_err` after a fix would itself be a stale/incorrect assertion.

## Explicit statement

skulmakov-oss/Semantic#1880 (SSF-08 Lane 1) does **not** repair this defect. It was found while writing that PR's #1663 regression tests, isolated, confirmed independent of Lane 1's changes, and deliberately left unfixed per that PR's scope (control-flow ownership-state joining only, not read-path base resolution).

## Routing

This is a frontend typecheck/read-path-validation defect (an incorrect base-resolution re-check), not a Lane 2 (IR/runtime) path-tracking concern — it belongs to the `sm-front` ownership/read-path track inside SSF-08, not Lane 2, even though it concerns paths. Suggested routing: SSF-08 follow-up work distinct from Lane 1 (Lane 1 is control-flow joining; this is single-expression base resolution) — left to the SSF-08 owner to sequence as either a Lane 1 follow-up or its own tracked item.

## Required repair (for whoever picks this up)

`infer_expr_type_no_check`'s skip condition needs to cover any base expression whose own top-level path was already subsumed by the caller's path check, not just a bare `Expr::Var` base — e.g. recognize a `RecordField`/`SequenceIndex` base as "already covered" the same way a `Var` base is, rather than re-deriving and re-checking a shorter path for it.

## Regression (for the eventual fix)

- the reproduction above (`c.pair.b` after moving only `c.pair.a`) must type-check successfully;
- a read that genuinely does overlap a moved ancestor/descendant path must still reject (no regression of the existing partial-move protection);
- `fa02_039_known_defect_nested_projection_rereads_intermediate_base_as_whole_value` must be updated to assert success instead of `expect_err`, with its documentation-of-a-bug framing removed.

skulmakov-oss:
SSF-08 frontend ownership/read-path follow-up.

Closes #1881.

Fixes a false rejection in nested projection reads where type resolution re-checked an intermediate base as a standalone whole-value access even though the caller had already validated the complete ownership path.

Position A unchanged. No Lane 2 changes. No #1664 changes.

## Baseline / HEAD

- Expected baseline: `ff3e65972adacf19c101207b27ad9d59f8c62001` (confirmed exact match: `git fetch origin && git rev-parse origin/main`, working tree clean before branching).
- This branch: `fix/1881-nested-projection-read`.

## Reproduction

```rust
record Pair { a: i32, b: i32 }
record Container { pair: Pair, other: i32 }

fn main() {
    let c: Container = Container { pair: Pair { a: 1, b: 2 }, other: 9 };
    let Pair { a: moved, b: _ } = c.pair;
    let _ = moved;
    let z: i32 = c.pair.b;
    let _ = z;
    return;
}
```

- `origin/main @ ff3e6597` (via the pre-fix `fa02_039_known_defect_...` test, which asserted `expect_err`): **FAILS incorrectly** — `use of partially moved value: cannot use whole variable because child path ["pair", "a"] was moved`, even though `c.pair.a` and `c.pair.b` are sibling paths.
- This branch's HEAD (via the same test, inverted to `fa02_039_nested_projection_sibling_read_remains_available`, now `.expect(...)`): **PASSES correctly**.

## Root cause (verified against this exact baseline before editing, per the issue's own checklist)

1. Confirmed the call path still exists exactly as #1881 describes: `infer_record_field_access_type` (`crates/sm-front/src/typecheck.rs`) resolves a field access's base via `infer_expr_type_no_check(field_expr.base, ...)`, whose comment claimed "no path check needed; the outer `infer_expr_type` call for the full field/index expression already checked the correct sub-path" — but `infer_expr_type_no_check` only actually honored that for a bare `Expr::Var` base; any other kind (including another `RecordField`) fell through to the fully-checked `infer_expr_type`.
2. Reproduced #1881 on this exact baseline (confirmed above).
3. Confirmed the *outer* full-path check for `c.pair.b` (`check_path_available(c, [pair, b])`) correctly finds no overlap with the stored `[pair, a] = Moved` restriction (true siblings, neither a prefix of the other) — traced by hand through `check_path_available`'s prefix/overlap logic, and confirmed empirically with temporary instrumentation (see below).
4. Confirmed the false rejection comes *only* from the second, intermediate call: resolving `c.pair`'s own type (as the base of `c.pair.b`) re-invokes the full, checked `infer_expr_type(c.pair)`, whose own top-of-function check re-derives and re-checks the shorter path `[pair]`, which *does* overlap `[pair, a]` as an ancestor — producing the spurious rejection.

None of the checklist's "if false, STOP" conditions were triggered; proceeded to the fix.

## Repair mechanism

`infer_expr_type_no_check` now recursively preserves its no-ownership-recheck contract for **any** base expression `expr_access_path` itself recognizes as part of the admitted static path family (`Expr::Var`, `Expr::RecordField`, `Expr::SequenceIndex` with a literal non-negative index) — not only a bare `Expr::Var`. For `RecordField`/`SequenceIndex` bases whose own `expr_access_path` resolves (proving they are themselves a segment of a chain whose full path the true top-level caller already validated), it now calls the existing `infer_record_field_access_type`/`infer_sequence_index_type` resolvers directly — the exact same functions the checked path already used, which themselves recurse into `infer_expr_type_no_check` for *their* base. This is **one canonical no-check traversal**, not a special case per expression kind: no new field/index-resolution logic was written; the fix is entirely in which function `infer_expr_type_no_check` dispatches to.

Anything **not** recognized by `expr_access_path` (a dynamic sequence index, a call expression, etc.) still falls through to the normal, fully-checked `infer_expr_type`, unchanged from before — this is not "skip the check when unsure." An unprovable/dynamic path stays fail-closed via the ordinary checked path exactly as it did pre-fix.

`ScopeEnv::check_path_available` and its prefix/overlap semantics in `crates/sm-front/src/lib.rs` are **untouched** — the issue's own evidence said the bug was not there, and empirical proof (below) confirms it. `lib.rs` has zero diff in this PR.

### Empirical proof of "single ownership check authority" (section 19)

Temporarily instrumented `check_path_available` with an `eprintln!` logging `(name, path)` on every call, ran the primary reproduction, and removed the instrumentation before this commit. Observed exactly two calls for the whole program: one for `[pair]` (the *separate*, legitimate destructuring-scrutinee check on `let Pair { .. } = c.pair;`), and one for `[pair, b]` (the `c.pair.b` read itself). Zero synthetic re-checks on the intermediate `[pair]` base for the read — confirming the fix produces exactly the desired control flow (`derive canonical full path → check once → resolve remaining type without re-checking shorter synthetic bases`), not the old `check [pair,b], check [pair]` double-check.

## Positive / negative regression matrix (`crates/sm-front/src/typecheck.rs`)

| Case | Test | Result |
|---|---|---|
| Two-level sibling: move `c.pair.a`, read `c.pair.b` | `fa02_039_nested_projection_sibling_read_remains_available` (inverted from the pre-existing known-defect test) | PASS |
| Three-level chain sibling: move `root.a.b.x`, read `root.a.b.y` | `fa02_039_three_level_chain_sibling_read_remains_available` | PASS |
| Disjoint branches under a common ancestor: move `root.a.x`, read `root.b.y` | `fa02_039_disjoint_branch_under_common_ancestor_remains_available` | PASS |
| Moved ancestor field blocks descendant read: move whole `c.pair`, read `c.pair.b` | `fa02_039_moved_ancestor_field_blocks_descendant_read` | REJECT (unchanged) |
| Exact-path re-read: move `c.pair.b`, read `c.pair.b` | `fa02_039_move_then_reread_exact_path_still_rejects` | REJECT (unchanged) |
| **Critical boundary case** — moved child still blocks a *genuine* whole-parent read: move `c.pair.a`, then `let whole: Pair = c.pair;` (a real whole-value read, reached via the checked path directly, never through `infer_expr_type_no_check`) | `fa02_039_moved_child_still_blocks_genuine_whole_parent_read` | REJECT (unchanged — proves the fix is bounded to intermediate-base resolution only) |
| Deeper ancestor/descendant overlap: move whole `root.a`, read deep descendant `root.a.b.x` | `fa02_039_moved_mid_level_ancestor_blocks_deep_descendant_read` | REJECT (unchanged) |
| Mixed projection families, type-resolution only: `holder.items[0].label` (`RecordField → SequenceIndex(literal) → RecordField`) | `fa02_039_mixed_record_sequence_record_chain_type_resolves` | PASS |
| Dynamic index base structurally excluded from the new routing: `items[i].value` with non-literal `i` | `fa02_039_dynamic_sequence_index_base_still_type_checks` | PASS (still routes through the fully-checked path, unchanged) |
| Dynamic/unrepresentable sequence-index scrutinee (pre-existing #1663 coverage) | `ssf08_1663_dynamically_indexed_sequence_scrutinee_with_capture_rejects` | kept green, unmodified |
| #1663 projected-scrutinee tracking (pre-existing) | `ssf08_1663_projected_record_field_scrutinee_move_is_tracked`, `ssf08_1663_projected_record_field_scrutinee_sibling_field_still_usable` | kept green, unmodified (only a stale comment referencing the old FA-02-039 test name was corrected) |

## A finding surfaced during this audit — filed separately, not fixed here

While constructing the mixed-projection matrix (section 17), I needed a way to move a specific `Sequence` element via pattern capture to build meaningful negative cases. Investigating this surfaced that **`ssf08_1663_dynamically_indexed_sequence_scrutinee_with_capture_rejects` (merged in #1880) does not test what its name and comment claim.**

Its assertion is only `!err.message.contains("unknown variable")`. Empirically (temporary instrumentation, since reverted), the program `match seq[i] { v => { let _ = v; } }` — with `i` dynamic — actually fails to **parse** at all: `"expected '::' in enum match pattern"`. This is because `MatchPattern` (`crates/sm-front/src/types.rs`) has no "bind by plain name" variant at all — only `Quad`, `Adt` (enum, requires `Name::Variant`), `Wildcard`, `IntRange`, `Or`. A bare identifier like `v` as a match arm pattern is not valid syntax against *any* scrutinee type, regardless of whether the underlying path is static or dynamic. The test's weak assertion happens to also hold for this parse error, so it reports green while proving nothing about ownership-path rejection specifically.

Filed as **#1883 (FA-02-040: "projected-scrutinee negative regression passes at parser layer and does not exercise ownership rejection")**, umbrella #1617, Module 02 (`sm-front`), Classification TEST-INTEGRITY · FALSE-READY · QUALIFICATION-GAP. Explicitly does **not** reopen #1663 (this finding proves one of its qualifying regressions is invalid evidence, not that its implementation is wrong) and explicitly does **not** belong in this PR. I did not fix or touch the test — it's out of #1881's scope, and #1880's own boundary explicitly required keeping it green and unmodified. It also means: there is currently no admitted source syntax to move a specific `Sequence` element via pattern capture at all (record/tuple destructuring support bind-by-name via a different, separate pattern representation; `match`/`if-let` do not) — which is why the mixed-projection matrix above is positive/type-resolution-only for sequence chains, not move/reject, and why I did not attempt a `sequence[index].field`-with-a-move negative case.

## Explicitly out of scope (not touched, not opportunistically repaired)

- `#1664` remainder (`is_const`, `sm-ir` migration) — not touched.
- Lane 2: `#1709`, `#1718`, `#1724`, `#1725`, `#1726` — not touched. No production changes under `crates/sm-ir`, `crates/sm-format`, `crates/sm-verify`, `crates/sm-vm`, `crates/sm-runtime-core`.
- `#1883` (FA-02-040, the `ssf08_1663_dynamically_indexed_sequence_scrutinee_with_capture_rejects` test-integrity finding above) — filed, not fixed here.
- Position A unchanged; plain root rebinding (`let y = x;`) stays copy-by-value (`scalar_root_rebind_copy_semantics_e2e.rs`, still 6/6).

## Fallback / hidden-behavior audit

Searched the diff for `unwrap_or`, `unwrap_or_default`, `Ok(())` on an unresolved path, silent `None => continue`, a fallback to full `infer_expr_type` used as an uncertainty-recovery mechanism, `Default::default()` as recovery, and ignored `Result` — none introduced. The `_ => infer_expr_type(...)` fallthrough arm is the pre-existing, intentional fail-closed default for anything `expr_access_path` doesn't recognize, unchanged in shape from before this PR (only the two new guarded arms sit above it).

## Files changed

- `crates/sm-front/src/typecheck.rs` only (+265/-37): `infer_expr_type_no_check`'s recursive fix, its updated doc comment, the inverted/renamed `fa02_039_*` test plus 8 new regression tests, and a corrected stale comment in one pre-existing #1663 test.
- `crates/sm-front/src/lib.rs`: **zero diff.** No `ScopeEnv` semantic change was needed or made.

## Local qualification (all commands from workspace root)

| Command | Result |
|---|---|
| `cargo test -p sm-front --lib` | **642 passed**, 0 failed (634 baseline + 8 net-new) |
| `cargo test --workspace` | **4436 passed, 0 failed** — full pass, not INCOMPLETE |

Do not merge. Draft PR pending review, matching the exact-head discipline established on #1879/#1880.

Closes #1881

### DIFF
--- a/crates/sm-front/src/typecheck.rs
+++ b/crates/sm-front/src/typecheck.rs
@@ -11095,31 +11095,24 @@ mod tests {
     }
 
     #[test]
-    fn fa02_039_known_defect_nested_projection_rereads_intermediate_base_as_whole_value() {
-        // Tracked as FA-02-039 (skulmakov-oss/Semantic#1881). NOT part of
-        // #1656-#1664, NOT fixed by this PR. Documents a pre-existing
-        // frontend defect found while writing the #1663 regression suite
-        // (SSF-08 Lane 1): reading a two-level-nested field projection
-        // (`c.pair.b`) spuriously rejects if a *sibling* field of the
-        // intermediate base was moved (`c.pair.a`), because the
-        // intermediate base (`c.pair`) is independently re-checked as if it
-        // were being read as a whole value, not merely projected through.
-        // Isolated to a straight-line program with zero if/loop/match and
-        // zero Lane 1 join-model code involved at all -- the root cause is
-        // in `infer_expr_type_no_check`, which only skips its own
-        // path-availability re-check when the base expression is a bare
-        // `Expr::Var`; a `RecordField` base (as in `c.pair.b`, whose base
-        // `c.pair` is itself a `RecordField` access, not a `Var`) falls
-        // through to a second, independent, shorter-path check that the
-        // outer expression's own top-level check has already correctly
-        // subsumed. This is a read-path-checking gap orthogonal to
-        // control-flow/branch-joining (Lane 1's actual scope) -- it would
-        // misfire in a single straight-line function with no
-        // `if`/`loop`/`match` anywhere. This test pins the *current*
-        // (buggy) behavior so it does not regress silently; when FA-02-039
-        // is repaired, this test must be replaced or inverted to assert
-        // success, not preserved as-is -- its continued `expect_err` after
-        // a fix would itself be a stale, incorrect assertion.
+    fn fa02_039_nested_projection_sibling_read_remains_available() {
+        // Fixes FA-02-039 (skulmakov-oss/Semantic#1881). Not part of
+        // #1656-#1664 (SSF-08 Lane 1's own scope); this is a follow-up
+        // frontend read-path repair. Reading a two-level-nested field
+        // projection (`c.pair.b`) must succeed after only a *sibling*
+        // field of the intermediate base was moved (`c.pair.a`): `pair.a`
+        // and `pair.b` are sibling paths under `pair`, and neither is a
+        // prefix of the other. Root cause was `infer_expr_type_no_check`
+        // only preserving its no-ownership-recheck contract for a bare
+        // `Expr::Var` base; a `RecordField` base (as in `c.pair.b`, whose
+        // own base `c.pair` is itself a `RecordField` access) fell through
+        // to the fully-checked `infer_expr_type`, re-deriving and
+        // re-checking a shorter, synthetic `[pair]` path that the outer
+        // expression's own top-level `[pair, b]` check had already
+        // correctly subsumed. `infer_expr_type_no_check` now recurses
+        // through the same no-check-preserving field/index resolvers for
+        // any base `expr_access_path` itself recognizes as part of the
+        // admitted static path family, not only `Expr::Var`.
         let src = r#"
             record Pair { a: i32, b: i32 }
             record Container { pair: Pair, other: i32 }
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
@codex review

Issue #1885, opened 2026-09-03
FA-02-041: CI public-api guard omits sm-front despite ScopeEnv's active public surface


skulmakov-oss:
Umbrella: #1617
Module: **02 — `sm-front`**
Classification: **FALSE-READY · QUALIFICATION-GAP**
Severity: **P2 — public API can drift for this crate without the repository's named public-API guard noticing**
Audit baseline: `528ba2a4b4973c6495f56627047aefab8a0bc27a`
Discovered while implementing: #1664 completion slice (`ScopeEnv::is_const` fail-closed migration)

## Finding

The CI job named `public-api-guard` runs `tests/public_api_contracts.rs`, but that test's `TARGETS` list does not include `crates/sm-front/src/lib.rs` or any `sm-front` public-surface snapshot, despite a snapshot file already existing on disk at `tests/golden_snapshots/public_api/sm_front_lib.txt`.

Confirmed by direct inspection: `grep -n "sm-front\|sm_front" tests/public_api_contracts.rs` returns zero matches. `tests/golden_snapshots/public_api/sm_front_lib.txt` is referenced nowhere in the Rust source (`grep -rn "sm_front_lib.txt" --include="*.rs" .` returns zero matches). It is a fully orphaned file: present, plausible-looking, and never read or validated by any test.

Directly observed consequence: while migrating `ScopeEnv::is_const` from `pub fn is_const(&self, name: SymbolId) -> bool` to `pub fn is_const(&self, name: SymbolId) -> Result<bool, FrontendError>` (a real, intentional public-API signature change) on this crate, `cargo test --test public_api_contracts` — including `public_api_inventory_matches_checked_in_contract_snapshots` specifically — passed without any complaint, and the checked-in snapshot was never flagged as stale. A reviewer trusting this test's name and its presence in CI would have no signal that `sm-front`'s public surface changed at all.

[...]

Issue #1700, opened 2026-08-15
FA-03-031: CI public-api guard omits sm-sema despite its broad public semantic surface


skulmakov-oss:
Umbrella: #1617
Module: 03 — sm-sema
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **FALSE-READY · QUALIFICATION-GAP**

The CI job named `public-api-guard` runs `tests/public_api_contracts.rs`, but that test's `TARGETS` list does not include `crates/sm-sema/src/lib.rs` or any sm-sema public-surface snapshot.

`sm-sema::lib.rs` publicly exposes semantic types, symbol/type tables, import/export policy helpers, provider APIs and std analysis entrypoints. These can drift without the repository's named public-API guard noticing.

Claim source: CI presents the job as a workspace public API inventory guard.

Evidence:
- `.github/workflows/ci.yml` public-api-guard
- `tests/public_api_contracts.rs::TARGETS`
- `crates/sm-sema/src/lib.rs`

Phase A evidence only. No repair during Phase A.

skulmakov-oss:
Resolved via PR #1946 (merge commit 5a20d1c7b43f9f3a39174fba122f6c8a7fa4c869). Full post-merge qualification suite and CI checks verified on main.

Issue #1579, opened 2026-08-09
SSF-08 — Ownership and memory model positioning


skulmakov-oss:
Parent: #1569
Depends on: #1578

Priority: P8.

Status gate: **BLOCKED until SSF-07 closes.** This phase must align public claims with actual verifier/runtime semantics; it must not imply Rust-equivalent ownership without evidence.

## Goal
Freeze the Stable Foundation ownership/memory promise and qualify it across value paths, frames, aggregates, host boundaries, and resource limits.

## Public positioning decision
Choose one:
- **Position A вЂ” bounded deterministic VM language:** ownership protects admitted value paths/runtime invariants; no full systems-language lifetime/region claim.
- **Position B вЂ” expanded systems ownership model:** extend beyond current ownership baseline only where Stable Foundation requires it and qualification evidence supports it.

## Safe execution steps
1. Read SSF-00вЂ¦07 outputs and inventory current ownership/move/borrow/path-state behavior from specs, verifier rules, VM behavior, tests, and public claims.
2. Write a decision record comparing Position A/B against current implementation, risk, required features, verifier complexity, and external expectations.
3. Select public positioning before changing ownership machinery.
4. Freeze copy/move/read/borrow/release behavior for scalar values and direct paths.
5. Define tuple/record/ADT/Sequence/Map path behavior, including sibling access, partial moves, nested paths, and invalid reuse.
6. Define frame/call boundary behavior and host-boundary value ownership.
7. Define memory/resource quotas and deterministic failure taxonomy; distinguish verifier rejection from runtime quota/trap outcomes.
8. Implement only gaps required by the selected positioning and Stable Foundation contour, in small verifier/runtime slices.
9. Add positive and adversarial negative fixtures: move-after-move, conflicting borrow/path use, partial aggregate move, sibling access, frame transfer, host transfer, quota exhaustion.
10. Verify deterministic agreement between verifier and runtime for all admitted/rejected cases.
11. Remove or correct public wording that exceeds the selected positioning; preserve historical documents as historical.
12. Run ownership/path-state, verifier malformed-artifact, runtime determinism/quota, and relevant workspace regression gates.

## Acceptance criteria
- [x] Public ownership/memory claim matches implementation evidence.
- [x] Aggregate/path/frame/host ownership behavior is documented and tested.
- [x] Partial move and sibling-access rules are deterministic.
- [x] Resource quotas and failure taxonomy are explicit.
[...]

PR #1886, opened 2026-09-03
fix(front,ir): make ScopeEnv const lookup fail closed

### LINKED ISSUE #1664: FA-02-032: ScopeEnv ownership APIs fail open when the requested binding is missing

skulmakov-oss:
SSF-08 #1664 completion slice.

This PR deliberately opens the `sm-front → sm-ir` boundary only to migrate the three existing `legacy_lowering` consumers of `ScopeEnv::is_const` after the final fail-open bool-returning API is replaced by the canonical `Result`-returning contract.

No Lane 2 ownership/lowering semantics are changed.

## Baseline / HEAD

- Expected baseline: `528ba2a4b4973c6495f56627047aefab8a0bc27a` (confirmed exact match: `git fetch origin && git rev-parse origin/main`, working tree clean before branching).
- This branch: `fix/1664-is-const-fail-closed`.

## Public API change (intentional, pre-stable contract correction)

```
ScopeEnv::is_const
bool
→ Result<bool, FrontendError>
```

This is not a cosmetic refactor. The `bool` contract was semantically unsound because a missing binding collapsed into a legitimate non-const result — the exact defect #1664 (FA-02-032) exists to eliminate, and the last of its seven named APIs to still have it.

## Confirmed residual architecture, then workspace-wide call-site audit

Confirmed on exact baseline before editing: `is_const`'s bool-returning shape and the `is_const_checked` split matched the expected pre-slice shape exactly (section 5 of the governing spec).

`rg -n "\.is_const\(" .` / `rg -n "is_const_checked\(" .` across the entire workspace found:

| Consumer | Kind |
|---|---|
| `crates/sm-front/src/typecheck.rs:1651` (`Stmt::Assign`) | internal, via `is_const_checked` |
| `crates/sm-front/src/typecheck.rs:1719` (`Stmt::AssignTuple`) | internal, via `is_const_checked` |
| `crates/sm-front/src/typecheck.rs:16507` (`ensure_const_initializer_safe`) | internal, via `is_const_checked` |
| `crates/sm-ir/src/legacy_lowering.rs:2467` (closure capture) | external, via `is_const` |
| `crates/sm-ir/src/legacy_lowering.rs:5011` (tuple-assignment target) | external, via `is_const` |
| `crates/sm-ir/src/legacy_lowering.rs:6107` (ordinary-assignment target) | external, via `is_const` |

Exactly the three logical `sm-ir` uses the governing spec expected, confirmed by search rather than assumed. Zero consumers found outside `sm-front`/`sm-ir`.

## Call-site ordering audit (no race between lookup and const query)

All three `sm-ir` call sites, and all three `sm-front` internal call sites, follow the identical, pre-existing pattern:

```rust
let target_ty = env.get(*name).ok_or(FrontendError { ...canonical diagnostic... })?;
if env.is_const(*name)? { ...reject... }
```

Existence is resolved first (producing the canonical source-facing diagnostic on absence); constness is queried only once that has already succeeded. No scope pop, rebinding, or environment replacement occurs between the two calls at any of the six sites — confirmed by direct inspection of each enclosing function, not assumed. No race exists; the migration is a pure `?`-propagation change at each site.

## Repair

**`crates/sm-front/src/lib.rs`**:
```rust
pub fn is_const(&self, name: SymbolId) -> Result<bool, crate::types::FrontendError> {
    Ok(self.require_binding(name)?.is_const)
}
```
via the existing canonical `require_binding` authority — the same shape as every other #1664 API. `is_const_checked` deleted entirely; zero production or test callers remain (one historical-documentation comment reference only, per the governing spec's own allowance).

**`crates/sm-front/src/typecheck.rs`**: the three internal call sites migrated from `is_const_checked(...)?` to the now-canonical `is_const(...)?`. Test module: the missing-binding regression renamed from `ssf08_1664_is_const_checked_fails_closed_on_missing_binding` to `ssf08_1664_is_const_fails_closed_on_missing_binding` (same assertion, canonical API); the now-obsolete `ssf08_1664_is_const_legacy_bool_api_remains_fail_open_for_lane2_compat` test **removed** (it documented behavior that no longer exists); three new direct unit tests added for known const/non-const/mutable bindings (see regression matrix below).

**`crates/sm-ir/src/legacy_lowering.rs`**: exactly three one-character diffs (`is_const(*x)` → `is_const(*x)?`) at the three call sites above. All three enclosing functions (`lower_closure_literal_expr`, `assign_tuple_items`, `lower_stmt`) already return `Result<_, FrontendError>`-compatible types — no signature widening needed anywhere. Two new tests added (see below). No other line in this file changed.

## All seven #1664 APIs — final audit

| API | Old missing-binding behavior | Final contract | Direct regression |
|---|---|---|---|
| `mark_consumed` | no-op | `Result<(), FrontendError>` / fail closed | `ssf08_1664_mark_consumed_fails_closed_on_missing_binding` |
| `is_consumed` | implicit `false` | `Result<bool, FrontendError>` / fail closed | `ssf08_1664_is_consumed_fails_closed_on_missing_binding` |
| `mark_path_state` | no-op | `Result<(), FrontendError>` / fail closed | `ssf08_1664_mark_path_state_fails_closed_on_missing_binding` |
| `check_path_available` | `Ok(())` | `Result<(), FrontendError>` / fail closed | `ssf08_1664_check_path_available_fails_closed_on_missing_binding` |
| `check_capture_allowed` | `Ok(())` | `Result<(), FrontendError>` / fail closed | `ssf08_1664_check_capture_allowed_fails_closed_on_missing_binding` |
| `is_const` | `false` | `Result<bool, FrontendError>` / fail closed | `ssf08_1664_is_const_fails_closed_on_missing_binding` (**this PR**) |
| `is_mutable` | `false` | `Result<bool, FrontendError>` / fail closed | `ssf08_1664_is_mutable_fails_closed_on_missing_binding` |

All seven satisfy the issue contract. `ScopeEnv::get` remains the deliberately-untouched optional-existence lookup (`Option<Type>`) — the architecture's intentional second authority, not a residual.

## `sm-front` regression matrix (new/renamed, `crates/sm-front/src/typecheck.rs`)

| Case | Test | Result |
|---|---|---|
| Missing binding | `ssf08_1664_is_const_fails_closed_on_missing_binding` | `Err(internal ownership state: required binding ... is missing)` |
| Known const binding | `ssf08_1664_is_const_known_const_binding_is_ok_true` | `Ok(true)` |
| Known ordinary binding | `ssf08_1664_is_const_known_ordinary_binding_is_ok_false` | `Ok(false)` |
| Known mutable binding | `ssf08_1664_is_const_known_mutable_binding_is_ok_false` | `Ok(false)` |
| Source-level: assignment to unknown target | `ssf08_1664_unknown_variable_diagnostic_remains_canonical` (pre-existing, still green) | canonical diagnostic, not internal error |
| Source-level: const initializer referencing unknown variable | `ssf08_1664_const_initializer_unknown_variable_diagnostic_remains_canonical` (pre-existing, still green) | canonical diagnostic, not internal error |

## `sm-ir` regression matrix (new, `crates/sm-ir/src/legacy_lowering.rs`)

| Family | Case | Test | Result |
|---|---|---|---|
| Ordinary assignment | known const target | `lowering_rejects_assignment_to_const_binding` (pre-existing, still green) | canonical `"cannot assign to const binding 'total'"` |
| Ordinary assignment | known non-const target | covered by the large existing assignment-lowering test corpus (unmodified, still green) | existing lowering behavior |
| Tuple assignment | known const target | `lowering_rejects_tuple_assignment_to_const_binding` (**new**) | canonical `"cannot assign to const binding 'count' in tuple destructuring assignment"` |
| Tuple assignment | known non-const target | `lower_tuple_destructuring_assignment_to_tuple_get_ir`, `lower_tuple_assignment_records_write_path_events` (pre-existing, still green) | existing lowering behavior |
| Closure capture | known const capture | `lowering_succeeds_for_closure_capturing_const_binding` (**new**; see note below) | lowers successfully, lifted helper emitted |
| Closure capture | known non-const capture | `first_class_closures_lower_to_runtime_carrier_and_semcod19` (pre-existing, still green) | existing lowering behavior |

### On the "missing target" diagnostics specifically

Investigated whether `"unknown captured value ..."` / `"unknown tuple assignment target ..."` / `"unknown assignment target ..."` are reachable at the `sm-ir` layer via `compile_program_to_ir`. Confirmed empirically (temporary instrumentation, since reverted) that `compile_program_to_ir_with_options_and_profile` calls `type_check_program(&program)?` **before** lowering — so any genuinely-unknown identifier is already rejected by `sm-front`'s own gate (which has its own, separately-checked equivalent diagnostics) before `sm-ir`'s own copies of these checks are ever reached. These three `sm-ir`-level messages are therefore defensive, internal-consistency assertions (guarding against `sm-ir`'s independently-constructed lowering-time `ScopeEnv` ever drifting from what `sm-front`'s typecheck-time `ScopeEnv` already proved), not end-user-reachable diagnostics — consistent with the governing spec's explicit instruction not to manufacture artificial states to force coverage of genuinely production-unreachable paths. Not pursued further; this does not affect the migration's correctness, since it only changes how a missing-identity result propagates *after* this (already-proven-safe) resolve step, not whether it can occur.

### On the closure-capture regression's actual scope

`lowering_succeeds_for_closure_capturing_const_binding` proves the closure-capture call site's `is_const(*capture)?` migration does not regress the ordinary case: a closure capturing a const binding still lowers successfully. It does **not** independently prove that `lifted_env.insert_const(...)` (rather than `lifted_env.insert(...)`) actually fired inside the lifted helper's own environment.

Investigated whether that could be observed directly. The only operation whose outcome differs between the two is an assignment to the captured name rejecting as const — and a closure body is a value-producing block (`infer_value_block_type` in `crates/sm-front/src/typecheck.rs`), which admits only `Stmt::Const | Let | LetTuple | Discard | Expr(_)`. `Stmt::Assign` is categorically not admitted inside any value-producing block, closure bodies included — confirmed both by reading `infer_value_block_type`'s match arms directly and empirically: `(x => { offset = 2.0; x })` fails to parse with `"expected '}' after value-producing block"`, independent of `offset`'s constness. No admitted source can place a reassignment inside a closure body at all, so no admitted source can make the lowering outcome differ based on which `lifted_env` call fired. The only way to observe it directly would be exposing `lifted_env` from `lower_closure_literal_expr` for test inspection — a production hook added solely for testing, which the governing spec explicitly forbids. The test and its own doc comment state this limitation explicitly; the PR does not claim more than what it proves.

## Public API snapshot — deliberately untouched

`tests/golden_snapshots/public_api/sm_front_lib.txt` is **not** modified by this PR. An earlier draft of this PR regenerated it to reflect the `is_const` signature change; that regeneration has been removed.

**Why:** investigating the expected snapshot delta (`cargo test --test public_api_contracts` passed all 88 tests with the snapshot *unmodified* even after the `is_const` signature change landed in source) revealed `tests/public_api_contracts.rs::TARGETS` does not include `crates/sm-front/src/lib.rs` at all — the file is completely orphaned, never read or validated by any test. Regenerating it locally (via the test's own `update_mode()` write path, `sm-front` temporarily wired into `TARGETS` for that one regeneration only, then reverted) showed the checked-in snapshot predates not just this PR's `is_const` change but the *already-merged* SSF-08 Lane 1 corrective round (PR #1880) entirely, plus several unrelated modules/types added since.

Filed as **#1885 (FA-02-041)**, the same class of finding as the already-tracked #1700 (FA-03-031, `sm-sema` omitted from the same guard). Per that finding: this snapshot is not currently owned by `public-api-guard`, and its regeneration would incorporate substantial historical drift unrelated to #1664 — #1886 must not establish that accumulated drift as a new baseline as a side effect of unrelated work. The orphaned file, `TARGETS` wiring, canonical regeneration, and review of the accumulated drift all belong to #1885's own, separately-scoped repair. `tests/public_api_contracts.rs` has zero diff in this PR.

## Determinism

`is_const`'s fail-closed path routes through the same `require_binding` authority every other #1664 API already uses — no `BTreeMap`-iteration-order sensitivity introduced; `ScopeEnv`'s scope storage is unchanged.

## Fallback audit

`rg -n "unwrap_or\(false\)|unwrap_or_default|map_or\(false|is_const_checked|is_const_legacy" crates/sm-front crates/sm-ir` — all matches are pre-existing, unrelated parser/lexer code (`TokenKind` lookahead, Logos-profile auto-detection) and one historical-documentation comment referencing the now-retired `is_const_checked` split; none touch the const-state path. `rg -n "\.is_const\(" crates` — all 10 occurrences (6 production, 4 test) handle the `Result` explicitly; zero `.unwrap_or(false)`, zero `.ok().unwrap_or(...)`, zero ignored `Result`.

## Explicitly out of scope

- `#1579` — not closed. #1664 completion means the Lane 1 API residual is closed, not that SSF-08 is complete.
- `#1709`, `#1718`, `#1724`, `#1725`, `#1726` — not touched. Diff-audited `legacy_lowering.rs`'s full changeset (above) and confirmed zero changes to nested-expression lowering, value-block lowering, ownership-event collection, `SymbolId`/root-identity handling, or loop lowering — exactly the three `?` insertions plus two new tests, nothing else.
- No new lifetimes, regions, borrow graphs, escape analysis, or move semantics. Plain root rebinding stays copy-by-value (unaffected — `scalar_root_rebind_copy_semantics_e2e.rs` still 6/6).
- `#1885` (FA-02-041, the orphaned public-api-guard gap found during this slice) — filed, not repaired here. `tests/golden_snapshots/public_api/sm_front_lib.txt` is deliberately left byte-for-byte identical to `origin/main`; this PR does not update it.

## Local qualification

| Command | Result |
|---|---|
| `cargo test -p sm-front --lib` | **646 passed**, 0 failed |
| `cargo check -p sm-ir --all-targets` | clean |
| `cargo test -p sm-ir` | **138 passed**, 0 failed |
| `cargo clippy -p sm-ir --all-targets -- -D warnings` | clean |
| `cargo test --workspace` | **4442 passed, 0 failed** — full pass, not INCOMPLETE (confirmed identical count after the corrective round, as expected: one test renamed/reframed, none added or removed) |

Do not merge. Draft PR pending review, matching the exact-head discipline established on #1879/#1880/#1882/#1884.

Closes #1664

### DIFF
--- a/crates/sm-front/src/lib.rs
+++ b/crates/sm-front/src/lib.rs
@@ -521,29 +521,25 @@ impl ScopeEnv {
         self.binding(name).map(|binding| binding.ty.clone())
     }
 
-    /// SSF-08 Lane 1 (#1664): kept fail-open (`false` on a missing binding)
-    /// -- unlike every other #1664 API, this exact bool-returning shape has
-    /// live production call sites in `crates/sm-ir/src/legacy_lowering.rs`,
-    /// a Lane 2 crate this PR is not permitted to modify (SSF-08 Lane 1's
-    /// scope is `crates/sm-front` only). Changing this signature would
-    /// break that crate's build. Every in-crate (`sm-front`) production
-    /// call site now uses the fail-closed `is_const_checked` below instead;
-    /// this method is retained solely for that external, out-of-scope
-    /// dependency and must gain no new call sites within `sm-front`.
-    pub fn is_const(&self, name: SymbolId) -> bool {
-        self.binding(name)
-            .map(|binding| binding.is_const)
-            .unwrap_or(false)
-    }
-
-    /// SSF-08 Lane 1 (#1664): fail-closed counterpart of `is_const`, used by
-    /// every `sm-front`-internal call site. See `is_const`'s doc comment
-    /// for why that original bool-returning method could not simply be
-    /// changed in place.
-    pub(crate) fn is_const_checked(
-        &self,
-        name: SymbolId,
-    ) -> Result<bool, crate::types::FrontendError> {
+    /// SSF-08 (#1664 completion): REQUIRED_BINDING -- fails closed if
+    /// `name` is not a known binding, rather than treating "missing" as
+    /// "not const". This is a semantic property query against a binding
+    /// the caller has already established must exist -- every current
+    /// call site (`crates/sm-front/src/typecheck.rs`'s assignment/const-
+    /// initializer checks, and `crates/sm-ir/src/legacy_lowering.rs`'s
+    /// closure-capture/tuple-assignment/ordinary-assignment lowering)
+    /// resolves the binding via `env.get(name)` -- reporting the
+    /// canonical source-level "unknown ..." diagnostic on absence -- and
+    /// only queries constness once that has already succeeded, so this
+    /// `Err` path represents an internal invariant failure (the identity
+    /// disappeared after being proven to exist), never an ordinary source
+    /// mistake. This was the last of #1664's seven APIs to be converted;
+    /// it originally stayed fail-open because `crates/sm-ir` (Lane 2) had
+    /// live consumers of the old bool-returning shape and Lane 1's scope
+    /// was frontend-only -- SSF-08's #1664 completion slice explicitly
+    /// authorized migrating those three call sites so this contract
+    /// correction could land.
+    pub fn is_const(&self, name: SymbolId) -> Result<bool, crate::types::FrontendError> {
         Ok(self.require_binding(name)?.is_const)
     }
 
--- a/crates/sm-front/src/typecheck.rs
+++ b/crates/sm-front/src/typecheck.rs
@@ -1648,7 +1648,7 @@ fn check_stmt(
                     resolve_symbol_name(arena, *name)?
                 ),
             })?;
-            if env.is_const_checked(*name)? {
+            if env.is_const(*name)? {
                 return Err(FrontendError {
                     pos: 0,
                     message: format!(
@@ -1716,7 +1716,7 @@ fn check_stmt(
                         resolve_symbol_name(arena, *name)?
                     ),
                 })?;
-                if env.is_const_checked(*name)? {
+                if env.is_const(*name)? {
                     return Err(FrontendError {
                         pos: 0,
                         message: format!(
@@ -12168,16 +12168,16 @@ mod tests {
     }
 
     #[test]
-    fn ssf08_1664_is_const_checked_fails_closed_on_missing_binding() {
-        // is_const_checked is the fail-closed sibling used by every
-        // sm-front-internal call site (see is_const's own doc comment and
-        // the next test for why the original is_const symbol itself could
-        // not be changed in place).
+    fn ssf08_1664_is_const_fails_closed_on_missing_binding() {
+        // #1664 completion: is_const is now the single, canonical
+        // fail-closed authority (the is_const_checked/is_const split has
+        // been retired -- see is_const's own doc comment for why the old
+        // bool-returning shape could finally be changed in place).
         let env = ScopeEnv::new();
         let unknown = SymbolId(993);
         let err = env
-            .is_const_checked(unknown)
-            .expect_err("is_const_checked on a genuinely missing binding must fail closed");
+            .is_const(unknown)
+            .expect_err("is_const on a genuinely missing binding must fail closed");
         assert!(
             err.message.contains("internal ownership state"),
             "unexpected: {}",
@@ -12186,22 +12186,38 @@ mod tests {
     }
 
     #[test]
-    fn ssf08_1664_is_const_legacy_bool_api_remains_fail_open_for_lane2_compat() {
-        // Deliberate, evidenced exception: `is_const`'s bool-returning shape
-        // has live production call sites in crates/sm-ir/src/
-        // legacy_lowering.rs (Lane 2), which SSF-08 Lane 1 is not permitted
-        // to modify. This pins that the old symbol still returns `false` on
-        // a missing binding -- not because it is correct, but because
-        // changing it would require editing a Lane 2 crate. Every
-        // sm-front-internal decision now goes through is_const_checked
-        // instead (see the prior test); this method must gain no new
-        // sm-front call sites.
-        let env = ScopeEnv::new();
-        let unknown = SymbolId(992);
-        assert!(
-            !env.is_const(unknown),
-            "legacy is_const must still fail open (false) on a missing binding -- this is the \
-             documented Lane 2 compatibility exception, not new behavior"
+    fn ssf08_1664_is_const_known_const_binding_is_ok_true() {
+        let mut env = ScopeEnv::new();
+        let sym = SymbolId(991);
+        env.insert_const(sym, Type::I32);
+        assert_eq!(
+            env.is_const(sym),
+            Ok(true),
+            "a known const binding must report Ok(true)"
+        );
+    }
+
+    #[test]
+    fn ssf08_1664_is_const_known_ordinary_binding_is_ok_false() {
+        let mut env = ScopeEnv::new();
+        let sym = SymbolId(990);
+        env.insert(sym, Type::I32);
+        assert_eq!(
+            env.is_const(sym),
+            Ok(false),
+            "a known ordinary (non-const, non-mutable) binding must report Ok(false)"
+        );
+    }
+
+    #[test]
+    fn ssf08_1664_is_const_known_mutable_binding_is_ok_false() {
+        let mut env = ScopeEnv::new();
+        let sym = SymbolId(989);
+        env.insert_mut(sym, Type::I32);
+        assert_eq!(
+            env.is_const(sym),
+            Ok(false),
+            "a known mutable binding must report Ok(false) -- mutable and const are exclusive"
         );
     }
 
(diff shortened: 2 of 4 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1887, opened 2026-09-03
fix(ir): preserve ownership events through nested lowering

### LINKED ISSUE #1709: FA-04-003: nested value and loop-expression lowering can erase frozen ownership events

skulmakov-oss:
## Summary

FA-04-003 / SSF-08 Lane 2a: nested value and loop-expression lowering can erase frozen ownership events.

**Architecture statement:** #1709 establishes one ownership-event sink per `IrFunction`. Nested lowering within the same function reuses that sink. Lifted child functions use a distinct child sink that is retained as the child's `IrFunction::ownership_events`. No event-producing lowering path writes into an unretained temporary sink, and for every currently admitted #1709 nested/value/loop/function-child lowering path, the existing ownership-event producer authorities feed that retained sink exactly once.

This PR shipped in two rounds. The first round threaded the sink; an exact-head review of that round found the sink was correct everywhere but several nested roots still never *called* the existing write-event producer authority, so no event was generated there in the first place regardless of sink correctness. The second round wires those calls in. Both are described below.

## Round 1 — sink threading and the two value-block gaps found alongside it

| Path | Baseline behavior | Fixed behavior | Event owner |
|---|---|---|---|
| value-block `LetTuple` | `bind_tuple_items` received `&mut Vec::new()` **and** a hardcoded `None` `AccessPath` (so no Borrow event was generated at all, regardless of sink) | real `AccessPath` derived via `sequence_access_path_from_expr` (mirroring `lower_stmt`'s twin arm); Borrow event retained exactly once | parent `IrFunction` |
| value-block `LetRecord` | same defect class as `LetTuple` | same fix applied at the IR layer for architectural consistency; **currently unreachable via admitted source** — confirmed by direct inspection that `sm-front::typecheck::infer_value_block_type`'s statement whitelist for any value-producing block is exactly `Const \| Let \| LetTuple \| Discard \| Expr(_)`, `LetRecord` is not admitted there. Mirrors the #1664 closure-capture-constness precedent. | parent `IrFunction` (defensive) |
| value-block record-update write (`Const`/`Let`/`LetTuple`/`LetRecord` arms) | `lower_value_block_expr`'s statement arms never called `append_record_update_write_events_from_expr` at all | called once per statement, mirroring `lower_stmt` | parent `IrFunction` |
| loop-expression fallback | `lower_loop_expr_stmt`'s `_` arm built a temporary `LoweringCtx` with `ownership_events: Vec::new()`, copied back every piece of state except `ownership_events` | `ownership_events: core::mem::take(ownership_events)` in, `*ownership_events = ctx.ownership_events` out — same idiom already used here for `closure_state` | parent `IrFunction` |
| lifted closure | `IrFunction { ownership_events: Vec::new(), .. }` hardcoded regardless of body content | body lowers into a fresh local `lifted_ownership_events`, retained as the pushed `IrFunction`'s own field | lifted child `IrFunction` |

`lower_expr`/`lower_expr_with_expected` and every helper reachable from them now take an explicit `ownership_events: &mut Vec<OwnershipPathEvent>` parameter, threaded the same way `next`/`out`/`loop_stack`/`closure_state` already are.

## Round 2 — producer-root audit and the six wiring gaps it found

The first round's sink threading was correct but incomplete: several admitted nested roots had a correct place to put the event, but never called `append_record_update_write_events_from_expr` at all. Each of the six below now mirrors its canonical top-level `lower_stmt` twin exactly:

| Root | Was prescanned? | Fixed |
|---|---|---|
| `lower_stmt`'s own `Stmt::Break(Some(value))` | No — unlike its `Stmt::Return(Some(value))` sibling | Yes. `break value;` is only reachable nested inside a loop expression, so this is squarely #1709's contour even though the fix lands in the canonical shared function. |
| `lower_value_block_expr`'s `Stmt::Discard` | No — unlike `lower_stmt`'s twin | Yes |
| `lower_value_block_expr`'s `Stmt::Expr` | No — unlike `lower_stmt`'s twin | Yes |
| `lower_loop_expr_stmt`'s dedicated `Stmt::If` (does not delegate to `lower_stmt`) | No — unlike `lower_stmt`'s `Stmt::If` | Yes, on `condition` |
| `lower_loop_expr_stmt`'s dedicated `Stmt::Match` (does not delegate to `lower_stmt`) | No — unlike `lower_stmt`'s `Stmt::Match` | Yes, on `scrutinee` |
| `lower_closure_literal_expr`'s `closure.body` | No — `append_record_update_write_events_from_expr` treats `Expr::Closure` as a deliberate leaf, so no enclosing scan ever reaches a closure body | Yes, into the child's own sink (new function boundary) |

### Non-#1709 residuals found, not fixed here

Per the audit's own classification rule (an inconsistency *within* the top-level canonical authority itself is not the same defect as a nested path losing something the top level already has), two more asymmetries were found and are explicitly **not** part of this PR:

- `lower_for_range_stmt`'s `range` and `lower_for_each_stmt`'s `iterable` are never prescanned, unlike `lower_while_stmt`'s `condition`, which already is. Both are top-level statement-loop headers, not nested/loop-*expression* paths.
- `Stmt::Guard`'s `else_return` payload is never prescanned before calling `lower_return_payload`, unlike its `Stmt::Return` sibling, which is. Guard cannot appear inside a loop expression at all (rejected outright there), so there is no nested/top-level pair to compare — this is a standalone inconsistency between two sibling statement kinds, not #1709-shaped.

Tracked separately as FA-04-025 / #1888. No repair is included in this PR.

## Mechanism

Same as before: small explicit channel threading (not a `LoweringCtx` rewrite). Top-level callers (`lower_stmt`, the function-entry driver, the `for`/`while`/statement-loop family) pass `&mut ctx.ownership_events`. The lifted-closure boundary gets a fresh local `Vec`. `lower_expr_to_ir` (a bare-expression entry point with no owning `IrFunction`) gets a local, discarded `Vec` — not the same defect class, since there is no retained sink for it to preserve into.

## Scope audit

- `#1724` lexical binding identity: unchanged
- `#1725` OWN0 root identity: unchanged
- `#1726` OWN0 event timing/interleaving: unchanged (IR-level event vector order preserved exactly; no `sm-format`/VM changes)
- `#1718` capability/path-family contour: unchanged (regressions use only `TupleIndex`/direct `Field` paths)
- VM overlap semantics, verifier admission, Position A: unchanged
- No new ownership semantics: every fix in both rounds calls an *existing* producer/decision function; none decide what is Borrow/Write differently than before

Nested IR-lowering event preservation is repaired for every currently admitted #1709 path. The ownership pipeline is **not** claimed to be fully correct — `#1724`/`#1725`/`#1726` remain open.

## Regression matrix (`crates/sm-ir/src/legacy_lowering.rs`)

Round 1:
- `ssf08_1709_nested_value_block_tuple_borrow_preserves_ownership_event`
- `ssf08_1709_nested_value_block_record_update_write_event_preserved`
- `ssf08_1709_loop_expr_fallback_preserves_ownership_event`
- `ssf08_1709_lifted_closure_owns_body_ownership_event_not_parent`

Round 2:
- `ssf08_1709_value_block_discard_write_event_preserved`
- `ssf08_1709_value_block_expr_statement_write_event_preserved`
- `ssf08_1709_loop_expr_special_if_condition_write_event_preserved`
- `ssf08_1709_loop_expr_special_match_scrutinee_write_event_preserved`
- `ssf08_1709_loop_expr_break_value_write_event_preserved`
- `ssf08_1709_closure_body_write_event_owned_by_lifted_helper`

Every assertion checks the exact event vector (`assert_eq!` against a literal `vec![...]`), not mere containment, catching both event loss and duplication. **All ten fixes were individually mutation-falsified** (reverted, confirming the corresponding test fails, restored, confirming it passes again) — not committed.

## Qualification (post round 2, HEAD `5f698e9c84964fee04598c87048d7c2561065313`)

- `cargo fmt --all -- --check` — clean
- `cargo check -p sm-ir --all-targets` — clean
- `cargo test -p sm-ir` — 148/148 (138 pre-existing + 10 new), 0 failed
- `cargo test --test runtime_ownership_e2e` — 32/32
- `cargo test --test scalar_root_rebind_copy_semantics_e2e` — 6/6
- `cargo test --test match_surface_qualification` — 3/3
- `cargo test --test ssf_status_drift` — 3/3
- `cargo test --test public_api_contracts` — 88/88 (no public API changed)
- `cargo test --workspace` — 4452/4452 passed, 0 failed, across all 324 test binaries
- `git diff --check` — clean
- Fresh-Vec audit (`rg "ownership_events:\s*Vec::new|&mut Vec::new\(\)"` in production code): exactly one occurrence, `LoweringCtx::new`'s own field initializer
- `LoweringCtx {` construction audit: exactly one literal construction (the loop-expression fallback), plus the `struct`/`impl` declarations

## Changed files

`crates/sm-ir/src/legacy_lowering.rs` only, across both commits.

Closes #1709

### DIFF
--- a/crates/sm-ir/src/legacy_lowering.rs
+++ b/crates/sm-ir/src/legacy_lowering.rs
@@ -680,6 +680,13 @@ pub fn lower_expr_to_ir(
     for (name, ty) in var_types {
         env.insert(*name, ty.clone());
     }
+    // #1709: this entry point has no owning `IrFunction` (it lowers a bare
+    // expression, not a compiled function - see the `lifted_funcs` rejection
+    // below), so there is no retained sink for these events to preserve
+    // into; a local, discarded `Vec` here is not the same defect class as a
+    // same-function nested lowering path silently dropping its parent
+    // function's events.
+    let mut ownership_events = Vec::new();
     let _ = lower_expr(
         expr,
         arena,
@@ -692,6 +699,7 @@ pub fn lower_expr_to_ir(
         &empty_adts,
         Type::Unit,
         &mut closure_state,
+        &mut ownership_events,
     )?;
     if !closure_state.lifted_funcs.is_empty() {
         return Err(FrontendError {
@@ -810,6 +818,7 @@ fn lower_function_to_ir_with_tables(
             adt_table,
             canonical_ret.clone(),
             &mut ctx.closure_state,
+            &mut ctx.ownership_events,
         )?;
         if cond_ty != Type::Bool {
             return Err(FrontendError {
@@ -837,6 +846,7 @@ fn lower_function_to_ir_with_tables(
         adt_table,
         func.ret.clone(),
         &mut ctx.closure_state,
+        &mut ctx.ownership_events,
     )?;
     for stmt in &func.body {
         lower_stmt(
@@ -867,6 +877,7 @@ fn lower_function_to_ir_with_tables(
                 adt_table,
                 func.ret.clone(),
                 &mut ctx.closure_state,
+                &mut ctx.ownership_events,
             )?;
             lower_invariant_clauses(
                 &ctx.invariants,
@@ -883,6 +894,7 @@ fn lower_function_to_ir_with_tables(
                 adt_table,
                 func.ret.clone(),
                 &mut ctx.closure_state,
+                &mut ctx.ownership_events,
             )?;
             ctx.instrs.push(IrInstr::Ret { src: None });
         } else {
@@ -2447,6 +2459,14 @@ fn lower_closure_literal_expr(
         message: "closure parameter/capture count exceeds register space".to_string(),
     })?;
     let mut local_loop_stack = Vec::new();
+    // #1709: the lifted helper is a new `IrFunction`, hence a new
+    // function-owned ownership-event sink (§25/§26) - it must NOT share the
+    // parent's `ownership_events` sink (that would leak the closure body's
+    // own Borrow/Write events into the parent frame's event stream, which
+    // later has to line up with the parent's own instruction/frame
+    // boundaries) and must NOT be discarded as `Vec::new()` at construction
+    // (that was the original #1709 defect for this call site).
+    let mut lifted_ownership_events: Vec<OwnershipPathEvent> = Vec::new();
     // #1773 (FA-09-005): the lifted helper's real invocation convention is
     // captures first (r0..captures.len()-1), then the closure's own param
     // (r{captures.len()}) - built here from the exact same `capture_ty`/
@@ -2502,6 +2522,7 @@ fn lower_closure_literal_expr(
         Some(expected_closure.ret.as_ref().clone()),
         expected_closure.ret.as_ref().clone(),
         closure_state,
+        &mut lifted_ownership_events,
     )?;
     if body_ty != expected_closure.ret.as_ref().clone() {
         return Err(FrontendError {
@@ -2518,7 +2539,7 @@ fn lower_closure_literal_expr(
     closure_state.lifted_funcs.push(IrFunction {
         name: helper_name.clone(),
         instrs: lifted_instrs,
-        ownership_events: Vec::new(),
+        ownership_events: lifted_ownership_events,
         params: lifted_signature_params,
     });
 
@@ -2554,6 +2575,7 @@ fn lower_direct_closure_call_expr(
     adt_table: &AdtTable,
     ret_ty: Type,
     closure_state: &mut ClosureLoweringState,
+    ownership_events: &mut Vec<OwnershipPathEvent>,
 ) -> Result<(u16, Type), FrontendError> {
     if closure_ty.family != ClosureValueFamily::UnaryDirect
         || closure_ty.capture != ClosureCapturePolicy::Immutable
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex new commit review please

Issue #1888, opened 2026-09-03
FA-04-025: top-level loop headers and guard return payload can skip ownership-event discovery


skulmakov-oss:
Severity: P1
Classification: SEMANTIC-DRIFT · FAIL-OPEN · FALSE-READY

Umbrella: #1617
Module: 04 — sm-ir
Discovered during #1709 / PR #1887 producer-root audit.

Finding:

The canonical top-level lowering authority does not invoke
append_record_update_write_events_from_expr for several admitted
expression roots:

- lower_for_range_stmt: range
- lower_for_each_stmt: iterable
- Stmt::Guard: else_return payload before lower_return_payload

Sibling paths already perform equivalent pre-scans:
- lower_while_stmt scans condition
- Stmt::Return scans its return payload

Therefore an admitted RecordUpdate nested in one of these expression
roots may lower successfully while its Write ownership metadata is
never generated.

This is distinct from #1709 / FA-04-003. #1709 repairs preservation
across nested/value/loop-expression/child-function lowering boundaries.
These gaps exist inside the canonical top-level statement authority
itself.

Impact:
source ownership-sensitive write
→ executable IR emitted
→ Write event absent
→ downstream ownership overlap enforcement is weaker than source intent

Evidence:
- crates/sm-ir/src/legacy_lowering.rs::lower_for_range_stmt
- crates/sm-ir/src/legacy_lowering.rs::lower_for_each_stmt
- crates/sm-ir/src/legacy_lowering.rs::lower_stmt Stmt::Guard
- append_record_update_write_events_from_expr
- sibling lower_while_stmt / Stmt::Return behavior

No repair in PR #1887.

skulmakov-oss:
## Reconciliation: superseded by #1891 Checkpoint W2A

**Original defect:** `RecordUpdate` Write-event generation depended on an outer prescan (`append_record_update_write_events_from_expr`) that `lower_for_range_stmt`'s range bound, `lower_for_each_stmt`'s iterable, and `Stmt::Guard`'s `else_return` payload never called — a real, executable `RecordUpdate` reachable only through one of those three roots could lower to a `MakeRecord` with no describing Write ownership event at all.

[...]

PR #1889, opened 2026-09-03
fix(ir): give lowered locals per-binding identity instead of per-spelling (#1724)

### LINKED ISSUE #1724: FA-04-018: lowering flattens lexical scopes into one string-local namespace and can leak shadowed bindings

skulmakov-oss:
## Summary

Fixes #1724 (FA-04-018): lowering flattened lexical scopes into one string-local namespace and could leak shadowed bindings (`let x = 1; if true { let x = 2; } return x;` returned `2` at the VM level, since both bindings lowered to the same runtime-local key).

**Architecture statement:** sm-ir now assigns runtime-local identity per lexical binding rather than per source spelling. Scope push/pop changes which lowered local identity a source name resolves to, while assignments and reads of the same binding reuse one identity. The repair affects ordinary local transport only; OWN0 root identity is unchanged.

## Investigation before implementation

Before any implementation, established empirically (code inspection + a probe test, since removed) whether `SymbolId` is declaration-unique or spelling-interned:

- `AstArena::intern_symbol` interns by a `BTreeMap<String, SymbolId>` keyed on raw spelling — **spelling-interned, not declaration-unique**. Two shadowed declarations with identical spelling share the exact same `SymbolId`.
- Shadowing is distinguished purely by `ScopeEnv`'s scope-stack position (`scopes: Vec<BTreeMap<SymbolId, ScopeBinding>>`, innermost-scope-wins lookup), not by SymbolId identity.
- Confirmed via a temporary probe test: outer decl, inner decl, inner use, and post-scope outer use all returned `SymbolId(1)`.

This ruled out "mangle by SymbolId" (would just relocate the same collision under a new name) and pointed to a scope-aware identity authority as the correct repair.

## Implementation (crates/sm-ir/src/legacy_lowering.rs only)

Added `LoweredLocalEnv`, mirroring `ScopeEnv`'s own scope-stack shape (`scopes: Vec<BTreeMap<SymbolId, String>>`, `next_id: u32`):
- `bind` mints a fresh `__sm_local_<id>_<spelling>` key per binding.
- `resolve` walks scopes innermost-first, fail-closed (`FrontendError`, no fallback to raw spelling) on lookup miss.
- `bind_raw` seeds a binding using the raw spelling as its own key, reserved for pre-existing single-use slots with no shadowing possibility (`lower_expr_to_ir`'s `var_types`, and the ensures/invariants "result" contract slot — the latter was a genuine pre-existing gap this fix also closes).
- `__` is a collision-safe reserved namespace: source identifiers must start with `is_ascii_alphabetic()`.

Threaded through the same recursive lowering call graph #1709 established (~21 function signatures), with scope push/pop paired at every If/Match/loop/value-block boundary, and function-boundary sinks (each `IrFunction` and each lifted closure gets its own fresh authority instance) matching #1709's `mem::take`/restore pattern for the loop-fallback `LoweringCtx` boundary.

## Cross-boundary discovery and resolution (#1725)

`tests/runtime_ownership_e2e.rs` broke (29/32) because it depended on source spelling being identical to the runtime-local key when synthesizing OWN0 test data — the exact same broken assumption already tracked as **#1725** (frontend `SymbolId` written to OWN0, but the VM's `RuntimeSymbolTable` is a separate table with independent numbering; this only "worked" before because simple test programs' raw spelling happened to align).

Per direction, adapted **test-only**: the harness now resolves source spelling → unique lowered runtime-local key → index in `layout.strings`, via a fail-closed `resolve_unique_lowered_local_key` (0 matches → panic, 1 → use it, >1 → panic "ambiguous", no first-match-wins fallback), with a dedicated regression proving it detects shadowing ambiguity. Mutation-checked (reverted to raw lookup → confirmed 29 failures reproduce → restored). **Zero lines changed in `emit_ownership_events`, `sm-format`, `sm-runtime-core`, or `sm-vm`'s production dispatch/borrow-check logic** — production OWN0 root identity is unchanged and remains tracked by #1725.

## Wider golden/snapshot fallout (same representation-drift class)

The same raw-spelling → mangled-key change surfaces as expected, purely-representational drift anywhere a golden/snapshot test pins compiled semcode bytes or a hash of them. For each, confirmed every non-hash field (`source_hash`, `ir_hash`, `verifier_status`, `vm_status`, trap-taxonomy fields, `all_runs_equal`) stayed byte-identical — proving zero semantic change — before regenerating:

- `ctf_e1_golden_traces`, `ctf_e2_collection_replay`, `ctf_e3_trap_taxonomy_regression` — regenerated via their own `SM_UPDATE_*` env vars.
- `golden_semcode` (raw `.smc` binaries, no built-in update mode) — added a temporary update hook, regenerated, reverted the hook (file is back to its original form).
- `golden_snapshots`'s `golden_emit_hash_stable` — regenerated via `SM_UPDATE_SNAPSHOTS=1`; the other 3 snapshots in that file (Logos lexer/parser/IR) are untouched, confirming the scope is isolated to sm-ir's lowering.

`sm-vm`'s own unit tests (`semcode_vm::tests`) hand-construct synthetic OWN0 bytes for borrow-conflict fixtures using the identical raw-spelling lookup pattern (including one case piggybacking plain local declarations, `"camera"`/`"quality"`, as stand-ins for record field names). Added the same fail-closed resolver there — **test-only, inside `#[cfg(test)] mod tests`** — and updated one test's 4 hardcoded golden observation strings. Zero lines of production VM dispatch or borrow-check logic changed. (One pre-existing, unrelated `dead_code` clippy finding on `VmProgramView.header` — dates to commit `6fc777e81`, long before this change — reproduces identically with or without this PR's diff; out of scope.)

## Corrective round (exact-head review on `9f694631`)

Independent exact-head review found one remaining scope-parity gap: `lower_loop_expr` (the **value-position** `Expr::Loop` path, e.g. `let y = loop { ... break x; };`) pushed a fresh `ScopeEnv` scope for its body but never pushed a matching `LoweredLocalEnv` scope — unlike every sibling construct in the same function family (the `Stmt::If`/`Stmt::Match` arms inside `lower_loop_expr_stmt`, a few lines below, already pair both). A `let` directly in a loop-expression body therefore bound into the *enclosing* scope frame instead of a fresh one, and since `SymbolId` is spelling-interned, that silently overwrote a same-spelling outer binding's lowered-local key in place — the outer binding kept resolving to the loop's value after the loop expression exited.

Fixed by pushing/popping `lowered_locals`' own scope alongside `body_env`'s, identical to the pattern already used everywhere else. A full scope-parity audit (`rg -n "push_scope\(\)|pop_scope\(\)" crates/sm-ir/src/legacy_lowering.rs`) confirms this was the *only* gap: `ScopeEnv` push/pop and `LoweredLocalEnv` push/pop each now occur exactly 23 times across the file.

New exact HEAD: `f0f5c870008d25589c2cb1b8d81a07eb8fbbebd6`.

## Regression coverage

`tests/lexical_binding_identity_e2e.rs` (new) — full source-to-VM programs, each proving `source lexical semantics == VM observable result`:
- if-branch shadow, both-branches shadow, match-arm shadow, statement-loop shadow, while-loop shadow
- for-loop variable non-collision with an outer binding of the same spelling
- nested value-block shadow composing with #1709's ownership-event threading
- same-binding reassignment (negative control — proves it reuses the existing key, not a fresh one)
- plain no-shadow baseline
- **(corrective round)** `loop_expression_body_shadow_restores_outer_binding_after_loop_exits` — forces the value-position `Expr::Loop` path specifically, proving the inner `x` is visible inside the loop, `break x` reads the inner binding, and the outer binding is restored after the loop expression exits

Plus 3 IR-structural regressions in `legacy_lowering.rs`'s own test module (the corrective round added one, proving outer/inner loop-expression key distinctness and post-loop key restoration), using relational (not exact-string) key-identity assertions.

Mutation-falsified: reverting the production `Stmt::Let` bind causes all 10 e2e tests to fail closed with `FrontendError`s; reverting the corrective round's `lowered_locals.push_scope()`/`pop_scope()` pair individually reproduces the loop-expression regression's failure in both the VM-behavioral and IR-structural test, confirmed, then restored.

## Boundaries confirmed untouched

- **#1725** (OWN0 root identity / `AccessPath.root` / SymbolId wire meaning) — open, untouched in production.
- **#1726** (OWN0 event timing) — open, untouched.
- **#1718** (capability/path-family) — open, untouched.
- **#1888** (FA-04-025 residual) — open, untouched.

## Test plan

- [x] `cargo test -p sm-ir --lib` — 151/151 (was 150/150; +1 corrective-round IR-structural regression)
- [x] `cargo test --test runtime_ownership_e2e` — 34/34 (32 original + 2 new)
- [x] `cargo test --test lexical_binding_identity_e2e` — 10/10 (was 9/9; +1 corrective-round loop-expression regression)
- [x] `cargo test -p sm-vm --lib` — 118/118 (unaffected by the corrective round — sm-vm not touched this round)
- [x] `cargo test --workspace` — clean, 0 failed (re-run on corrective HEAD)
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo check -p sm-ir --all-targets` / `cargo check --workspace --all-targets` — clean
- [x] `cargo clippy -p sm-ir --all-targets -- -D warnings` — clean
- [x] `cargo clippy --test runtime_ownership_e2e --test lexical_binding_identity_e2e -- -D warnings` — clean
- [x] `git diff --check` — clean

Closes #1724


skulmakov-oss:
@codex review

PR #1890, opened 2026-09-04
fix(ir,vm): resolve OWN0 ownership-path roots through the function's own string table (#1725)

### LINKED ISSUE #1725: FA-04-019: OWN0 roots serialize frontend SymbolId but VM interprets them as function string-table indexes

skulmakov-oss:
## Summary

Repairs #1725 (FA-04-019): OWN0 roots serialized the frontend's own `SymbolId` while the VM's decode/remap already treated the wire `root` field as an index into the function's own string table — two identity domains that were never the same table, aligned only by numeric coincidence in simple pre-#1724 programs.

**Baseline:** `main @ 8387c46d17cafe5a283e7b72da55dcb00199c658` (verified before any work started).

## Investigation (confirmed by direct code inspection)

1. **`AccessPath.root`'s origin/domain** (`crates/sm-ir/src/legacy_lowering.rs`): built from `Expr::Var(name)` — the frontend's own spelling-interned `SymbolId` (`AstArena::intern_symbol`), an identity domain with zero relationship to any runtime table.
2. **Where OWN0 roots are serialized**: `emit_ownership_events` wrote `event.path.root.0` (the raw `SymbolId` number) directly into the wire `root: u32` field.
3. **How the function string table is built**: `emit_semcode_function` builds a `StringInterner` by walking every `IrInstr` and interning `LoadText`/`LoadVar`/`StoreVar`/`Call`/etc. string operands in first-use order — `emit_ownership_events` is called immediately after `interner.emit_table(&mut code)`, but never consulted `interner` at all.
4. **How the VM resolves OWN0 roots today**: `build_vm_program_view_from_decoded`'s `remap_paths` already does `let local_root = path.root_symbol_id as usize; symbol_ids.get(local_root)...` — i.e. it already treats the wire number as an index into this same function-local string table (`symbol_ids`, built in identical order from the same `strings`). The consumer's design already assumed the correct identity domain; only the producer wrote the wrong number. This same remap also carried a forbidden fail-open fallback — `.unwrap_or(SymbolId(path.root_symbol_id))` — for when that index was out of bounds: not merely wrong data, an explicit guess.
5. **Field/variant components**: `PathComponent::Field(SymbolId)` / `AdtPayload.variant: SymbolId` use the same raw-`SymbolId` representation as root did — but **verified, not assumed, to be a materially different situation, not the same bug**. See "Field/variant identity — verified, not assumed" below.
6. **Existing canonical identity authority**: yes — the function's own `StringInterner` (`interner.lookup`, already used identically for e.g. `LoadText`) is exactly the right authority; no second/duplicate symbol table was introduced.
7. **Repair location**: a minimal combination of sm-ir emission (resolve-at-construction via `LoweredLocalEnv`, resolve-at-emission via `StringInterner`) and VM decode/remap (fail-closed instead of guessing). No `sm-format` or `sm-verify` changes were needed — the wire format's `root: u32` slot is unchanged in shape, and neither of those crates references `root_symbol_id`'s value today.

## Before/after identity model (root only)

```
frontend lexical binding identity (SymbolId)
    -> lowered runtime-local key (LoweredLocalEnv::resolve, same authority StoreVar/LoadVar use)
    -> function-local artifact identity (StringInterner index, same table LoadVar/StoreVar operands share)
    -> OWN0 serialized root (u32 index into that table)
    -> VM resolves that exact artifact-local identity (remap_paths, now fail-closed on a miss)
```

`AccessPath.root`'s type changes from `SymbolId` to `String` (the resolved lowered-local key). Every real-lowering construction site resolves through `LoweredLocalEnv` at the point of construction, capturing the identity of whichever binding is lexically current — correct under shadowing, exactly mirroring how StoreVar/LoadVar already resolve.

## Field/variant identity — verified, not assumed (review follow-up)

An earlier version of this PR body claimed field/variant `SymbolId` components were "the identical broken assumption" as root and left deliberately deferred. Independent review correctly flagged that claim as unproven — asserted by extension from the root fix, not demonstrated. Investigated properly:

- Exhaustively grepped every consumer of `PathComponent::Field`/`AdtPayload::variant` across `sm-vm`, `sm-runtime-core`, and `sm-format`: **zero** call sites ever pass these values to `.resolve()`, use them as a table index, or display them. Their one and only consumer is `access_paths_overlap`:
  ```rust
  fn access_paths_overlap(lhs: &AccessPath, rhs: &AccessPath) -> bool {
      if lhs.root != rhs.root { return false; }
      let shared_len = lhs.components.len().min(rhs.components.len());
      lhs.components[..shared_len] == rhs.components[..shared_len]
  }
  ```
  root-gated first, then pure derived-`PartialEq` structural comparison — never resolved through `RuntimeSymbolTable`/`vm.symbols`.
- This makes the raw `SymbolId` value sufficient as an **opaque, deterministic equality key** (spelling-interned, so the same field/variant spelling always produces the same number within one compilation): a spelling collision with an *unrelated* root (e.g. two distinct record types both declaring a `value` field) can never reach the comparison at all, because the root-gate already filters to same-root-only comparisons; and within one root's own record/ADT type, every field or variant spelling is unique by construction (duplicate field names don't type-check). Root identity's bug was different in kind: the raw number *was* resolved, through the wrong table.
- Documented this explicitly on `PathComponent` in both crates that declare it (`crates/sm-runtime-core/src/lib.rs` is canonical; `crates/sm-ir/src/legacy_lowering.rs`'s copy points to it), so it isn't left implicit again.
- Added a proof pair to `tests/own0_root_identity_e2e.rs` (see Tests below), mutation-falsified.

No SemCode/artifact-format redesign was introduced for this — the finding closes the open question about field/variant identity with evidence, it doesn't require a fix, because there is no proven bug there.

## Corrective finding during requalification (#1709-adjacent, fixed within #1725's scope)

#1709's prescan producer (`append_record_update_write_events_from_expr`) walks the AST strictly ahead of and separately from real lowering. It can recurse into a match arm's own body while that body references a variable the same arm's own pattern introduces (`Option::Some(dir) => { match dir {...} }`) — a binding `LoweredLocalEnv` genuinely doesn't have yet, since real binding only happens later. Before #1725 this was silently wrong-but-usually-unconsumed (raw `SymbolId`, no resolution attempted); the new fail-closed resolve turned it into a hard compile failure on a real qualification fixture (`examples/qualification/match_surface/positive_nested_match`).

Fixed by gating the one prescan-internal resolve attempt (the match-scrutinee ADT-payload-borrow path) on whether any arm's pattern actually needs it — matching exactly when the computed path would ever be consumed. The common case (no such capture) never attempts a resolve that would otherwise fail closed for no reason; a genuine need still resolves correctly or fails closed with an honest diagnostic. No event is ever silently dropped.

## Production files changed

- `crates/sm-ir/src/legacy_lowering.rs` — `AccessPath.root` type change, `LoweredLocalEnv`-based resolution at construction sites, `StringInterner`-based resolution at emission, the prescan gating fix, `PathComponent` doc comment.
- `crates/sm-vm/src/semcode_vm.rs` — `remap_paths` fails closed (`RuntimeError::BadFormat`) instead of the removed `.unwrap_or` guess.
- `crates/sm-runtime-core/src/lib.rs` — `PathComponent` doc comment (the field/variant identity finding, canonical copy).
- `crates/sm-emit/src/lib.rs` — one pre-existing test updated to decode the actual string table and assert the wire root by resolved index, since `.root` is no longer a raw comparable number.

## Tests (`tests/own0_root_identity_e2e.rs`, new — 6 tests)

- `shadowed_binding_borrow_and_write_decode_to_distinct_string_table_roots` — direct wire-level decode proving an outer borrow and an inner shadowed reassignment resolve to two distinct string-table entries (both ending `_b`). Mutation-falsified: forcing the producer to always write index 0 collapses both roots to `"Box"` (caught by this test) — restored, confirmed green.
- `shadowed_binding_write_does_not_falsely_conflict_with_outer_borrow` / `same_binding_reassignment_still_conflicts_with_its_own_borrow` — VM-behavioral positive/negative pair: shadowing doesn't cause false conflicts; a genuine same-binding conflict is still rejected.
- `corrupted_own0_root_index_rejects_deterministically_on_load` — fail-closed negative: a hand-corrupted out-of-bounds OWN0 root index is rejected with `RuntimeError::BadFormat`, never silently accepted.
- `field_symbol_collision_across_unrelated_record_types_does_not_false_conflict` / `field_symbol_same_root_same_field_still_conflicts` — the field/variant proof pair (review follow-up): two unrelated record types sharing a field spelling must not be confused (root-gated); the same field on the same root must still conflict (negative control). Both use `x = x with { field: v }` self-update reassignment specifically — confirmed by direct experimentation to be the shape that actually routes a field-level `Write` `AccessPath` through the VM's StoreVar-time checker (`ensure_write_path_allowed` only fires when the reassigned symbol already exists as a local; a `with`-update into a *new* variable never reaches that check, which would have made a naive version of this test vacuous). Mutation-falsified: removing `access_paths_overlap`'s root-gate makes the positive test fail with the exact predicted false conflict — restored, confirmed green.

Plus `crates/sm-ir/src/legacy_lowering.rs::opt_tests::ssf08_1725_nested_match_on_arm_pattern_bound_scrutinee_does_not_fail_closed_when_unneeded` (mutation-falsified: reverting the gate reproduces the exact original `"cannot resolve lexical binding 'dir'"` failure — restored, confirmed green).

Existing `runtime_ownership_e2e.rs` and `sm-vm` unit tests required zero logic changes — the #1724-era test resolver already anticipated this exact representation. One assertion updated (`vm_tracks_borrowed_paths_on_frame_push`): its expected root string was itself only coincidentally `"pair"` before this fix (a stray fail-open match via an unrelated `Call`-target string sharing that text); now correctly the local's own key.

## Golden/snapshot fallout (same representation-drift class as #1724)

Regenerated 3 fixtures pinning compiled semcode bytes (`CTF-E1-PCC4`, `CTF-E2-SEQUENCE-ITERATION`, `CTF-E2-SEQUENCE-MUTATION`) — confirmed zero-semantic-drift (`source_hash`/`ir_hash`/`verifier_status`/`vm_status`/`all_runs_equal` all byte-identical) before regenerating.

## Boundaries confirmed untouched

- **#1726** (OWN0 event timing/ordering) — open, untouched. No OWN0 event timing semantics changed.
- **#1718** (capability/path-family) — open, untouched.
- **#1888** (FA-04-025 residual) — open, untouched.
- No SemCode/artifact-format redesign in this PR — root fix stays as originally scoped; the field/variant finding is evidence-only.

## Test plan

- [x] `cargo test -p sm-ir --lib` — 152/152
- [x] `cargo test -p sm-vm --lib` — 118/118
- [x] `cargo test -p sm-runtime-core` — clean
- [x] `cargo test --test own0_root_identity_e2e` — 6/6 (new; was 4/4 before the field/variant proof pair)
- [x] `cargo test --test runtime_ownership_e2e` — 34/34 (unchanged)
- [x] `cargo test --test lexical_binding_identity_e2e` — 10/10 (unchanged)
- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo test --workspace` — clean, 0 failed
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p sm-ir/-p sm-vm/-p sm-emit/-p sm-runtime-core --all-targets -- -D warnings` — clean (sm-vm's sole warning is the pre-existing, unrelated `VmProgramView.header` dead_code finding documented in #1724, predating this work)
- [x] `git diff --check` — clean

Closes #1725

### DIFF
--- a/crates/sm-emit/src/lib.rs
+++ b/crates/sm-emit/src/lib.rs
@@ -47,6 +47,25 @@ mod tests {
         cursor
     }
 
+    // #1725 (FA-04-019): OWN0's wire `root` field is now the *index* of the
+    // lowered-local key in this same table, not a raw frontend SymbolId, so
+    // asserting its value requires decoding the actual strings to find that
+    // index - unlike `skip_string_table`, which only needs their lengths.
+    fn read_string_table(code: &[u8]) -> (Vec<String>, usize) {
+        let mut cursor = 0usize;
+        let count = read_u16_le(code, &mut cursor).expect("string count") as usize;
+        let mut strings = Vec::with_capacity(count);
+        for _ in 0..count {
+            let len = read_u16_le(code, &mut cursor).expect("string length") as usize;
+            let s = std::str::from_utf8(&code[cursor..cursor + len])
+                .expect("utf8 string")
+                .to_string();
+            cursor += len;
+            strings.push(s);
+        }
+        (strings, cursor)
+    }
+
     fn skip_optional_ownership_section(code: &[u8], mut cursor: usize) -> usize {
         if !code[cursor..].starts_with(&OWNERSHIP_SECTION_TAG) {
             return cursor;
@@ -175,7 +194,12 @@ mod tests {
         assert_ne!(spec.capabilities & CAP_OWNERSHIP_FIELD_PATHS, 0);
 
         let code = function_code(&bytes, "main");
-        let mut cursor = skip_string_table(code);
+        let (strings, mut cursor) = read_string_table(code);
+        let root_index = strings
+            .iter()
+            .position(|s| s == &borrow.path.root)
+            .expect("lowered-local root key interned in this function's string table")
+            as u32;
         assert_eq!(&code[cursor..cursor + 4], &OWNERSHIP_SECTION_TAG);
         cursor += 4;
         assert_eq!(read_u16_le(code, &mut cursor).expect("event count"), 1);
@@ -183,10 +207,7 @@ mod tests {
             read_u8(code, &mut cursor).expect("event kind"),
             OWNERSHIP_EVENT_KIND_BORROW
         );
-        assert_eq!(
-            read_u32_le(code, &mut cursor).expect("root"),
-            borrow.path.root.0
-        );
+        assert_eq!(read_u32_le(code, &mut cursor).expect("root"), root_index);
         assert_eq!(read_u16_le(code, &mut cursor).expect("component count"), 1);
         assert_eq!(
             read_u8(code, &mut cursor).expect("component kind"),
--- a/crates/sm-ir/src/legacy_lowering.rs
+++ b/crates/sm-ir/src/legacy_lowering.rs
@@ -363,14 +363,26 @@ pub enum IrInstr {
 /// Canonical execution-layer access path for ownership transport.
 ///
 /// This is intentionally IR-owned and append-only.
+///
+/// #1725 (FA-04-019): `root` is the *lowered runtime-local key* - the same
+/// `__sm_local_<id>_<spelling>` string a matching `StoreVar`/`LoadVar` for
+/// that exact binding carries (see `LoweredLocalEnv::resolve`) - not a raw
+/// frontend `SymbolId`. OWN0 emission (`emit_ownership_events`) resolves
+/// this string against the function's own `StringInterner` (the same table
+/// `LoadVar`/`StoreVar` operands are interned into) to get the artifact-
+/// local index the VM already expects on load (`build_vm_program_view_from_decoded`'s
+/// `remap_paths`). Before #1725 this field held the raw frontend `SymbolId`
+/// number directly, which the VM's *consumer* code already (silently,
+/// incorrectly) treated as a string-table index - the producer and
+/// consumer were never speaking the same identity domain.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AccessPath {
-    pub root: SymbolId,
+    pub root: String,
     pub components: Vec<PathComponent>,
 }
 
 impl AccessPath {
-    pub fn new(root: SymbolId) -> Self {
+    pub fn new(root: String) -> Self {
         Self {
             root,
             components: Vec::new(),
@@ -381,7 +393,7 @@ impl AccessPath {
         let mut components = self.components.clone();
         components.push(PathComponent::TupleIndex(index));
         Self {
-            root: self.root,
+            root: self.root.clone(),
             components,
         }
     }
@@ -390,7 +402,7 @@ impl AccessPath {
         let mut components = self.components.clone();
         components.push(PathComponent::SequenceIndexStatic(index));
         Self {
-            root: self.root,
+            root: self.root.clone(),
             components,
         }
     }
@@ -399,7 +411,7 @@ impl AccessPath {
         let mut components = self.components.clone();
         components.push(PathComponent::Field(name));
         Self {
-            root: self.root,
+            root: self.root.clone(),
             components,
         }
     }
@@ -408,7 +420,7 @@ impl AccessPath {
         let mut components = self.components.clone();
         components.push(PathComponent::AdtPayload { variant, index });
         Self {
-            root: self.root,
+            root: self.root.clone(),
             components,
         }
     }
@@ -1540,7 +1552,12 @@ fn emit_semcode_function(
             write_u16_le(&mut code, col);
         }
     }
-    emit_ownership_events(&f.ownership_events, require_ownership_section, &mut code)?;
+    emit_ownership_events(
+        &f.ownership_events,
+        require_ownership_section,
+        &interner,
+        &mut code,
+    )?;
     // #1773 (FA-09-005): unconditional, never a per-function/emission-time
     // choice like `debug_symbols` - every function gets a SIG0 section once
     // the chosen header can carry one, so `sm-format`'s decoder can expect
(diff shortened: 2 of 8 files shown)


skulmakov-oss:
@codex review

Issue #1891, opened 2026-09-05
FA-04-026: next_write_path can permanently desync from real reassignments, disabling write-conflict checks (live false negative)


skulmakov-oss:
Umbrella: #1617
Module: 04 — sm-ir / sm-vm
Audit HEAD: 8b67cc65a87e361778e7a9c597fcd530d8a524c2
Severity: P1
Classification: SOUNDNESS · FALSE-NEGATIVE · FAIL-OPEN · OWNERSHIP-ENFORCEMENT

Finding: discovered as a byproduct of #1726's adversarial evidence-gathering (explicit instruction to audit `next_write_path`'s own control-flow correctness before reusing its matching technique for Borrow-event activation, rather than trusting it because it exists). Independent of #1726 — reproduces in straight-line code with zero loops and zero branches, and does not require the Borrow/Write timing defect to manifest at all.

`Frame.next_write_path` (crates/sm-vm/src/semcode_vm.rs:82) is a single monotonic cursor into `write_paths: Vec<AccessPath>`, advanced only inside the `StoreVar` handler when `write_paths.get(cursor).filter(|path| path.root == symbol)` matches the executing StoreVar's own target symbol (semcode_vm.rs:2373-2396). On any mismatch, the check is silently skipped **and the cursor is not advanced** — there is no resync mechanism anywhere in the interpreter.

This is exploitable today because a `Write`-kind `OwnershipPathEvent`'s `root` is **not always the reassignment's own target**: `Stmt::Assign`'s lowering calls `append_record_update_write_events_from_expr` (crates/sm-ir/src/legacy_lowering.rs:6392-6397) before pushing its own target-rooted Write event, and that function's `Expr::RecordUpdate` arm (legacy_lowering.rs:10818-10842) pushes an *additional* Write event rooted at the **update expression's own source base variable**, not the assignment's target. This is confirmed by the pre-existing test `ssf08_1709_value_block_expr_statement_write_event_preserved` (legacy_lowering.rs:14941-14974): `sink(ctx with { quality: 1.0 });` — where `ctx` is never reassigned anywhere in the function — alone produces a `Write{root: ctx}` event.

Once such a foreign-rooted Write event lands in `write_paths` ahead of a real conflicting reassignment's own Write event, the cursor gets **permanently stuck** on the mismatched entry for the rest of the function: every subsequent real StoreVar peeks the same stuck entry, never matches, and its write-conflict check silently never runs again.

[...]

PR #1894, opened 2026-09-06
test(ownership): reconcile FA-04-025 after exact Write-site lowering (#1888)

skulmakov-oss:
## Summary

SSF-08 closure audit follow-up. #1888 (FA-04-025) claimed `lower_for_range_stmt`, `lower_for_each_stmt`, and `Stmt::Guard`'s `else_return` payload could skip `append_record_update_write_events_from_expr`, letting a `RecordUpdate` lower to a real `MakeRecord` with no Write ownership event at all. That prescan was, at filing time, the only mechanism producing a RecordUpdate's Write events.

Checkpoint W2A (#1891) relocated that generation into `lower_expr_with_expected`'s own `Expr::RecordUpdate` arm — the one canonical expression-lowering entry point every admitted expression must pass through to become executable IR, regardless of which statement-level caller reaches it. Confirmed by direct code reading (`lower_expr` is a thin wrapper around `lower_expr_with_expected`) that all three of #1888's named roots still never call the old prescan, and confirmed empirically that none of them need to anymore.

**This PR contains tests and one closure-audit doc correction only. Zero production Rust changed.**

## Evidence

`tests/fa_04_025_reconciliation.rs` — 8 new permanent regression tests, all green:
- For each of the three named roots (for-range bound, for-each iterable, guard else-return payload): the fixture actually reaches lowering, and the resulting `MakeRecord` carries a `write_site` correlated 1:1 with its `Write` ownership events (never merely counted globally).
- Downstream chain (one case, reusing existing #1891 infrastructure): a real, active Borrow conflicting with the for-range-bound RecordUpdate's write produces a real `BorrowWriteConflict` at the exact `MakeRecord` PC; a companion control with no borrow succeeds.
- A temporary mutation-proof test (clone the real IR, clear `ownership_events`, confirm the correlation check panics via `catch_unwind` — zero production code touched) confirmed the check discriminates presence from absence before being deleted.

`docs/roadmap/stable_foundation/ssf08_closure_audit.md` — #1888's disposition moves from REQUIRED to ALREADY SATISFIED BY LATER WORK, AC3 moves to SATISFIED, AC5's remaining blocker narrows to #1718 alone, and the minimal residual DAG drops from three independent tracks (#1718, #1888, Lane 5) to two (#1718, Lane 5). Original text is struck through rather than deleted so the audit trail stays honest about what changed and why.

## Recommendation for #1888 (not executed by this PR)

**CLOSE AS COMPLETED** — left for explicit human decision, per repository policy on issue lifecycle.

## Qualification

- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean (full `cargo clean` rebuild, no cache-masking)
- `cargo check --workspace --all-targets`: clean
- `cargo test -p sm-ir`: 198/198
- `cargo test -p sm-format`: 55/55
- `cargo test -p sm-verify --features sm-ir/profile-rust`: 212/212
- `cargo test -p sm-vm`: 120/120
- `cargo test --test write_execution_site_e2e`: 10/10
- `cargo test --test fa_04_025_reconciliation`: 8/8 (new)
- `cargo test --workspace --no-fail-fast`: 332 test-result summaries, 0 failed
- `git diff --check`: clean

### DIFF
--- a/docs/roadmap/stable_foundation/ssf08_closure_audit.md
+++ b/docs/roadmap/stable_foundation/ssf08_closure_audit.md
@@ -1,0 +1,333 @@
+# SSF-08 Closure Audit and Residual Dependency Reconciliation
+
+Status: audit only, no repair authorized by this document
+Audit SHA: `3ad550da55a4ff56db8dd33c9d82f26cc7fe8817` (exact `main`, confirmed via
+`git fetch origin && git checkout main && git pull --ff-only && git rev-parse HEAD`
+immediately before this audit; `origin/main` had not moved past this SHA)
+Umbrella: #1579 (SSF-08 — Ownership and memory model positioning), parent #1569
+Purpose: answer, with fresh repository/GitHub evidence, exactly what still
+prevents #1579 from closing. This is not an implementation checkpoint; no
+production Rust changed while producing this document.
+
+**2026-09-06 addendum — #1888 reconciled, not repaired.** This audit's
+original pass (below) classified #1888 REQUIRED against AC3/AC5, following
+the issue's own on-file description without cross-checking it against the
+later #1891/W2A architectural change. A dedicated reconciliation pass (same
+audit SHA `3ad550da5`) found, via direct code reading and 8 new empirical
+regression tests (`tests/fa_04_025_reconciliation.rs`), that #1888's root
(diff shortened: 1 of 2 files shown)


skulmakov-oss:
@codex review

Issue #1900, opened 2026-09-07
FA-08-011: EffectCalls quota counter can wrap at usize::MAX and fail open


skulmakov-oss:
Umbrella: #1617
SSF-08 umbrella #1579 remains OPEN
Lane: SSF-08 Lane 5 / AC4.a

Module owner: sm-runtime-core quota vocabulary
Enforcement site: sm-vm

Audit / discovery HEAD:
c5b01851b93d62d4c9b507c8388a44e2671e2fca

Severity: P2
Classification:
FAIL-OPEN EDGE · QUALIFICATION-GAP

## Trigger

`bump_effect_calls` computes the next usage with an unchecked integer
increment:

    vm.effect_calls + 1

while `ExecutionConfig` / `RuntimeQuotas` permit caller-supplied quota
envelopes.

At the usize numeric ceiling, release-mode integer wrapping can return the
safety-relevant counter to a small value instead of failing closed.

## Claim / confidence source

`EffectCalls` is an active, production-enforced RuntimeQuota and is part of
the Stable Foundation bounded-execution/resource contract.

## Actual guarantee

Ordinary configured limits are enforced, but the counter increment itself
is not overflow-safe.

## Gap

An active quota resource is not fail-closed over its entire representable
counter domain. This violates SSF-08 AC4.a's requirement that an active
runtime quota have deterministic exhaustion behavior.

## Required contract direction

Follow the numeric-ceiling discipline frozen for Steps/Calls in:

docs/roadmap/stable_foundation/ssf08_1759_steps_calls_contract_decision.md

- checked increment
- no wrapping
- deterministic RuntimeError::QuotaExceeded
- exact `used` where representable
- saturated `usize::MAX` reporting only for the unrepresentable
  `usize::MAX + 1` attempted-use case
- no continued effect execution after exhaustion

## Scope

This finding is independent of #1759.
Do not fold its production fix into the Steps/Calls implementation PR.

It is also not #1760; #1760 concerns TraceEntries/trace_enabled.

## Evidence

- crates/sm-vm/src/semcode_vm.rs::bump_effect_calls
- crates/sm-runtime-core/src/lib.rs::{RuntimeQuotas,QuotaKind}
- docs/spec/quotas.md
- docs/roadmap/stable_foundation/ssf08_lane5_resource_failure_closure_audit.md
- docs/roadmap/stable_foundation/ssf08_1759_steps_calls_contract_decision.md

Repair not started.

PR #1899, opened 2026-09-07
docs(quota): freeze #1759 Steps/Calls execution semantics

skulmakov-oss:
## Summary

CONTRACT DECISION ONLY. NO PRODUCTION BEHAVIOR CHANGE.

Freezes the exact Steps/Calls bounded-execution contract ahead of any
implementation:

- Step = one decoded-opcode dispatch attempt, charged before that
  instruction's semantic body executes (precedence over BorrowWriteConflict,
  semantic traps, host effects, and a same-instruction Calls charge).
- Call = one admitted, non-root `push_frame` invocation - charged last,
  after signature/Frames/StackDepth/Registers admission - using the
  `vm.callstack.len() > 0` signal `push_frame` already computes for its own
  Frames check to exempt the root/entry frame.
- Steps and Calls are orthogonal (a `CALL` opcode legitimately spends one of
  each).
- Both counters are execution-wide (not per-frame), shared across recursion
  and sibling calls.
- Raw/unverified execution paths share the same quota-bound VM engine.
- Zero-limit semantics for both quotas are frozen as direct applications of
  the general rules above, not special cases.
- **`usize::MAX` overflow is fully frozen**: a `checked_add`-based charge
  primitive, unconditionally fail-closed, with the externally reported
  `used` saturating at `usize::MAX` only for the single unrepresentable
  attempted-usage case (reporting-only - no wrap, no new error variant, no
  `RuntimeTrap` involvement). A future helper-level unit test design is
  specified that exercises this without executing `usize::MAX`
  instructions/calls.
- Failure channel is `RuntimeError::QuotaExceeded` (top-level) - explicitly
  not `RuntimeTrap::QuotaExceeded`, which stays a separate, independent,
  later question for the RuntimeTrap/RuntimeError taxonomy track.
- A reproduction-test design (non-recursive iterative Calls proof, flat
  backward-loop Steps proof) is specified, not implemented.

Full falsification-against-alternatives record is in the artifact.

**New residual finding surfaced during this pass (recorded, not fixed, not
in this issue's scope):** `EffectCalls` - an already-active, already-
enforced runtime quota - shares the same unchecked `+ 1` overflow defect
this decision closes for the new Steps/Calls counters. Classified FAIL-OPEN
EDGE / QUALIFICATION GAP against AC4.a. Not the trace-entries issue, not
repaired by this track, no tracking issue allocated yet.

Addresses issue tracked under the number one thousand seven hundred
[...]

Issue #1760, opened 2026-08-16
FA-08-002: trace_enabled and max_trace_entries expose a runtime trace contract with no production consumer


skulmakov-oss:
Umbrella: #1617
Module: **08 — `sm-runtime-core`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **INERT-CONTRACT · FALSE-READY · QUALIFICATION-GAP**

## Trigger

Configure execution with either:

```rust
ExecutionConfig { trace_enabled: true, ... }
```

or a restrictive `max_trace_entries` value.

`sm-runtime-core` exposes both as part of the shared runtime configuration/quota vocabulary, and `docs/spec/vm.md` states that the VM consumes `ExecutionConfig`, which binds trace enablement.

Repository-wide consumer tracing at the audit SHA finds no production read of `trace_enabled` outside `sm-runtime-core`, and the VM has no `QuotaKind::TraceEntries` enforcement path or trace-entry resource governed by this field.

## Claim / confidence source

- `ExecutionConfig::trace_enabled` is a public execution-config dimension.
- `RuntimeQuotas::max_trace_entries` is present in every baseline profile.
- `QuotaKind::TraceEntries` is listed by the runtime quota specification.
- `docs/spec/vm.md` says the VM consumes trace enablement as part of `ExecutionConfig`.

## Actual guarantee

Changing `trace_enabled` does not alter the production VM execution path found in this audit, and `max_trace_entries` is never charged.

The public API therefore presents a configured trace/budget surface without an authoritative execution consumer.

## Gap

A caller can configure a trace mode and trace budget that look like runtime policy but have no effect on execution. This is a classic inert-contract signal: configuration exists, is snapshotted as public API, and has baseline values, but cannot currently prove or constrain behavior.

## Evidence

- `crates/sm-runtime-core/src/lib.rs::{ExecutionConfig,QuotaKind,RuntimeQuotas}`
- `crates/sm-vm/src/semcode_vm.rs`
- `docs/spec/quotas.md`
- `docs/spec/vm.md`
- repository-wide `trace_enabled` / `TraceEntries` consumer search

Phase A evidence only. **No repair during Phase A.**

PR #1901, opened 2026-09-07
fix(vm): enforce Steps and Calls runtime quotas

### LINKED ISSUE #1759: FA-08-001: max_steps and max_calls are published bounded-execution quotas but are never enforced by sm-vm

skulmakov-oss:
## Summary

Closes #1759

Implements the contract frozen by PR #1899
(docs/roadmap/stable_foundation/ssf08_1759_steps_calls_contract_decision.md).
Does not repair the separate EffectCalls overflow finding (#1900).

- `VM.steps`/`VM.calls`: new execution-wide counters, initialized once per
  execution, never reset by frame entry/exit or recursion.
- `charge_counter(used, limit, kind)`: shared primitive for both counters.
  `checked_add(1)`; representable and within limit -> admit; representable
  and over limit -> `QuotaExceeded { kind, limit, used: next }`; increment
  itself overflows (`used` was `usize::MAX`) -> fail closed with
  `QuotaExceeded { kind, limit, used: usize::MAX }` (saturated reporting
  only, never a silent wrap, never a new error variant).
- Steps charged in the single shared instruction-dispatch loop, immediately
  after a successful opcode decode and before that instruction's semantic
  body, the write-execution-site check, and any Calls charge on the same
  `CALL` instruction.
- Calls charged in `push_frame` - the single, universal frame-construction
  choke point - immediately after the Frames/StackDepth/Registers quotas
  already admitted the invocation and strictly before any frame state
  (registers, ownership-borrow setup, the `Frame` itself) is materialized,
  matching the exact order frozen by PR #1899: Frames -> StackDepth ->
  Registers -> Calls -> construct frame, push (one combined final step
  entirely after Calls). Charged only when the callstack was non-empty
  before the push (the existing `Frames`-quota signal), so the root/entry
  frame is structurally exempt: `max_calls = 0` still runs the entry, and
  only blocks its first nested call.
- Builtins resolved inline via `try_eval_builtin_call` and effect opcodes
  never call `push_frame`, so neither consumes a Call.

## Tests

Direct helper-level `charge_counter` tests (admit under/at limit, reject at
zero/one-past limit, `usize::MAX` saturated-reporting overflow with both a
`usize::MAX` and a small distinct `limit`).

End-to-end VM regression suite: `max_steps = 0` rejects the first opcode;
an empirically-derived exact Step count completes while one fewer rejects
with the same `used`; a verified backward loop (natural bound 10_000, ~50x
the configured budget) is stopped deterministically by Steps; `max_calls =
0` succeeds for a root-only program and rejects the first nested call; an
iterative (non-recursive, so StackDepth cannot mask the quota) caller
succeeds at exactly N calls and rejects at N+1; `ClosureCall` consumes one
Call; a builtin call and a `GateRead` effect opcode consume zero Calls; a
`CALL` opcode with zero Step fuel is stopped by Steps before any Calls
charge; a nested call rejected by `Frames`, `StackDepth`, `Registers`, or
signature/arity validation does not consume a Call (four independent
rejection-path proofs, one per admission gate ahead of Calls).

## Mutation proofs (applied, confirmed predicted failure, fully reverted)

- M1 (remove Steps charge): the backward-loop test fails fast (<0.1s) -
  a small natural bound (10_000) was deliberately chosen so this mutation
  produces an ordinary test failure, not a multi-minute hang.
- M2 (move Steps charge after semantic execution): both the zero-limit and
  the Steps-before-Calls precedence tests flip from erroring to completing
  successfully.
- M3 (charge the root frame): the `max_calls = 0` root-only test fails.
- M4 (charge Calls before Frames/StackDepth/Registers): all four
  rejection-non-accounting tests fail (`vm.calls` becomes 1, not 0).
- M5 (omit `ClosureCall` charging): the ClosureCall-consumes-a-Call test
  fails.
- M6 (`wrapping_add` instead of `checked_add`): both `usize::MAX` helper
  tests fail (wraps to `Ok(0)` instead of erroring).

`grep -rn "MUTATION-"` returns zero matches on the final diff.

## Docs

Narrow updates only: `docs/spec/quotas.md` (Steps/Calls now enforced,
pointer to the frozen contract and the `EffectCalls` overflow note),
`docs/spec/vm.md` (Steps charge point, Calls root-exemption in the existing
"Callable Runtime Family Enforcement" section), and the Lane 5 audit doc
(§3 implementation-update addendum, §10 inventory row ACTIVE, §12 AC4.a
text, §8 two independently-tracked residual findings) - AC4.a is **not**
marked globally satisfied, since the `EffectCalls` overflow residual
(#1900) and #1760-#1763 are unaffected.

## Corrective checkpoints after the first push (each: fresh HEAD, fresh review)

- **Public API snapshot**: `VM`'s two new `pub` fields are a real, intended
  public API change - regenerated
  `tests/golden_snapshots/public_api/sm_vm_semcode_vm.txt` via
  `SM_UPDATE_PUBLIC_API_SNAPSHOTS=1` (2-line diff, nothing else drifted).
- **snake_learning residual, own tracking issue**: full-workspace
  qualification found `examples/benchmarks/snake_learning.sm` genuinely
  exceeds the already-published `VerifiedLocal` `max_steps = 100000`
  budget by exactly one opcode - the first real violation Steps
  enforcement exists to catch, not a defect in this implementation. Left
  as a documented known-issue (`#[ignore]`d test), filed as its own issue
  **#1902** (FA-08-012) rather than citing #1759 itself, since #1759
  closes with this PR and a citation to it would otherwise go stale.
  Recorded as Lane 5 audit §8 finding 5.
- **Calls charge placement corrected**: a fresh review found the Calls
  charge fired after the callee's registers/ownership state were already
  materialized, one step later than PR #1899's frozen order. Moved to fire
  immediately after the Registers quota check, before any frame state is
  built - see the Calls-charging bullet above. Added StackDepth,
  Registers, and signature/arity rejection-non-accounting tests
  (Frames was already covered); rewrote the Registers test after
  discovering its own construction bug (it was tripping the *caller's*
  register-growth check while marshaling 17 arguments, never reaching the
  *callee's* own check at all).
- **Fast mutation-proof guard**: shrank the backward-loop test's natural
  iteration bound from 1e9 to 10_000 so removing the Steps charge (M1) now
  fails the test in under a second instead of hanging past a 300s timeout.

## Explicitly out of scope

- The separate `EffectCalls` overflow finding (#1900)
- The `snake_learning` benchmark's own quota-exceeding content (#1902)
- `#1760`-`#1763`
- Any `RuntimeError`/`RuntimeTrap` restructuring

## Test plan

- [x] `cargo test -p sm-runtime-core --all-features` (9/9)
- [x] `cargo test -p sm-vm --all-features` (171/171, incl. 24 new)
- [x] `cargo test --workspace --no-fail-fast` (exit 0, zero failures, 1 documented ignore)
- [x] `git diff --check`
- [x] Hosted CI green at exact HEAD `ba81abb33ad0a934478bc3ec027cda609e29ee52`
- [x] Fresh Codex review clean at exact HEAD `ba81abb33a`

### DIFF
--- a/crates/sm-vm/src/semcode_vm.rs
+++ b/crates/sm-vm/src/semcode_vm.rs
@@ -179,6 +179,16 @@ pub struct VM {
     pub callstack: Vec<Frame>,
     pub config: ExecutionConfig,
     pub effect_calls: usize,
+    /// #1759: execution-wide opcode-dispatch fuel counter, per
+    /// `ssf08_1759_steps_calls_contract_decision.md` - charged once per
+    /// successfully decoded opcode in `exec_loop_with_profile`, before that
+    /// instruction's semantic body runs. Never reset by frame entry/exit.
+    pub steps: usize,
+    /// #1759: execution-wide admitted non-root call counter, per
+    /// `ssf08_1759_steps_calls_contract_decision.md` - charged once per
+    /// `push_frame` invocation that occurs while the callstack is already
+    /// non-empty. The root/entry frame never charges this counter.
+    pub calls: usize,
     pub symbols: RuntimeSymbolTable,
     /// PRNG state for random_seed / random_next_i32 (xorshift64; 0 = unseeded).
     pub prng_state: u64,
@@ -837,6 +847,8 @@ pub fn run_verified_entry_semcode_with_host_and_capabilities_and_config<
         callstack: Vec::new(),
         config,
         effect_calls: 0,
+        steps: 0,
+        calls: 0,
         symbols: program.runtime_symbols,
         prng_state: 0,
     };
@@ -861,6 +873,8 @@ pub fn run_verified_entry_semcode_with_application_host_and_capabilities_and_con
         callstack: Vec::new(),
         config,
         effect_calls: 0,
+        steps: 0,
+        calls: 0,
         symbols: program.runtime_symbols,
         prng_state: 0,
     };
@@ -944,6 +958,8 @@ pub fn run_verified_function_semcode_with_args_and_config(
         callstack: Vec::new(),
         config,
         effect_calls: 0,
+        steps: 0,
+        calls: 0,
         symbols: program.runtime_symbols,
         prng_state: 0,
     };
@@ -971,6 +987,8 @@ pub fn run_verified_entry_semcode_with_profile(
         callstack: Vec::new(),
         config,
         effect_calls: 0,
+        steps: 0,
+        calls: 0,
         symbols: program.runtime_symbols,
         prng_state: 0,
     };
@@ -1027,6 +1045,8 @@ fn run_vm_program_view_with_entry_and_config_with_observation_runtime<'a>(
         callstack: Vec::new(),
         config,
         effect_calls: 0,
+        steps: 0,
+        calls: 0,
         symbols: program.runtime_symbols,
         prng_state: 0,
     };
@@ -1992,6 +2012,14 @@ where
         let mut cur = f.instr_start + pc;
         let opcode = Opcode::from_byte(read_u8(&f.code, &mut cur).map_err(map_format_err)?)
             .map_err(map_format_err)?;
+        // #1759: the sole Steps choke point - charged immediately after a
+        // successful opcode decode, before profiling (non-authoritative),
+        // the write-execution-site check, and this opcode's own semantic
+        // body. A failed decode above never reaches this line, so a
+        // malformed opcode byte charges no Step. No refund on a later
+        // failure within this same dispatch - see
+        // ssf08_1759_steps_calls_contract_decision.md §2-§3.
+        vm.steps = charge_counter(vm.steps, vm.config.quotas.max_steps, QuotaKind::Steps)?;
         profile.record_opcode(opcode);
         // #1891 Checkpoint W2F: check every Write path attached to this
         // exact PC against the frame's currently active borrows BEFORE the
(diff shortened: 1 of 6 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1904, opened 2026-09-07
docs(quota): freeze #1761 ConstPool contract disposition

skulmakov-oss:
## Summary

Addresses #1761
SSF-08 umbrella #1579 remains OPEN

CONTRACT DECISION ONLY
NO PRODUCTION BEHAVIOR CHANGE

Freezes the disposition of `RuntimeQuotas::max_const_pool` /
`QuotaKind::ConstPool` as **REMOVE**:

- Fresh, independent re-derivation confirms no real ConstPool-shaped
  resource exists anywhere in the current codebase (`sm-runtime-core`,
  `sm-vm`, `sm-format`, `sm-verify`, `sm-emit` all checked). Every
  candidate value-storage mechanism (per-function string tables, the
  program-wide `RuntimeSymbolTable`, `SIG0` signature tables, inline
  instruction immediates) is either a decode-time structural bound owned
  by `sm-format`, already governed by the distinct, already-enforced
  `QuotaKind::SymbolTable`, or has no pooling mechanism at all.
- RE-SCOPE is falsified: every candidate referent either duplicates an
  existing, correctly-named quota kind (worst case: `SymbolTable`) or is
  not a resource at all.
- A RESERVED/deprecated compatibility surface is falsified: no
  compatibility authority requires one, and this codebase's own dominant
  struct-literal-with-`..Default` construction pattern would make a
  `#[deprecated]` attribute silently ineffective at most call sites.
- Removal is confirmed **source-visible**: the field and enum variant are
  locked into `tests/golden_snapshots/public_api/sm_runtime_core_lib.txt`.
  No current compatibility policy blocks intentional narrowing of this
  surface today - the platform's own stability/compatibility policy stack
  (`docs/roadmap/language_maturity/stability_and_compatibility.md`) is
  itself only "proposed v0," with no stability labels yet assigned to any
  `sm-runtime-core` surface.
- No compatibility ghost: the future implementation must not keep
  `max_const_pool` inert-but-present, silently map it to `SymbolTable` or
  string-table limits, or return a fake `QuotaExceeded(ConstPool)`.
- `#1763` consequence recorded: once REMOVE lands, `RuntimeError::
  QuotaExceeded` can never legitimately carry `kind: ConstPool` (already
  true today; this makes it structural). `RuntimeTrap`/`RuntimeError`
  themselves are not touched by this checkpoint.

## Changes

- Add `docs/roadmap/stable_foundation/ssf08_1761_constpool_contract_decision.md`
[...]

Issue #1902, opened 2026-09-07
FA-08-012: snake_learning benchmark exceeds the published VerifiedLocal max_steps budget


skulmakov-oss:
Umbrella: #1617
SSF-08 umbrella #1579 remains OPEN
Lane: SSF-08 Lane 5 / #1759 follow-on residual

Module owner: examples/benchmarks
Discovery HEAD: on branch fix/1759-steps-calls-runtime-enforcement, at the
commit that first enforced Steps/Calls (see PR #1901)

Severity: P3
Classification: QUOTA VIOLATION SURFACED BY NEWLY-REAL ENFORCEMENT

## Trigger

`examples/benchmarks/snake_learning.sm`, run via `smc run` under the
`VerifiedLocal` execution context (`RuntimeQuotas::verified_local()`,
`max_steps = 100000`), now fails:

    QuotaExceeded { kind: Steps, limit: 100000, used: 100001 }

`tests/snake_learning_benchmark.rs::snake_learning_passes_check_run_compile_verify`
is `#[ignore]`d citing this issue rather than left failing or silently
patched.

## Why this is not a defect in #1759

#1759 (FA-08-001) made the already-published `max_steps = 100000` baseline
for `VerifiedLocal` real for the first time. This benchmark's own step
usage always nominally exceeded that budget - it simply went undetected
because the quota was never enforced. This is the first real violation
Steps enforcement exists to catch, not a bug in the enforcement itself.

## Scope

This finding is independent of #1759 and must not be folded into its
implementation PR. Fixing it requires a decision the #1759 checkpoint is
not authorized to make unilaterally - candidates, in no particular order:

- Reduce `snake_learning.sm`'s own simulation parameters (`n_episodes` /
  its internal per-episode `max_steps`) so it fits the existing budget.
- Add a CLI mechanism to `smc run` for selecting a higher-budget execution
  context (e.g. `kernel_bound`, `max_steps = 250000`) or an explicit quota
  override - `cmd_run_controlled_observation` currently hardcodes
  `ExecutionConfig::for_context(ExecutionContext::VerifiedLocal)` with no
  override surface.
- Raise the published `verified_local` baseline itself - a numeric-contract
  change requiring its own review, not a side effect of an unrelated PR.

## Evidence

- examples/benchmarks/snake_learning.sm
- tests/snake_learning_benchmark.rs
- crates/smc-cli/src/app.rs::cmd_run_controlled_observation
[...]

Issue #1762, opened 2026-08-16
FA-08-004: ExecutionConfig can decouple ExecutionContext from its quota baseline while audit records only the context label


skulmakov-oss:
Umbrella: #1617
Module: **08 — `sm-runtime-core`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **IDENTITY/TRUST-GAP · INERT-CONTRACT · FALSE-READY · QUALIFICATION-GAP**

## Trigger

Construct an `ExecutionConfig` whose context and quotas disagree, for example:

```rust
ExecutionConfig::new(
    ExecutionContext::PureCompute,
    RuntimeQuotas::verified_local(),
)
```

or use the public fields to assign any custom quota envelope while keeping a canonical context label.

`ExecutionConfig::for_context()` has the documented mapping, but neither `ExecutionConfig::new()` nor the public struct validates that the supplied quotas remain within the baseline associated with the context.

The VM does not read `config.context`; it enforces only `config.quotas`.

`prom-runtime` then copies only `config.context` into `RuntimeSessionDescriptor`, and `prom-audit::AuditSessionMetadata` records that context label but not the actual quota envelope.

## Claim / confidence source

`docs/spec/quotas.md` states:

- context selection is explicit through `ExecutionConfig`;
- `ExecutionContext -> RuntimeQuotas` has a defined baseline mapping;
- higher integration layers may choose context-specific envelopes but must not silently weaken the core safety contract.

`docs/spec/vm.md` says context selects the runtime quota baseline.

## Actual guarantee

The type allows arbitrary context/quota pairings, VM semantics depend on quotas rather than context, and PROMETHEUS audit/session metadata preserves only the context label.

Thus two materially different execution envelopes can produce the same recorded `ExecutionContext` identity.

## Gap

The public context label can overstate what quota policy actually governed execution. This does not itself bypass capabilities — capability policy remains a separate boundary — but it breaks provenance/interpretability of the runtime configuration and makes the documented context-to-baseline relation non-authoritative.

## Evidence

- `crates/sm-runtime-core/src/lib.rs::{ExecutionConfig,ExecutionContext,RuntimeQuotas}`
- `crates/sm-vm/src/semcode_vm.rs` — no `config.context` consumer; quota enforcement reads `config.quotas`
- `crates/prom-runtime/src/lib.rs::{ExecutionSession,RuntimeSessionDescriptor}`
[...]

PR #1905, opened 2026-09-08
fix(quota): remove inert ConstPool runtime vocabulary

### LINKED ISSUE #1761: FA-08-003: ConstPool quota is public runtime vocabulary without a corresponding VM resource or enforcement path
Umbrella: #1617
Severity: **P2**
Classification: **INERT-CONTRACT · FALSE-READY · QUALIFICATION-GAP**

## Trigger

Rely on the configured `RuntimeQuotas::max_const_pool` / `QuotaKind::ConstPool` as a bound on runtime resource usage.

Every baseline profile publishes:

```text
max_const_pool = 65_536
```

and `docs/spec/quotas.md` lists `ConstPool` as one of the current quota kinds.

At the audit SHA the VM has no `QuotaKind::ConstPool` enforcement call and no corresponding runtime constant-pool object whose size is checked against this field. The active SemCode value-loading paths use instruction operands and function-local string tables rather than a shared runtime resource named by this quota.

## Claim / confidence source

`sm-runtime-core` is the quota model owner, and the quota specification presents `ConstPool` alongside actually enforced resources such as frames/registers/effects as part of the current runtime quota taxonomy.

## Actual guarantee

`max_const_pool` is a public number with no demonstrated production consumer in `sm-vm` at this audit SHA.

## Gap

The runtime contract exposes a resource budget that cannot currently reject or constrain anything. A caller inspecting `RuntimeQuotas` can reasonably believe the const-pool resource is bounded, while execution provides no such proof.

This is P2 rather than P1 because no active VM resource matching this quota was identified; the defect is an inert/overstated contract rather than a demonstrated unbounded active pool.

## Evidence

- `crates/sm-runtime-core/src/lib.rs::{QuotaKind,RuntimeQuotas}`
- `crates/sm-vm/src/semcode_vm.rs`
- `docs/spec/quotas.md`
- repository-wide `max_const_pool` / `QuotaKind::ConstPool` consumer search

skulmakov-oss:
## Summary

Closes #1761

Implements the REMOVE disposition frozen on main by PR #1904.
This is an intentional source-visible public API narrowing.

- Removes `QuotaKind::ConstPool` and `RuntimeQuotas::max_const_pool` from
  `crates/sm-runtime-core/src/lib.rs`: the enum variant, the struct field,
  the matching assignment in `verified_local`/`pure_compute`/`kernel_bound`,
  and the `RuntimeQuotas::exceed()` match arm.
- No replacement field, no re-scope to `SymbolTable`/string-table
  limits/`SIG0`, no deprecated alias, no compatibility ghost.

## Unexpected production consumers

**None.** `cargo check --workspace --all-targets` immediately after the
Rust removal produced zero compile errors anywhere in the workspace -
empirical confirmation, not inference, that `ConstPool` had no live
downstream consumer. The "STOP if an unexpected consumer appears" gate in
this checkpoint's own brief did not fire.

## Public API guard - deliberate RED, then approved GREEN

With the Rust surface already removed but the golden snapshot still
stale, `public_api_contracts` failed RED, and the diff showed precisely
`ConstPool,` and `pub max_const_pool: usize,` disappearing - nothing else.
The golden was then regenerated via the existing
`SM_UPDATE_PUBLIC_API_SNAPSHOTS=1` mechanism and the resulting 2-line diff
reviewed manually. A temporary reinsertion of the enum variant (reverted
immediately after) confirmed the same guard also catches `ConstPool`'s
return, not just its departure.

## Docs

- `docs/spec/quotas.md`: removes `ConstPool` from the taxonomy,
  descriptor-field list, and all three baseline-profile sections. Also
  narrows the file's blanket "Enforcement owner: `sm-vm`" statement
  (header, "Quota model rule", "Ownership Rule") to name `sm-verify`'s
  static enforcement of `SymbolTable` explicitly - the Lane 5 audit
  already proved this discrepancy, and this PR was already the
  authoritative edit point for this exact file.
- `docs/roadmap/stable_foundation/ssf08_1761_constpool_contract_decision.md`:
  preserved unedited as historical evidence; a new "Implementation update"
  section appended recording the outcome.
- `docs/roadmap/stable_foundation/ssf08_lane5_resource_failure_closure_audit.md`:
  `#1761` moves from "disposition frozen, not yet implemented" to "REMOVE
  implemented"; the inventory no longer lists `ConstPool` as a current
  `RuntimeQuotas`/`QuotaKind` member; AC4.b text updated - no longer
  blocked by `ConstPool`, still not satisfied globally pending `#1760`.

## Tests

One focused `sm-runtime-core` test added,
`removing_const_pool_leaves_every_surviving_baseline_value_unchanged`,
pinning every surviving quota value across all three baseline profiles -
so an accidental value change hiding inside this removal would fail
loudly rather than pass silently. **No surviving quota baseline changed.**

## Residue audit

Zero active `ConstPool`/`max_const_pool` references in `crates/**`,
`docs/spec/**`, or the public API golden. Remaining mentions are
historical only: this PR's own new test's doc comment, the decision
document, the Lane 5 audit, and pre-existing SSF-08 audit-trail documents
from before this checkpoint (`ssf08_closure_audit.md`,
`ssf08_ownership_position_decision.md`,
`semantic_stable_foundation_matrix.md`) - all correctly describing the
removed historical surface, none claiming it currently exists.

## #1763 consequence (recorded, not acted on)

After this removal, `RuntimeError::QuotaExceeded` can never carry
`QuotaKind::ConstPool` because the variant no longer exists - structural,
not merely coincidental. `RuntimeTrap`, `RuntimeError`, and `#1763`
taxonomy are untouched by this PR.

## Explicitly out of scope

- `#1760` `TraceEntries` semantics
- `#1762` `ExecutionContext` provenance
- `#1763` `RuntimeTrap`/`RuntimeError`
- `#1902` snake_learning
- SemCode format
- Verifier policy unrelated to compile fallout (none found)

## Test plan

- [x] `cargo check --no-default-features` (no_std gate)
- [x] `cargo test -p sm-runtime-core --all-features` (10/10)
- [x] `cargo test --test public_api_contracts` (88/88, RED-then-GREEN proof)
- [x] `cargo test --workspace --no-fail-fast` (exit 0, zero failures)
- [x] `git diff --check`
- [x] Hosted CI green at exact HEAD `2775ad17d2699bcf03d0322240ca03696c863042`
- [x] Fresh Codex review clean at exact HEAD `2775ad17d2`


skulmakov-oss:
@codex review

PR #1908, opened 2026-09-08
docs(quota): freeze #1762 execution-envelope provenance contract

skulmakov-oss:
## Summary

Addresses #1762
SSF-08 umbrella #1579 remains OPEN

CONTRACT DECISION ONLY
NO PRODUCTION BEHAVIOR CHANGE
NO AUDIT FORMAT CHANGE

Freezes the contract for the relationship between `ExecutionContext`,
`RuntimeQuotas`, `ExecutionConfig`, `RuntimeSessionDescriptor`, and
`AuditSessionMetadata`.

## Headline finding

Fresh evidence confirms `RuntimeQuotas` (as carried in
`ExecutionConfig.quotas`) is the effective authority for the quota/profile
dimensions it represents - `sm-vm` runtime-quota charging, plus the
`sm-verify` resource checks that explicitly consume a `RuntimeQuotas`
value. This does not subsume `VerificationLimits` (an orthogonal verifier
envelope), `sm-format`'s structural decode-time caps, verifier rules
unrelated to `RuntimeQuotas`, or capability policy. `sm-vm` never reads
`config.context` anywhere in its execution path, and `sm-verify`'s
`RuntimeQuotas`-consuming checks never see `ExecutionContext` at all.
`ExecutionConfig::new` is a legitimate, actively-used public seam that
pairs an arbitrary `ExecutionContext` with arbitrary `RuntimeQuotas`
(exercised today by the SSF-04 effect-quota qualification tests in
`tests/ssf04_effect_quota.rs`, tightening `max_effect_calls` well below
the `verified_local` baseline). But `RuntimeSessionDescriptor` and
`AuditSessionMetadata` record only the `ExecutionContext` label - never
the effective `RuntimeQuotas` - so two sessions under the identical
`VerifiedLocal` label, one running the canonical baseline and one running
a custom, tightened envelope, are recorded with an identical `session`
line with respect to the effective quota envelope: the archive contains
no field from which a reader can distinguish those two quota
configurations (differing quotas can still alter what actually executes -
this is a claim about representational provenance loss, not that entire
archives must be byte-identical). A reader of the archive alone cannot
determine which envelope actually governed either run.

## Selected disposition: RECORD_EFFECTIVE

- `ExecutionContext` — frozen as a construction-time baseline-selector and
  audit-class label, not proof of the actual effective envelope.
- `RuntimeQuotas` — frozen as the effective authority for the quota/profile
  dimensions it represents (not a claim over `VerificationLimits`,
[...]

PR #1911, opened 2026-09-09
fix(audit): record effective runtime quota envelope

### LINKED ISSUE #1762: FA-08-004: ExecutionConfig can decouple ExecutionContext from its quota baseline while audit records only the context label

skulmakov-oss:
## Summary

Closes #1762

Implements the RECORD_EFFECTIVE contract frozen by PR #1908
(`docs/roadmap/stable_foundation/ssf08_1762_execution_envelope_provenance_decision.md`).

- records `RuntimeQuotas` in `RuntimeSessionDescriptor`
- propagates the same effective envelope into `AuditSessionMetadata`
- persists all eight fields in canonical audit archive v2
- keeps the multi-session outer format at v1
- rejects legacy audit archive v1 rather than reconstructing missing quotas
- does not restrict custom `ExecutionConfig` envelopes
- threads exactly one `ExecutionConfig` through admission, execution, and
  audit provenance in `smc-cli` - never reconstructing a second,
  independently-hardcoded config that happens to agree today

Intentional public API widening:
- `RuntimeSessionDescriptor::quotas`
- `AuditSessionMetadata::quotas`
- `sm_vm::run_semcode_collecting_hello_observations_with_config` (new
  config-aware sibling, added during review - see "Corrective round" below)

## The central trust invariant, proven end to end

`ExecutionContext` alone never proved which `RuntimeQuotas` actually
governed a session - `ExecutionConfig::new` legitimately pairs any context
with any quotas (the SSF-04 effect-quota qualification seam has always
done exactly this). Before this PR, `RuntimeSessionDescriptor` and
`AuditSessionMetadata` recorded only the context label, so two sessions
under `VerifiedLocal` - one at the canonical baseline, one with
`max_effect_calls` tightened to `1` - produced an identical `session` line
in the canonical audit archive.

Now: `ExecutionConfig.quotas` is copied one-way into
`RuntimeSessionDescriptor` → `AuditSessionMetadata` → the canonical
archive, never re-derived from `context`. A dedicated regression proves a
distinct custom envelope (values 101-108, chosen so none could be mistaken
for a baseline) survives `ExecutionConfig` →
`AuditReplayArchive::to_canonical_text` →
`AuditReplayArchive::from_canonical_text` → `ReplayMetadata`
value-for-value, and a `MultiSessionReplayArchive` regression proves two
sessions sharing one `ExecutionContext` with different `RuntimeQuotas`
remain distinguishable after a full canonical round-trip.

## Changes

- `crates/prom-runtime/src/lib.rs`: `RuntimeSessionDescriptor` gains
  `pub quotas: RuntimeQuotas`. `ExecutionSession::new` and
  `GateExecutionSession::new` copy `quotas: config.quotas` alongside the
  pre-existing `context: config.context`. `build_audit_session` copies
  `quotas: descriptor.quotas` one hop further - the
  `ExecutionConfig` → `RuntimeSessionDescriptor` → `AuditSessionMetadata`
  path is unchanged in shape, only widened.
- `crates/prom-audit/src/lib.rs`: `AuditSessionMetadata` gains
  `pub quotas: RuntimeQuotas`. `AUDIT_REPLAY_ARCHIVE_FORMAT_VERSION`
  bumped `1` → `2`; `MULTI_SESSION_REPLAY_ARCHIVE_FORMAT_VERSION` stays at
  `1` (verified: `MultiSessionReplayArchive::from_canonical_text`
  genuinely delegates each embedded session to
  `AuditReplayArchive::from_canonical_text`, so the outer wire shape is
  untouched). The canonical `session` line now emits 13 tab-separated
  tokens (was 5), the eight `RuntimeQuotas` fields appended in struct
  declaration order as plain decimal text. The parser consumes all eight
  via the existing `parse_usize_field` discipline into a `RuntimeQuotas`
  literal - no defaults, no baseline fallback. Four pre-existing hand-built
  raw-text fixtures (testing event-count-mismatch / non-monotonic-id
  rejection) were updated to the new wire shape so they keep testing what
  they were written to test, rather than failing on version mismatch.
- `crates/sm-vm/src/semcode_vm.rs`: new `pub fn
  run_semcode_collecting_hello_observations_with_config(bytes, config)`,
  the config-aware sibling of `run_semcode_collecting_hello_observations`
  (mirroring the existing `run_semcode_with_entry`/`run_semcode_with_config`
  pattern). The no-config wrapper now delegates to it with the canonical
  default, preserving its existing behavior/signature for other callers.
- `crates/smc-cli/src/app.rs`: `collect_controlled_observation_envelope`
  now constructs exactly **one** `ExecutionConfig` and threads it through
  all three consumers - `verify_semcode_token_with_quotas(bytes,
  execution_config.quotas)` for admission,
  `run_semcode_collecting_hello_observations_with_config(bytes,
  execution_config)` for execution, and the same binding's `.context`/
  `.quotas` for the audit metadata literal. See "Corrective round" below
  for why the first attempt at this site was insufficient.
- `tests/prometheus_audit.rs`: two direct `AuditSessionMetadata` literals
  fixed - one copies `session.descriptor().quotas` (the real session
  object already available), one is a purely synthetic fixture using
  `RuntimeQuotas::kernel_bound()` matching its own labeled context.
- `tests/golden_snapshots/public_api/prom_runtime_lib.txt`,
  `tests/golden_snapshots/public_api/prom_audit_lib.txt`,
  `tests/golden_snapshots/public_api/sm_vm_semcode_vm.txt`: updated via
  `SM_UPDATE_PUBLIC_API_SNAPSHOTS=1` after proving the guard fails RED
  first, in each case.
- `docs/spec/quotas.md`, `docs/spec/audit.md`: active-document the
  implemented provenance rule, the archive wire/version consequence, and
  the explicit scope boundary (this does not extend to
  `VerificationLimits`, `sm-format`'s structural caps, or capability
  policy). `docs/spec/vm.md` inspected; no active contradiction found, left
  unchanged.
- `docs/roadmap/stable_foundation/ssf08_1762_execution_envelope_provenance_decision.md`:
  original decision text preserved unedited; §25 implementation-status
  addendum appended, then corrected once review caught it still described
  the rejected first-attempt CLI fix (see "Corrective round" below).
- `docs/roadmap/stable_foundation/ssf08_lane5_resource_failure_closure_audit.md`:
  §6 implementation-complete addendum, AC4.c text updated to SATISFIED
  (freshly re-audited against this branch); addendum heading reworded to
  a merge-stable form so it does not read as false once this PR merges.

## A gap discovered mid-implementation, resolved with explicit owner input

`tests/golden_snapshots/public_api/prom_audit_lib.txt` existed, but
`tests/public_api_contracts.rs` never actually tracked
`crates/prom-audit/src/lib.rs` in its guarded-file list - a pre-existing
omission unrelated to #1762, which meant `AuditSessionMetadata`'s own
public API (the exact type this PR widens) was unguarded going in. Flagged
to the owner before proceeding; the owner chose to restore `prom-audit` to
the tracked list within this same PR. Doing so surfaced accumulated,
unrelated drift (expanded `AuditEventKind` variant text, multi-line
constructor signatures reflecting source changes that predate this PR)
alongside the two intended changes (`quotas` field, format-version bump);
the full diff was reviewed manually before regenerating.

## Corrective round: the CLI fix's first attempt was itself a reconstruction

Review caught that the first version of the `smc-cli` fix constructed a
**second**, independently-hardcoded `ExecutionConfig::for_context(VerifiedLocal)`
purely to source the audit metadata, while the actual execution
(`run_semcode_collecting_hello_observations`, no config parameter existed)
still hardcoded its **own**, separate canonical config internally. Both
values agreed today by construction, but nothing tied them together - a
second instance of exactly the "reconstruct a nicer-looking policy after
the fact" pattern #1762's own trust rule forbids, just one level removed
from the original defect. Fixed by adding the config-aware `sm-vm` sibling
above and constructing the `ExecutionConfig` **once** in `smc-cli`,
threading that single instance through verification, execution, and
provenance. A dedicated `sm-vm` regression
(`run_semcode_collecting_hello_observations_with_config_enforces_the_passed_quotas`)
proves the new function genuinely enforces the caller's quotas rather than
silently falling back to a hardcoded baseline - see **M4** below.

A second review pass then caught that the §25/§6 decision-doc addenda
still described the rejected first attempt and counted only three
mutation proofs; both were corrected in a docs-only follow-up commit
(no production code, tests, snapshots, or archive format touched).

## Mutation proofs (applied and reverted, zero markers remain)

- **M1** (provenance reconstruction): `ExecutionSession::new`'s descriptor
  copy temporarily replaced with a context-derived
  `RuntimeQuotas::verified_local()`. The custom-envelope propagation
  regression turned RED as predicted. Reverted.
- **M2** (parser reconstruction): the archive parser's eight quota fields
  temporarily replaced with a hardcoded `RuntimeQuotas::verified_local()`
  regardless of what was actually serialized. The full-pipeline custom
  envelope regression turned RED. Reverted.
- **M3** (version discipline): `AUDIT_REPLAY_ARCHIVE_FORMAT_VERSION` left
  at `1` after the wire-shape change. Both the v1-rejection and
  embedded-v1-rejection regressions turned RED (wrong error message,
  proving the version gate itself is what makes rejection meaningful, not
  incidental). Reverted.
- **M4** (config-blind fallback): `run_semcode_collecting_hello_observations_with_config`
  temporarily ignored its own `config` parameter and hardcoded canonical
  `VerifiedLocal` internally instead. The passed-quotas-enforcement
  regression turned RED (a program that should have failed under a
  drastically tightened `max_steps` succeeded instead). Reverted.

## AC4.c, freshly re-audited (not assumed)

Every production construction site of `RuntimeSessionDescriptor` and
`AuditSessionMetadata` was manually classified as either SAME AUTHORITY
(copies an available `config.quotas`/`descriptor.quotas`) or explicit
synthetic/audit-only fixture. Zero production paths discard an available
effective envelope in favor of a context-derived reconstruction, including
the corrected `smc-cli` site, which now sources verify/execute/audit from
one shared binding rather than three independent (if value-equal)
constructions. **AC4.c: SATISFIED.** `AC4.d`/`AC4.e` remain `#1763`
territory, not claimed here.

## Explicitly out of scope

- Moving quota storage off `RuntimeQuotas` (digest/delta/boolean alternatives)
- `VerificationLimits`, `sm-format`'s structural caps, capability policy
- `RuntimeQuotas` baseline value changes
- Restricting custom `ExecutionConfig` envelopes
- `MULTI_SESSION_REPLAY_ARCHIVE_FORMAT_VERSION` bump
- Backward-compatible v1 archive parsing
- `RuntimeTrap`/`RuntimeError` restructuring (`#1763`)
- `#1902` snake_learning

## Test plan

- [x] `cargo test -p sm-vm --lib` (incl. the new config-enforcement regression)
- [x] `cargo test -p prom-audit --all-features` (26/26)
- [x] `cargo test -p prom-runtime --all-features` (10/10)
- [x] `cargo test --test prometheus_audit` (2/2)
- [x] `cargo test --test public_api_contracts` (RED with stale golden each
      time a surface widened - `prom-runtime`, `prom-audit`, then
      `sm-vm`'s new helper - isolated diffs reviewed; GREEN after
      regeneration, 88/88)
- [x] `cargo test --workspace --no-fail-fast` (zero failures)
- [x] `cargo check -p prom-audit --no-default-features`,
      `cargo check -p prom-runtime --no-default-features` (both hit a
      pre-existing, unrelated `format!`-macro no_std bug in `prom-cap`/
      `prom-state`/`prom-gates` - empirically confirmed identical on the
      pre-#1762 baseline commit in a throwaway worktree, not caused by
      this change)
- [x] `git diff --check`
- [x] Custom-envelope full-pipeline exact-preservation regression
- [x] Same-context/different-quotas multi-session distinguishability regression
- [x] Config-aware `sm-vm` helper genuinely enforces passed-in quotas regression
- [x] M1/M2/M3/M4 mutation proofs (RED, reverted)
- [x] Hosted CI green at exact HEAD fd3d9c47ffc1c859c5b2486c255008aee17be149
- [x] Fresh Codex review clean at exact HEAD fd3d9c47ff

**Exact-head qualification complete. Wait for owner GO before merge.**

### DIFF
--- a/crates/prom-audit/src/lib.rs
+++ b/crates/prom-audit/src/lib.rs
@@ -8,7 +8,7 @@ use alloc::string::String;
 use alloc::vec::Vec;
 use hello_observation_audit::ControlledObservationAuditDecision;
 use prom_cap::{CapabilityKind, CapabilityManifestMetadata, CapabilityManifestVersion};
-use sm_runtime_core::ExecutionContext;
+use sm_runtime_core::{ExecutionContext, RuntimeQuotas};
 
 const AUDIT_REPLAY_ARCHIVE_MAGIC: &str = "semantic_audit_replay_archive";
 const MULTI_SESSION_REPLAY_ARCHIVE_MAGIC: &str = "semantic_multi_session_replay_archive";
@@ -19,6 +19,7 @@ pub struct AuditEventId(pub u64);
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AuditSessionMetadata {
     pub context: ExecutionContext,
+    pub quotas: RuntimeQuotas,
     pub capability_manifest: CapabilityManifestMetadata,
     pub gate_registry_bound: bool,
 }
@@ -80,7 +81,7 @@ pub struct ReplayMetadata {
     pub last_event_id: Option<AuditEventId>,
 }
 
-pub const AUDIT_REPLAY_ARCHIVE_FORMAT_VERSION: u32 = 1;
+pub const AUDIT_REPLAY_ARCHIVE_FORMAT_VERSION: u32 = 2;
 pub const MULTI_SESSION_REPLAY_ARCHIVE_FORMAT_VERSION: u32 = 1;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -197,6 +198,28 @@ impl AuditReplayArchive {
         } else {
             "false"
         });
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_steps.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_calls.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_stack_depth.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_frames.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_registers.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_symbol_table.to_string());
+        out.push('\t');
+        out.push_str(&self.session.quotas.max_effect_calls.to_string());
+        out.push('\t');
+        out.push_str(
+            &self
+                .session
+                .quotas
+                .max_debug_symbols_per_function
+                .to_string(),
+        );
         out.push('\n');
         out.push_str("events\t");
         out.push_str(&self.events.len().to_string());
@@ -242,7 +265,7 @@ impl AuditReplayArchive {
             .next()
             .ok_or_else(|| AuditReplayArchiveFormatError::new("missing session line"))?;
         let session_parts = split_archive_line(session_line);
-        if session_parts.len() != 5 || session_parts[0] != "session" {
+        if session_parts.len() != 13 || session_parts[0] != "session" {
             return Err(AuditReplayArchiveFormatError::new("invalid session line"));
         }
         let session = AuditSessionMetadata {
@@ -252,6 +275,19 @@ impl AuditReplayArchive {
                 version: parse_manifest_version(session_parts[3])?,
             },
             gate_registry_bound: parse_bool_field(session_parts[4], "gate registry bound")?,
+            quotas: RuntimeQuotas {
+                max_steps: parse_usize_field(session_parts[5], "max_steps")?,
+                max_calls: parse_usize_field(session_parts[6], "max_calls")?,
+                max_stack_depth: parse_usize_field(session_parts[7], "max_stack_depth")?,
+                max_frames: parse_usize_field(session_parts[8], "max_frames")?,
+                max_registers: parse_usize_field(session_parts[9], "max_registers")?,
+                max_symbol_table: parse_usize_field(session_parts[10], "max_symbol_table")?,
+                max_effect_calls: parse_usize_field(session_parts[11], "max_effect_calls")?,
+                max_debug_symbols_per_function: parse_usize_field(
+                    session_parts[12],
+                    "max_debug_symbols_per_function",
+                )?,
+            },
         };
 
         let events_line = lines
@@ -980,6 +1016,7 @@ mod tests {
     fn sample_session() -> AuditSessionMetadata {
         AuditSessionMetadata {
             context: ExecutionContext::KernelBound,
+            quotas: RuntimeQuotas::kernel_bound(),
             capability_manifest: CapabilityManifestMetadata {
                 schema: "prom.cap.manifest".to_string(),
                 version: prom_cap::CapabilityManifestVersion::V1,
@@ -1271,8 +1308,8 @@ mod tests {
     #[test]
     fn replay_archive_rejects_event_count_mismatch() {
         let text = "\
-semantic_audit_replay_archive\t1\n\
-session\tkernel-bound\tprom.cap.manifest\tv1\ttrue\n\
+semantic_audit_replay_archive\t2\n\
+session\tkernel-bound\tprom.cap.manifest\tv1\ttrue\t250000\t32768\t256\t256\t8192\t16384\t4096\t16384\n\
 events\t1\n\
 replay\t0\tnone\n";
 
@@ -1284,8 +1321,8 @@ replay\t0\tnone\n";
     #[test]
     fn replay_archive_rejects_non_monotonic_event_ids() {
         let text = "\
-semantic_audit_replay_archive\t1\n\
-session\tkernel-bound\tprom.cap.manifest\tv1\ttrue\n\
+semantic_audit_replay_archive\t2\n\
+session\tkernel-bound\tprom.cap.manifest\tv1\ttrue\t250000\t32768\t256\t256\t8192\t16384\t4096\t16384\n\
 events\t1\n\
 event\t9\tsession-finished\n\
 replay\t1\t9\n";
@@ -1355,7 +1392,7 @@ replay\t1\t9\n";
         assert_eq!(lines[0], "semantic_multi_session_replay_archive\t1");
         assert_eq!(lines[1], "sessions\t2");
         assert!(lines[2].starts_with("session\t0\t"));
-        assert!(lines[3].starts_with("archive\tsemantic_audit_replay_archive\t1"));
+        assert!(lines[3].starts_with("archive\tsemantic_audit_replay_archive\t2"));
     }
 
     #[test]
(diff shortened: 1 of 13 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

Issue #1920, opened 2026-09-12
compile_program_to_ir_with_options_and_profile probes Logos unconditionally (including under explicit RustLike) and swallows a real Logos failure under Auto


skulmakov-oss:
`crates/sm-ir/src/legacy_lowering.rs:1201-1220` computes
`logos_detected` via `parse_logos_program_with_profile(input,
parser_profile).map(...).unwrap_or(false)` **unconditionally**, for
every `CompileProfile` value, before `profile` is consulted for this
purpose. Two independent defects follow:

1. Under `CompileProfile::Auto`, a genuine Logos parse/policy failure
   collapses to `false` via `unwrap_or`, and execution silently falls
   through to RustLike lowering on the same input without surfacing the
   real failure - the same defect class as #1670, in a different
   function.
2. Under explicit `CompileProfile::RustLike`, the Logos parser is
   invoked at all, with any error discarded, before `profile` is ever
   branched on - a real probe of the non-selected grammar under an
   explicit single-grammar request, independent of whether its result
   currently changes `RustLike`'s outcome.

Found during the SSF09-E1 Diagnostic Authority Decision's Source
Surface Authority evidence pass
(`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`,
Decision A, path 3); not covered by #1670's filed scope. Repair must
follow that decision's fail-closed invariant: explicit `RustLike` must
not invoke the Logos parser at all, and `Auto`'s classification must
distinguish a real failure from "no Logos content" instead of
collapsing both to `false`.

Related: #1580, #1670

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1919, opened 2026-09-12
smc check discards a real multi-module project failure in favor of a narrower single-file fallback


skulmakov-oss:
`cmd_check` (`crates/smc-cli/src/app.rs:588-589`) calls
`check_file_with_provider_and_profile` first - a strict, Logos-only
multi-module project loader (`crates/sm-sema/src/std_adapters.rs:134-171`,
`load_module_recursive` returns a real `Err`/`E0239` on any non-Logos
or otherwise invalid module, with no fallback internally). On **any**
error from that whole multi-module load - including a genuine
cross-module import error unrelated to grammar choice - `cmd_check`
discards it via `.or_else(|_| check_source_with_profile(&src,
&parser_profile))` and silently retries as a single-file, non-project
check of only the root source. For a multi-file project whose entry
point is not valid standalone Logos, this means `smc check
<project-root>` can silently validate only the entry file in isolation,
never surfacing the discarded project-level error and never checking
imported modules against it.

Found during the SSF09-E1 Diagnostic Authority Decision's Source
Surface Authority evidence pass
(`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`,
Decision A, path 1); not covered by #1670's filed scope. Repair must
follow that decision's fail-closed invariant: an authoritative
multi-module load failure must never be discarded in favor of a
narrower mechanism.

Related: #1580, #1670

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1918, opened 2026-09-12
docs(ssf09): freeze diagnostic authority decision (E1)

skulmakov-oss:
## Summary

SSF-09 umbrella #1580 remains OPEN.

ARCHITECTURE / CONTRACT DECISION ONLY
NO RUST IMPLEMENTATION
NO DIAGNOSTIC/CARRIER/SCHEMA CODE
NO LSP CODE
NO FORMATTER CHANGE
NO HARNESS/DEPENDENCY CHANGE
NO NEW CRATE CREATED
NO ISSUE CREATED
NO ISSUE LIFECYCLE TRANSITION

This PR freezes the SSF-09 Diagnostic Authority contract - the
prerequisite decision before any repair of #1670, #1697, #1698, #1699,
or #1704/E0243, and before any canonical carrier, versioned schema, or
LSP/editor code is written. It changes exactly two files: the decision
document, and a one-line, mandatory addition to `tests/legacy_guards.rs`'s
explicit TON618-content inventory (a CI-caught requirement, not a
content choice - see "History" below).

Went through nine owner-review correction rounds after the initial
freeze; each is summarized under "History." The version at current HEAD
states, honestly: Decisions A-D (source surface, diagnostic identity,
file identity, source range authority) are frozen; the canonical-carrier
*owner* is explicitly **not** selected - no existing crate satisfies the
combined dependency-graph, ownership-policy, and responsibility-cohesion
constraints (`ton618-core` is disqualified by a closed governance track,
not merely un-authorized), and a separate future governance/ownership
checkpoint must select and authorize the actual owner before any carrier
code is written.

## Changes

- **`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`**
  (new): freezes four authority questions with evidence gathered from
  this exact baseline (`4fa3331452c6730ab380c00b5c9ead35e1433ad2`, the
  SSF09-E0 governance-activation commit, [#1917](https://github.com/skulmakov-oss/Semantic/pull/1917)):
  - **Source Surface Authority** - `sm_front::CompileProfile`
    (`Auto`/`RustLike`/`Logos`) is named the canonical surface selector
    for every entry point going forward; `sm_profile::ParserProfile`
    stays a separate, surface-independent admission/policy
    configuration. Three independent current dispatch paths are
    documented (none of them a single "sole" authority), and a
    classification law is frozen with three outcomes - NO SURFACE
    CLAIM (a bare probe rejection with no prior positive evidence,
[...]

Issue #1670, opened 2026-08-15
FA-03-001: sm-sema surface dispatch swallows Logos parse/policy errors and retries as Rust-like source


skulmakov-oss:
Umbrella: #1617
Module: **03 — `sm-sema`**
Audit HEAD: `4ac4d1a169902777e760e23eaee92aa233edc04d`
Severity: **P2**
Classification: **FAIL-OPEN · SEMANTIC-DRIFT · FALSE-READY**

## Finding

`check_source_with_profile()` probes Logos with:

```rust
if let Ok(logos) = parse_logos_program_with_profile(input, profile) {
    if logos.system.is_some() || !logos.entities.is_empty() || !logos.laws.is_empty() {
        return analyze_logos_program(&logos, input);
    }
}
```

Any Logos parser/policy error is discarded. The same text is then retried through `parse_program_with_profile()` as Rust-like source. An import-only Logos parse can also fall through because the surface discriminator only checks System/Entity/Law presence.

## Claim / confidence source

`docs/spec/logos.md` says one file belongs to one source surface for a tool invocation and that Logos profile rejection is a source-level policy violation. `docs/spec/diagnostics.md` distinguishes syntax from policy failures.

## Actual guarantee / gap

`sm-sema` does not preserve the first Logos failure as authoritative. A Logos-disabled profile or malformed Logos input can be reinterpreted under another grammar and surfaced with a different failure class/code, rather than preserving the owning surface decision.

## Cross-phase impact

Source-surface identity and diagnostic authority can change inside `sm-sema` after `sm-front` has already made a policy/parse decision.

## Evidence

- `crates/sm-sema/src/std_adapters.rs::check_source_with_profile`
- `docs/spec/logos.md`
- `docs/spec/diagnostics.md`

## Phase A note

Audit evidence only. **Do not repair during Phase A.**

PR #1921, opened 2026-09-12
docs(ssf09): correct shared Import surface classification

skulmakov-oss:
## Summary

SSF-09 umbrella #1580 remains OPEN. #1670 remains OPEN, untouched in
lifecycle terms. #1919/#1920 untouched.

DOCS-ONLY CONTRACT CORRECTION
NO PRODUCTION RUST
NO TESTS
NO DEPENDENCY CHANGE
NO HARNESS CHANGE
NO ISSUE LIFECYCLE CHANGE

This corrects a factual error in the merged SSF09-E1 decision
(`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`,
merged in [#1918](https://github.com/skulmakov-oss/Semantic/pull/1918)),
discovered during SSF09-E2 implementation reconnaissance for `#1670`
- before any production code was written for that checkpoint.

## What was wrong

Decision A's UNIQUE POSITIVE SURFACE CLAIM law claimed that any
successful parse of a Logos-recognized declaration form - explicitly
including import-only Logos source - by itself constitutes *unique*
Logos evidence. That's false specifically for `import`.

Verified directly against both parsers at the exact merged baseline:
`crates/sm-front/src/parser.rs:89-135` (`parse_program`) lists
`TokenKind::KwImport` as one of RustLike's own seven recognized
top-level forms, and `parse_import_decl` (`parser.rs:137-160`) accepts
exactly `Import <optional pub> <string-literal> <optional as/*/{}>` -
a shape a real Logos fixture already in this codebase satisfies
verbatim (`crates/sm-sema/src/std_adapters.rs:995`:
`Import "math::core.sm"`). A bare top-level `Import "a.sm"` and nothing
else therefore parses to completion, successfully, under **both**
`parse_logos_program_with_profile` (under `CompatibilityMode::LegacySupport`,
which is exactly what `cli_profile()` sets via `ParserProfile::foundation_default()`)
and `parse_program_with_profile`. `KwImport` is the only keyword shared
between the two grammars - confirmed `KwSystem`/`KwEntity`/`KwLaw`/
`KwPulse`/`KwProfile` never appear in RustLike's parser, and RustLike's
own top-level keywords never appear in Logos's.

## The correction (final, after three owner-review rounds)

A declaration/token being *valid* Logos syntax does not by itself imply
it is *unique* Logos evidence:
- Logos-exclusive evidence - `system`/`entity`/`law`, **or `pulse`/
  `profile` on their own, no combination needed** (neither ever appears
  in RustLike's grammar) - with no RustLike evidence -> UNIQUE POSITIVE
  LOGOS CLAIM.
- RustLike-exclusive evidence with no Logos evidence -> UNIQUE POSITIVE
  RUSTLIKE CLAIM.
- `import` specifically is *shared vocabulary*, not automatically
  shared evidence: whether a given `import`-led input is ambiguous
  depends on whether that *specific* content is admitted by both
  grammars, not merely on which keyword introduces it (RustLike's
  `parse_import_decl` requires a string-literal token; Logos's handling
  has no equivalent requirement). The confirmed bare source
  `Import "a.sm"` independently admits both grammars to completion ->
  AMBIGUOUS/CONFLICTING CLAIMS for that concrete input - this is not
  asserted as a blanket rule for every `import`-led source.

`#1670`'s own filed observation - that the current discriminator can't
see `import`/`pulse`/`profile` content at all - remains a real, valid
defect. Only the *previously expected correct classification* for the
confirmed `Import "a.sm"` case was wrong: it isn't "recognize it as
Logos," it's "recognize the conflict and report it," since that exact
content is independently valid RustLike too.

This concerns the implicit `Auto` classification path only. By the
frozen INVARIANT, explicit caller authority (`CompileProfile::Logos`/
`RustLike`) resolves the surface without any `Auto` ambiguity question
- conceptually. Current production code does not yet satisfy this for
path 3 (`compile_program_to_ir_with_options_and_profile` still
unconditionally invokes the Logos parser under explicit `RustLike`);
that gap remains tracked separately by `#1920`, which is OPEN.

## Changes

- Decision A's current-state evidence (path 2), the corrected UNIQUE
  POSITIVE SURFACE CLAIM law itself, the AMBIGUOUS/CONFLICTING CLAIMS
  definition (now naming the confirmed concrete instance), the
  INVARIANT's shared-discriminator note, implementation consequence,
  test consequence, and the Exit gate's public-claim check.
- Decisions B, C, D untouched.
- `#1670`, `#1919`, `#1920` issue bodies untouched - not edited in this
  checkpoint per the correction's own boundary.

## Test plan

- [x] `cargo test --test ssf_status_drift` - 3/3 pass, unmodified
- [x] `cargo test --test legacy_guards` - 10/10 pass, unmodified
- [x] `cargo test --workspace --all-targets` - 4,677 passed, 0 failed
- [x] `cargo clippy --workspace --all-targets -- -D warnings` - clean
- [x] `cargo fmt --all --check` - known Windows path-length
  environmental limitation reproduced identically (`os error 206`);
  hosted `pr-ready` CI (ubuntu-latest) is the authoritative fmt gate
- [x] `git diff --check` - clean
- [x] Confirmed diff boundary is exactly the one decision doc; zero
  production Rust, test, dependency, or harness changes

## Round 1 correction (owner review)

The first draft over-generalized its own finding: it correctly
identified `KwImport` as the only keyword shared between the two
grammars, then several lines later described `import`/`pulse`/`profile`
as one "shared, ambiguous" category anyway - self-contradicting its own
evidence. `KwPulse`/`KwProfile` never appear in RustLike's parser at
all (confirmed: RustLike's `parse_program` recognizes exactly `Import`/
`enum`/`fn`/`record`/`schema`/`trait`/`impl`), so a bare `Pulse`/
`Profile` directive has no RustLike-side counterpart to conflict with -
it's already unique Logos evidence on its own, no combination with
`system`/`entity`/`law` required. Corrected the law, the AMBIGUOUS/
CONFLICTING CLAIMS definition, implementation/test consequence (added
distinct regressions for bare `Pulse`/`Profile` alone and for `import`
combined with genuinely exclusive evidence), and the public-claim
check.

Also corrected: the explicit-`CompileProfile` paragraph described
"explicit RustLike/Logos already bypasses classification" as
present-tense fact, when path 3 (`compile_program_to_ir_with_options_and_profile`,
tracked as `#1920`) still unconditionally invokes the Logos parser
under explicit `RustLike`. Reworded as a normative INVARIANT
requirement, with an explicit note that current code doesn't yet
satisfy it there.

Also fixed bookkeeping: the Durable Tracking section still said the two
newly-discovered defects were "not created" candidate issues; both were
actually filed post-merge as `#1919` and `#1920` (both OPEN). Updated
the record to state this - no issue body/state touched.

## Round 2 correction (owner review)

Round 1's fix over-corrected in the opposite direction: it stated that
`import`, "when it is the only positive evidence," is unconditionally
AMBIGUOUS/CONFLICTING - stronger than the grammars actually prove.
Verified precisely: RustLike's `parse_import_decl` requires a
string-literal token specifically (`expect_string_literal_text`, which
errors if the next token isn't `TokenKind::String`); Logos's `KwImport`
handling has no equivalent content requirement at all. So an
`import`-led line whose content does *not* satisfy RustLike's structured
shape (e.g. an unquoted path) is accepted by Logos but rejected by
RustLike - that's not ambiguity, it's UNIQUE POSITIVE LOGOS CLAIM
(RustLike produced no claim to conflict with). Corrected the law,
the AMBIGUOUS/CONFLICTING CLAIMS definition, and the public-claim check
to state ambiguity is about the **concrete input's actual admissibility
under both grammars**, not about which keyword introduces it - keeping
`Import "a.sm"` (a quoted string, satisfying both) as the one confirmed
concrete instance, without generalizing to every `import`-led source.

Also fixed a residual `import`/`pulse`/`profile` collapse the INVARIANT
section's shared-discriminator note still had (re-introducing the
already-rejected "combined with exclusive evidence" requirement for
`pulse`/`profile`), and updated the FROZEN/NOT-FROZEN framing: this
decision now explicitly freezes `pulse`/`profile`'s evidence status
(Logos-exclusive, unconditionally) alongside the semantic classification
law itself, while leaving the concrete classifier implementation
(Rust type, scanning strategy, evaluation mechanics) unfrozen - and
explicitly states it does not freeze a first-token classifier as the
implementation strategy.

## Round 3 correction (owner review)

The main blocker: the "Qualification requirements" section still
required that "`#1670`'s fix does not change any admitted program's
compilation result - only failure-classification behavior." That's no
longer satisfiable alongside this document's own corrected law - the
confirmed `Import "a.sm"` collision is *currently admitted* (silently
succeeds today via `check_source_with_profile`'s RustLike fallback),
and the corrected law explicitly requires it to become a deterministic
classification error under implicit `Auto`. Restated precisely: inputs
with exactly one authoritative surface must preserve their existing
result; inputs currently admitted only because today's fail-open
dispatch silently resolves a real authority conflict are explicitly not
covered by that guarantee; `Import "a.sm"` is the concrete instance
expected to change, and that change is not a regression.

Also fixed two residual wording defects that predated rounds 1/2's
sweeps: Decision A's current-state evidence still called `pulse`/
`profile` "shared, not unique, evidence" (wrong - they're unconditionally
Logos-exclusive) and still described the repaired outcome for "a bare
import-only file" in universal terms round 2's own concrete-source law
already rejects. Rewrote to distinguish `pulse`/`profile`'s outcome
(recognize as Logos, unconditionally) from `import`'s outcome (depends
on content; `Import "a.sm"` is the confirmed dual-admissible instance),
and tightened the classification law's intro and implementation
consequence to name the confirmed fixture instead of a generic category.

**Do not merge. Wait for owner review and a separate implementation GO
before resuming SSF09-E2 / `#1670`.**

### DIFF
--- a/docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md
+++ b/docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md
@@ -106,22 +106,36 @@ is rewritten below from direct evidence.
    mode via `require_legacy_compatibility` and is consumed without
    error) but its content is discarded - the loop skips to the next
    newline without populating `out.system`/`out.entities`/`out.laws`.
-   A source file containing **only** import directives therefore parses
-   as `Ok(LogosProgram { system: None, entities: vec![], laws: vec![] })`
+   A source file containing **only** `import`/`pulse`/`profile`
+   directives therefore parses as
+   `Ok(LogosProgram { system: None, entities: vec![], laws: vec![] })`
    - a **successful** Logos parse that the `system.is_some() ||
    !entities.is_empty() || !laws.is_empty()` discriminator cannot see at
    all, so it silently falls through to RustLike without ever
-   representing that this input was also valid Logos syntax. This is a
-   distinct failure mode from the swallowed-`Err` case above: the parse
-   here does not fail at all, it succeeds and is invisible to the
-   discriminator. **Correction (E1A)**: the discriminator's blindness to
-   `import`/`pulse`/`profile` content is real and is the defect - but
-   the correct repaired outcome for a *bare* import-only file is not
-   "recognize it as Logos" (see the corrected UNIQUE POSITIVE SURFACE
-   CLAIM law below - `import` alone is shared, not unique, evidence);
-   it is recognizing that the input independently satisfies both
-   grammars and reporting that conflict explicitly, rather than
-   resolving it silently toward either one.
+   representing that this input was also recognized Logos syntax. This
+   is a distinct failure mode from the swallowed-`Err` case above: the
+   parse here does not fail at all, it succeeds and is invisible to the
+   discriminator. **Correction (E1A, round 3)**: the discriminator's
+   blindness to `import`/`pulse`/`profile` content is real and is the
+   defect for both - but its correct repaired outcome differs by which
+   keyword is present, and round 1/2's own fixes to this same document
+   had already established that difference elsewhere without this
+   paragraph being swept to match. For a bare `pulse`-only or
+   `profile`-only file: the correct repaired outcome **is** "recognize
+   it as Logos" - these keywords are Logos-exclusive (see the corrected
+   UNIQUE POSITIVE SURFACE CLAIM law below), so the fix is exactly
+   surfacing what the discriminator currently discards. For a bare
+   `import`-only file: `import` is *shared vocabulary*, so the correct
+   repaired outcome depends on the concrete content, not on "recognize
+   it as Logos" - the confirmed dual-admissible instance is specifically
+   `Import "a.sm"` (a quoted string, satisfying both grammars' actual
+   requirements to completion), which must be reported as an explicit
+   conflict rather than resolved silently toward either grammar; an
+   import-only file whose content RustLike's own `parse_import_decl`
+   would reject (e.g. an unquoted path) has no RustLike claim to
+   conflict with and is UNIQUE POSITIVE LOGOS CLAIM instead. This
+   paragraph does not generalize "bare import-only" to mean "always
+   ambiguous" - see the corrected law below for the precise rule.
 3. `smc compile`/`dump-ir`/`hash-ir`/`hash-smc` route through
    `sm_ir::legacy_lowering::compile_program_to_ir_with_options_and_profile`
    (`crates/sm-ir/src/legacy_lowering.rs:1176-1220`), which **does**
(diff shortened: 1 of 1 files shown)


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1922, opened 2026-09-13
docs(ssf09): correct Import "a.sm" baseline-result description (E1B)

skulmakov-oss:
## Summary

Docs-only correction to the SSF09 Diagnostic Authority Decision
(`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`),
merged in [#1918](https://github.com/skulmakov-oss/Semantic/pull/1918)
and corrected in [#1921](https://github.com/skulmakov-oss/Semantic/pull/1921)
(E1A).

Discovered during SSF09-E2 implementation for `#1670`, during the
TDD baseline run, before the production fix's qualification proceeded
further: the E1A-merged text describes the confirmed `Import "a.sm"`
collision as **"currently admitted"** and claims it **"silently
succeeds today via `check_source_with_profile`'s RustLike fallback."**

Re-verified directly against `check_source_with_profile` on the E1A
merge SHA (`262adc00369e23fb1ba92e5f093656041e983aa9`), in a disposable
temporary test (added, run, then reverted - no production or test file
is changed by this PR):

```
LOGOS PARSE   = Ok
RUSTLIKE PARSE = Ok
FINAL check_source_with_profile RESULT = Err("E0201")
```

Today's discriminator cannot see import-only Logos evidence
(`system`/`entities`/`laws` all empty), so it silently falls through to
RustLike; RustLike's own parse also succeeds, but the subsequent
`type_check_program` call rejects the program with `E0201` ("program
must define fn main()"), since a bare `Import "a.sm"` file has no
`main`. Today's concrete result is a **RustLike-side terminal failure**,
not a silent success.

## What does NOT change

The authority law itself is unaffected by this correction:

- `Import "a.sm"` remains the confirmed, concrete dual-admissible
  instance - independently valid under both the Logos and RustLike
  grammars.
- Under implicit `Auto`, that dual-admissibility is still required to
  be classified as a deterministic AMBIGUOUS/CONFLICTING error.
- `#1670` is still required to replace today's result with that
  deterministic classification error - only the *description* of
  today's concrete result changes, from "silent success" to "wrong-
  authority terminal failure (`E0201`, missing `fn main`)."
- Pulse/Profile unconditional Logos-exclusivity, the no-first-token-
  classifier rule, unique-surface result preservation, and `#1919`/
  `#1920` independence are all unchanged.

## Changed files

Exactly one file:
`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`

[...]

Issue #1580, opened 2026-08-09
SSF-09 — Diagnostics and editor baseline


skulmakov-oss:
Parent: #1569
Depends on: #1579
Related roadmap: #1375

Priority: P9.

Status gate: **BLOCKED until SSF-08 closes.** Editor tooling consumes canonical Semantic diagnostics; it must not become a second parser/typechecker/verifier authority.

## Goal
Provide stable machine-readable diagnostics, formatter behavior, and a minimal external editor/language-server baseline derived from canonical Semantic tooling.

## Safe execution steps
1. Read SSF-00…08 outputs and #1375; inventory existing diagnostic structures, spans, formatter behavior, CLI JSON/text outputs, symbol data, and editor-related experiments.
2. Freeze a versioned machine-readable diagnostic schema: code, severity, file, range/span, message, optional structured notes/fixes.
3. Define source-span behavior across parser, semantic analysis, lowering, verifier-facing source mapping, project/module loading, and manifest diagnostics.
4. Freeze formatter contract and idempotence rules before wiring editor formatting.
5. Define the language-server authority boundary: `smc check`/canonical compiler services remain source of truth; server may orchestrate/cache but not reinterpret language semantics.
6. Implement diagnostics-only LSP baseline first, with project-root awareness and deterministic cancellation/staleness behavior.
7. Add hover/go-to-definition/document-symbol support only where canonical symbol data already exists; do not build a parallel symbol engine.
8. Add formatting bridge using the canonical formatter.
9. Add structured code actions only if they can be represented as proposals derived from canonical diagnostics; no silent source mutation.
10. Sanitize external command/output handling before rich rendering; prohibit hidden telemetry/source upload.
11. Add protocol fixtures for malformed requests, stale documents, invalid UTF-8/positions where applicable, project-root changes, and CLI/LSP diagnostic parity.
12. Document and qualify at least one editor integration or a protocol-generic installation path.
13. Run formatter/diagnostic/LSP fixtures, project tests, and broader regression gates for touched compiler/tooling surfaces.

## Acceptance criteria
- [ ] Machine-readable diagnostics have a versioned stable schema.
- [ ] Diagnostics preserve code, severity, file and range where source information exists.
- [ ] Formatter is deterministic and idempotent.
- [ ] LSP/editor results derive from canonical tooling and cannot silently disagree with CLI truth.
- [ ] Project-root awareness works.
- [ ] No hidden telemetry or source upload exists.
[...]

PR #1923, opened 2026-09-13
fix(ssf09): make sm-sema surface dispatch fail closed (#1670)

skulmakov-oss:
## Summary

Repairs `sm_sema::check_source_with_profile()`'s fail-open source-surface
dispatch, per the frozen Source Surface Authority law in
`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`
(Decision A, as corrected by SSF09-E1A/[E1B](https://github.com/skulmakov-oss/Semantic/pull/1922)).

This addresses `#1670`'s own filed observation, referenced from that
SSF09-09 tracking issue. Public signature unchanged
(`pub fn check_source_with_profile(input: &str, profile: &ParserProfile)
-> Result<SemanticReport, SemanticError>`). No `CompileProfile` plumbing
added in this checkpoint. `#1919`/`#1920`'s independent scope (smc-cli
fallback, sm-ir `Auto` probe) is untouched.

## What changed

A private three-outcome classifier inside `sm-sema`
(`crates/sm-sema/src/std_adapters.rs`), driven by top-level token-kind
evidence only (never a scan of arbitrary tokens anywhere in the stream):

- `System`/`Entity`/`Law`/`Pulse`/`Profile` are unconditionally
  Logos-exclusive.
- `enum`/`fn`/`record`/`schema`/`trait`/`impl` (and role-marked schema
  declarations) are unconditionally RustLike-exclusive.
- `Import` - the one keyword both grammars recognize - is never treated
  as evidence by itself; ownership is derived from which grammar(s)
  actually admit the *concrete* input to completion.

Dispatch law:

- Unique exclusive evidence on one side → that surface is authoritative
  (its own success **or failure** is final, never replaced by a retry).
- No exclusive evidence and the input is blank → defers to RustLike
  (unchanged existing behavior).
- Otherwise both parsers are consulted on the concrete input directly:
  Logos-only success → unique Logos; RustLike-only success → unique
  RustLike; **both succeeding → deterministic `AMBIGUOUS/CONFLICTING`
  error** (new); neither succeeding → today's existing terminal
  RustLike-attributed error is preserved (no ambiguity manufactured).

## The defect

Previously, a genuine Logos parse/policy failure was silently
discarded (`if let Ok(logos) = ...`) and execution fell through to
RustLike regardless of the failure's cause. Separately, an
import-only Logos parse (`system`/`entities`/`laws` all empty) was
invisible to the old discriminator, causing the same silent-fallthrough
for shared vocabulary. The confirmed, concrete dual-admissible
collision `Import "a.sm"` (SSF09-E1A) reached RustLike this way and
terminated with `E0201` ("program must define fn main()") - a
wrong-authority terminal failure, not the deterministic classification
error the corrected law requires.

## Test-first (T1-T9)

Written and confirmed **failing against baseline** before any
production change:

| Test | Proves |
|---|---|
| T1 `check_source_preserves_authoritative_malformed_logos_failure` | A genuine Logos syntax failure with exclusive evidence present is never reinterpreted as RustLike |
| T2 `check_source_preserves_logos_policy_rejection` | Same, for a policy rejection (`allow_logos_surface = false`) |
| T3/T9 `check_source_reports_confirmed_import_collision_as_ambiguous` | `Import "a.sm"` → deterministic ambiguity, not silent success |
| T4 `check_source_does_not_treat_shared_keyword_alone_as_ambiguous` | Unquoted `Import a.sm` (RustLike rejects, no string literal) → unique Logos, not ambiguity |
| T5 `check_source_treats_bare_{pulse,profile}_as_unique_logos_evidence` | Bare Pulse/Profile is unconditional Logos-exclusive evidence, with a second assertion proving no RustLike fallback occurs under a Logos-policy-disabled profile |
| T6 `check_source_preserves_ordinary_rustlike_program` | Ordinary RustLike program still admitted |
| T7 `check_source_preserves_existing_valid_logos_program` | Existing valid Logos program still admitted |
| T8 `check_source_does_not_manufacture_ambiguity_for_blank_input` | Blank input never reported as a conflict |

## Mutation evidence (M1-M4)

Each mutation applied one at a time, confirmed to break the specific
test(s) it targets, then fully reverted (never committed) - final
working-tree fingerprint (`git diff | git hash-object --stdin`)
confirmed identical before and after all four:

- **M1** - restored the old fail-open dispatch → T1, T2 fail (Logos
  error silently swallowed, misattributed as a RustLike parse error).
- **M2** - first-token "any `Import`-led input is ambiguous" heuristic
  → T4 fails. This is the direct proof that *shared vocabulary ≠
  concrete-source ambiguity*.
- **M3** - first-positive-wins on a genuine dual-admissibility conflict
  → T3/T9 fails (silently succeeds instead of reporting the conflict).
- **M4** - dropped Pulse/Profile from exclusive evidence → T5 fails.
  (Note: the *original* T5 assertions, which only checked `is_ok()`,
  did not catch this mutation, since the generic dual-parse fallback
  path coincidentally also succeeds for a bare `Pulse`/`Profile` input.
  T5 was strengthened with a second assertion, under a Logos-policy-
  disabled profile, that distinguishes "reached via the Logos-exclusive
  authoritative path" from "reached via the fallback" - this is
  included in this PR's diff.)

## Qualification

- `cargo test -p sm-sema` - 53/53 passed
- `cargo test --test ssf_status_drift` - 3/3 passed
- `cargo test --test legacy_guards` - 10/10 passed
- `cargo test --workspace --all-targets` - 301 test binaries, 4686
  tests, 0 failed (full untruncated log verified, not just tail output)
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo fmt --all --check` - hit the known Windows path-length
  limitation (`os error 206`), identical class to the previously
  established environmental limitation, not a code issue
- `git diff --check` - clean
- Scope verified: exactly `crates/sm-sema/src/lib.rs` and
  `crates/sm-sema/src/std_adapters.rs` changed; no other crate touched

## Scope note

Does not change any admitted program's semantic result for
single-authoritative-surface inputs. The confirmed `Import "a.sm"`
collision's result is *intentionally* changed, from today's
wrong-authority RustLike terminal failure (`E0201`) to a deterministic
`AMBIGUOUS/CONFLICTING` classification error - this is a required
acceptance criterion of the corrected authority law, not a regression.

`#1670`, `#1919`, `#1920`, `#1580` remain OPEN.

**Do not merge** - awaiting hosted CI, fresh Codex review, and owner
review.


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
## Correction round 1 applied (HEAD `226a5c15`)

Owner review of the exact-HEAD code (independent of the mutation/test evidence, which was accepted) found the classifier violated this PR's own stated invariant - it scanned every token in the file for exclusive keywords, not just top-level declaration-head positions, since a keyword's token kind is assigned by the lexer from its text alone regardless of syntactic position (e.g. `Entity` inside a RustLike `fn` body still lexes as `KwEntity`).

Fixed in commit `226a5c15`:
- New `top_level_token_kinds` helper: depth-tracking filter over already-public structural tokens (brace/paren/bracket + the lexer's own Indent/Dedent). No sm-front grammar duplicated.
- `has_logos_exclusive_evidence`/`has_rustlike_exclusive_evidence` now consult only top-level-position tokens.
- Lexer failures (`lex(input).unwrap_or_default()`) no longer silently become "blank input" - propagated as a deterministic error instead.
- Ambiguity message reworded to a generically-true "conflicting source-surface claims," since the old wording ("independently valid under both grammars") was false for the dual-exclusive-evidence conflict class.
- Stale T3/T9 comment referencing a pre-E1B "silent success" baseline corrected to the accurate wrong-authority `E0201` description.

New regressions T10/T11 (nested-keyword-must-not-claim-opposite-surface, both directions, fixtures verified against the real parsers first) plus a lexer-failure-preservation test, all confirmed failing pre-fix and passing post-fix. Mutation M5 (revert to whole-stream scan) confirmed T10/T11 both fail, then fully restored - fingerprint identical before/after. M1-M4 re-verified against the corrected code, all still catch their regressions.

Full requalification re-run on this HEAD: `cargo test -p sm-sema` 56/56, workspace 4689/4689, clippy clean, `cargo fmt -p sm-sema --check` clean, `git diff --check` clean. Hosted CI 22/22 internal checks green; `Workers Builds: semantic` (external Cloudflare deploy check, unrelated to this diff) remains red - not part of this PR's qualification scope.

Still **do not merge** - awaiting owner review of this round.

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
## Round 4 reverted — STOP, structural finding

Commit `7bae555c` (round 4: "stop treating bare newlines as declaration boundaries") has been **reverted** in `60c1e44c`, back to round 3's exact state (`300c9fae`).

### Why

Round 4 fixed a real bug (a bare `Newline` at depth 0 wrongly re-armed the declaration-head boundary, letting a keyword in a still-open header - e.g. `fn\nEntity() {}` - be double-counted). The fix removed `Newline` as a boundary trigger entirely, keeping only `RBrace`/`Dedent` returning to depth 0.

A fresh independent adversarial review (Antigravity) found that this fix is a **worse regression** than what it fixed. Confirmed empirically against the committed code:

- `Pulse "tick"\nfn main() {\n    return;\n}\n` - genuine dual evidence (`Pulse` = Logos-exclusive, `fn` = RustLike-exclusive) - silently misclassified as **Logos-only**, `fn` never detected, result: `"expected Logos declaration"` instead of the required ambiguity error.
- `Import "a.sm"\nEntity A:\n    state x: quad\n` under `allow_logos_surface = false` - genuine Logos-exclusive evidence (`Entity`) masked by a preceding `Import`, silently misclassified as **no evidence at all**, result: a RustLike parse error instead of the required Logos policy-rejection (breaks the same invariant class as T2).

Root cause: `System`, `Pulse`, `Profile`, and a bare `Import` all have **no block body at all** (confirmed directly against `parse_logos_system`, `parse_logos_program`'s legacy-directive handling, and `parse_import_decl`) - none of them ever produce an `RBrace` or `Dedent`. Round 4's rule silently disables all further evidence detection after any one of these, for the rest of the file.

### The actual structural problem

Different declaration kinds terminate via genuinely different, non-uniform signals: `Newline` (Pulse/Profile's legacy skip), `Semi` (expression-bodied `fn`), `RBrace` (braced bodies), `Dedent` (indented bodies), or **no signal at all** (bare `Import`, which can be immediately followed by another declaration on the same physical line with zero separator - `Import "a.sm" fn main() { return; }`, found in the same review round and left unfixed for the same reason).

No single lexical rule over token kinds can track "has the current declaration finished" without knowing *which kind* of declaration is currently open - which requires the kind of grammar-level knowledge this checkpoint's own brief explicitly prohibits duplicating in sm-sema.

### Current state

- HEAD `60c1e44c` is byte-identical to round 3 (`300c9fae`), which passed hosted CI 22/22 previously and is expected to again.
- Round 3's own known residual gaps (multi-line declaration headers, e.g. `fn\nEntity()`, and same-line multiple declarations) remain present and undocumented as a fix - they are narrower, less common patterns than round 4's regression, which is why this is the safer state to sit in while this is resolved.
- No further implementation attempted pending owner input on the underlying architecture question (see full report in chat).

**Do not merge.**

PR #1924, opened 2026-09-13
docs(ssf09): freeze Surface Admission Contract for sm-front (Decision E)

skulmakov-oss:
## Summary

Decision-only investigation, per an explicit narrow owner GO after PR
[#1923](https://github.com/skulmakov-oss/Semantic/pull/1923)'s four
correction rounds each fixed one lexical-heuristic hole in `sm-sema`'s
own token-stream reconstruction of "which grammar owns this
declaration," and each was subsequently found - by independent
adversarial review, not by CI or the existing test suite, both of
which stayed green every round - to have introduced or left a
*different* hole. Round 4 was reverted
([60c1e44c](https://github.com/skulmakov-oss/Semantic/commit/60c1e44c668825fc6cd621b6d5b6c6242f533819))
after being shown to regress worse than what it fixed.

**Root cause**: different declaration kinds terminate via genuinely
different lexical signals (`Newline`, `Semi`, `RBrace`, `Dedent`, or
none at all), so no uniform token-kind/depth rule can determine "has
this declaration finished" without knowing *which kind* is open - the
exact grammar-level knowledge `#1670`'s own brief prohibits `sm-sema`
from duplicating. This is documented fully, with all four rounds'
concrete counterexamples, in the "Decision E" section.

## What this proposes (NOT frozen, NOT implemented)

Two new, purely additive `sm-front` functions
(`admit_program_with_profile`, `admit_logos_program_with_profile`)
returning a new `GrammarAdmission<T>` type instead of today's flat
`Result`, so `sm-sema` can ask each real parser directly what it claims
instead of re-deriving the answer from raw tokens.

A workspace census (included in the doc) confirms this additive shape
is not merely convenient but the *only* non-breaking option: ~270+
existing call sites across every crate depending on `sm-front`
(directly or transitively: `smc-cli`, `sm-sema`, `sm-ir`, `sm-emit`,
`sm-vm`, `sm-verify`, `prom-runtime`, two examples) are all written
against `Result<T, FrontendError>` syntax incompatible with a bespoke
enum - an additive function pair touches zero of them.

## Frozen by owner ruling (four review rounds)

- **Naming**: `GrammarAdmission<T>` (not `SurfaceAdmission<T>`), kept
  lexically separate from cross-grammar `SurfaceAuthority`/
  `SurfaceVerdict` coordinator vocabulary.
- **Two independent axes, not one flat result** (the core correction,
  round 4): a grammar-local admission reports **evidence basis**
  (`NoClaim` / `Shared` / `Exclusive` - did this grammar establish
  ownership only through vocabulary shared with the other grammar, or
  through vocabulary exclusive to it) and **parse outcome** (`Ok`/`Err`)
  as two separate facts, not a single three-state result. `Exclusive`
  is absorbing: once a grammar sees its own exclusive evidence anywhere
  in the input, an earlier or later `Import` line never weakens it.
  This is not a new invention - it promotes Decision A's own
  already-frozen shared-vs-exclusive distinction from prose into the
  type.
- **Two-stage cross-grammar resolver**: evidence strength is compared
  first (`Exclusive` beats `Shared` beats `NoClaim`; `Exclusive` vs
  `Exclusive` is always AMBIGUOUS/CONFLICTING regardless of outcome),
  and only a `Shared`-vs-`Shared` tie falls through to a second stage
  resolved by parse outcome (`Ok` beats `Err`; `Ok`+`Ok` and `Err`+`Err`
  are AMBIGUOUS/CONFLICTING). A grammar-local `Err` becomes an
  *authoritative* failure only after that grammar has won ownership
  resolution - never grounds by itself to override a stronger claim,
  and never itself enough to let a weaker `Shared(Ok)` silently beat a
  stronger `Exclusive(Err)` (which would reintroduce the fail-open
  pattern `#1670` exists to remove).
- Verified against six frozen concrete examples (bare quoted/unquoted
  `Import`, `Import` alongside each grammar's exclusive evidence,
  exclusive-only malformed input for each grammar) plus two additional
  validating traces, all of which broke under at least one of rounds
  1-3's models and all resolve correctly under this one - with **zero
  `Import`-specific logic anywhere in `sm-sema`**.

Rounds 1-3 each proposed and then had adversarial review disprove a
simpler, flat three-state model (`NoClaim`/`Accepted(T)`/
`ClaimedButFailed(FrontendError)`, computed via progressively refined
"claim at dispatch" rules). Round 1's own 3x3 matrix over that flat
model is preserved in the document as a collapsed, clearly-marked audit
note - superseded, not deleted, and not implementable, since the
evidence-strength dimension it's missing cannot be recovered by adding
rows to it.

## Scope

- **No implementation.** `sm-front`, `sm-sema`, and every other
  production file are untouched - exactly one file changed
  (`docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md`).
- Does not reopen or reinterpret Decisions A-D (frozen via
  [#1918](https://github.com/skulmakov-oss/Semantic/pull/1918)/[#1921](https://github.com/skulmakov-oss/Semantic/pull/1921)/[#1922](https://github.com/skulmakov-oss/Semantic/pull/1922)) - purely additive new section.
- `#1670`/PR #1923 stays parked at `60c1e44c` throughout; `#1580`
  stays OPEN; no issue-lifecycle changes.

## Qualification

- `cargo test --test ssf_status_drift` - 3/3 passed
- `cargo test --test legacy_guards` - 10/10 passed
- `git diff --check` - clean

**Do not merge** - awaiting explicit owner merge GO.


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1925, opened 2026-09-13
feat(sm-front): implement GrammarAdmission per frozen SSF-09 Decision E

skulmakov-oss:
## Summary

Owner GO: sm-front implementation checkpoint, Stage 1 of the `#1670` repair. Implements exactly the architecture frozen in [Decision E - Surface Admission Contract](https://github.com/skulmakov-oss/Semantic/blob/main/docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md), including the round-5 precision amendment ([#1926](https://github.com/skulmakov-oss/Semantic/pull/1926), merged) discovered midway through this same implementation - no more, no less.

Adds `GrammarAdmission<T>` and two new, purely additive public functions - `admit_program_with_profile`, `admit_logos_program_with_profile` - so a future `sm-sema` rewrite can ask each real parser directly what it claims about an input instead of re-deriving it from raw tokens.

## What this implements

`GrammarAdmission<T>` reports two independent facts per the frozen contract:
- **evidence basis**: `NoClaim` / `Shared` / `Exclusive` - did this grammar establish ownership only through vocabulary shared with the other grammar (`Import`, the only current member), or through vocabulary exclusive to it. `Exclusive` is absorbing and monotonic.
- **parse outcome**: `Ok`/`Err`, nested inside `Shared`/`Exclusive`.

Both `admit_*_with_profile` functions take `(source: &str, tokens: &[Token], profile: &ParserProfile)` - `source` is diagnostic context only (round 5): never re-lexed, never a factor in classification, used solely so the nested `FrontendError` carries the same caret-diagnostic source line the existing `parse_*_with_profile` functions already produce.

`admit_program_with_profile` mirrors `parse_program`'s exact existing abort-on-first-error loop. `admit_logos_program_with_profile` mirrors `parse_logos_program`'s exact existing accumulate-and-recover loop. **Neither loop's scan mechanics are touched or aligned with the other** - this asymmetry is explicitly required by the frozen decision, not incidental.

Policy handling follows the frozen round-5 finalization law: a legacy-compatibility policy violation is tracked **separately** from ordinary syntax errors, never merged with them, and returned as the grammar's outcome verbatim even when the evidence-basis scan continues past it to discover stronger (`Exclusive`) evidence afterward. The global `require_logos_surface` gate outranks a legacy-compatibility failure. Only when neither policy failed does the ordinary syntax path run, reproducing `parse_logos_program`'s exact existing `merge_logos_errors` aggregation with no special case.

## Development history on this PR

An earlier revision of this PR (round 4) used a 2-argument signature and an ad hoc shortcut to preserve `FrontendErrorKind::PolicyViolation` classification. An owner-directed investigation found its errors didn't match `parse_logos_with_profile`'s errors for the same input - which led to the separate round-5 amendment (#1926) to Decision E itself, frozen and merged first. This PR was then rebased onto that amendment and reimplemented to match it exactly, including the exact-`FrontendError`-equality tests the amendment's own migration plan requires.

A subsequent adversarial review round also caught and disproved a claim in an earlier commit here - that evaluating the whole-surface policy gate at scan-start vs. after the scan was "unobservable" for any input. It isn't: a leading comment (which the lexer consumes without emitting a token) breaks the coincidence that made two earlier, narrower test fixtures agree. Fixed and covered by a permanent regression.

## Hard boundaries respected (owner GO)

- `sm-sema` untouched.
- `#1670`/PR #1923 untouched, still parked at `60c1e44c`, `#1670` stays OPEN.
- `CompileProfile`, `ParserProfile`, `FrontendError`'s own structure unchanged.
- Every existing public sm-front function (`parse_program(_with_profile)`, `parse_logos_program(_with_profile)`, `parse_rustlike(_with_profile)`, `parse_logos(_with_profile)`, `lex`) keeps its exact current signature and behavior - purely additive.
- No new classifier logic anywhere outside `sm-front`.
- RustLike abort-first / Logos recover-and-continue mechanics preserved as-is, not aligned.
- Policy gates never erase established evidence basis, and never leak into `NoClaim`.

## Testing

33 unit tests in `crates/sm-front/src/parser.rs` (`mod grammar_admission_tests`), zero `#[ignore]`d, zero probe-only assertions:
- Reachability of every `GrammarAdmission` variant/outcome combination, both grammars.
- All six frozen concrete examples.
- Absorption of `Exclusive` over `Shared`, both declaration orderings.
- Five exact-`FrontendError`-equality parity tests against `parse_logos_with_profile` (ordinary single/multiple syntax failures, legacy-policy failure preserved across later exclusive evidence in a clean and a mixed-with-syntax-error shape, global surface denial).
- A dedicated precedence test for both policy gates failing simultaneously (surface must outrank legacy-compatibility) - a real coverage gap found during mutation testing, not previously exercised.
- The RustLike/Logos scan-mechanics asymmetry itself, as an explicit regression.

Mutation-tested across both implementation passes: shared-threshold bypass, policy-masking-at-`NoClaim`, non-monotonic promotion, hard-abort reintroduction, always-merge, promote-on-success-instead-of-dispatch, reverting the separated policy-error tracking, swapping gate precedence, and the scan-start-vs-post-scan pos-capture timing - each confirmed caught by the test suite (or, in one case, found to need a better fixture before it could be caught), each cleanly reverted to an identical diff fingerprint before the next.

## Qualification

- `cargo test -p sm-front` - 677 passed, 0 ignored
- `cargo clippy -p sm-front --all-targets -- -D warnings` - clean
- `cargo fmt -p sm-front --check` - clean
- `cargo test --test ssf_status_drift` - 3/3
- `cargo test --test legacy_guards` - 10/10
- `cargo test --test dependency_boundaries` - 3/3
- `cargo test --test ir_opt_boundaries` - 2/2
- `cargo test --test frontend_boundaries` - 2/2
- `cargo test --test public_api_contracts` - 88/88 (sm-front not covered by this guard's TARGETS list, confirmed by inspection)
- `cargo check --workspace --all-targets` - clean
- `cargo test --workspace --lib --bins` - 0 failures anywhere downstream

**Do not merge** - awaiting explicit owner merge GO.


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

PR #1926, opened 2026-09-13
docs(ssf09): amend Decision E - source-context param + policy finalization law

skulmakov-oss:
## Summary

Owner-ruled round-5 amendment to the already-frozen [Decision E - Surface Admission Contract](https://github.com/skulmakov-oss/Semantic/blob/main/docs/roadmap/stable_foundation/ssf09_diagnostic_authority_decision.md), discovered during PR [#1925](https://github.com/skulmakov-oss/Semantic/pull/1925)'s Stage 1 implementation. This is a **precision correction**, not a redesign - evidence strength, `Shared`/`Exclusive` semantics, the RustLike-abort-first/Logos-recover-and-continue scan asymmetry, and the cross-grammar evidence resolver are all unchanged.

## What was found

A regression comparing `admit_logos_program_with_profile`'s nested `FrontendError` against `parse_logos_with_profile`'s error for the identical source failed, exposing two independent gaps in round 4's frozen contract:

1. `admit_*` took only `tokens` and `profile` - no access to the original source text - so its `Parser` was built with `source: String::new()`. `Parser::error_at_token` reads `self.source` to embed a caret-diagnostic source line into `FrontendError.message`, so `admit_*`'s errors carried an empty line where `parse_*`'s carry the real one.
2. Round 4's text established policy as an *outcome* modifier but left the exact aggregation underspecified; the implementation filled the gap with an ad hoc `errors.len() == 1` shortcut that itself produced a message shape matching neither the old parser's always-wrapped single-error format nor a policy-preserving one.

## Ruling A - source-context parameter

Both `admit_*` signatures gain `source: &str`:

```rust
pub fn admit_program_with_profile(source: &str, tokens: &[Token], profile: &ParserProfile) -> GrammarAdmission<Program>;
pub fn admit_logos_program_with_profile(source: &str, tokens: &[Token], profile: &ParserProfile) -> GrammarAdmission<LogosProgram>;
```

`source` is diagnostic context only, governed by an exact-correspondence precondition (`tokens == lex(source)`) and three prohibitions: never re-lexed, never a factor in evidence classification, never a second source-surface authority. `tokens` alone remains the sole syntactic/admission scan input.

## Ruling B - policy-vs-syntax finalization law

A legacy-compatibility policy violation is tracked separately from the ordinary syntax `errors` vector, never routed through `merge_logos_errors`, and returned as the grammar's outcome verbatim even when the evidence-basis scan continues past it to discover stronger (`Exclusive`) evidence afterward - preserving the old parser's terminal-policy behavior for the *outcome* while still allowing Decision E's continued evidence discovery for the *basis*. The global `require_logos_surface` gate outranks a legacy-compatibility failure, matching the existing parser's own gate order. Only when no policy failure occurred does the ordinary syntax path run, reproducing `parse_logos_program`'s exact existing aggregation with no length-based special case.

## Scope

- Docs-only. `sm-front` is not touched by this revision.
- PR #1923 and `sm-sema` untouched.
- **PR #1925 is deliberately left unmodified** pending this amendment landing on `main`, per explicit owner instruction - it will be rebased onto this correction once merged.

## Qualification

- `cargo test --test ssf_status_drift` - 3/3
- `cargo test --test legacy_guards` - 10/10
- `git diff --check` - clean

**Do not merge** - awaiting fresh Codex review, an Antigravity adversarial review scoped to (1) whether the widened signature restores diagnostic parity without creating a second grammar authority, (2) whether policy failure remains an outcome modifier rather than an evidence modifier, (3) whether the deterministic policy precedence matches existing parser behavior, and (4) whether the cross-grammar resolver is unchanged, and explicit owner merge GO.


skulmakov-oss:
@codex review

skulmakov-oss:
@codex review

Issue #1228, opened 2026-06-30
HARNESS-M1: specify deterministic quad-state backend equivalence harness


skulmakov-oss:
## Summary

Specify a deterministic quad-state backend equivalence harness for Semantic / Pulsar / future Atlas backends.

The goal is to formalize an evidence infrastructure where `.sm` scenarios and `.expect.json` contracts are compiled, verified, executed through one or more backend adapters, and compared with bit-exact quad-state rules.

This issue is architecture/planning only. It does **not** approve runtime changes, verifier changes, SemCode changes, WGPU integration, Pulsar promotion, or Atlas implementation work.

## Architecture diagram

```mermaid
flowchart LR
    repo["Test Case Repository<br/>.sm + .expect.json contract"]
    verifier["Frontend + Verifier / Admission Guard<br/>compile · verify · admission token"]
    expected["Expected State / Assertion Contract<br/>assertions · expectations"]

    controller["Verified Execution Controller<br/>deterministic run mode<br/>step mode"]
    cpu["CPU Reference Implementation"]
    pulsar["Pulsar Packed CPU Candidate"]
    wgpu["WGPU / GPU Candidate Backend"]
    hw["Future Hardware / ASIC Backend"]
    adapter["Backend Adapter Layer"]
    comparator["Quad-State Comparator<br/>BIT_EXACT<br/>T/F/S/N masks · deltas · states"]

    ledger["Evidence Ledger"]
    stdout["CI stdout<br/>compact first-mismatch summary"]
    artifacts["Failure Artifacts<br/>.failure.json<br/>bounded trace dump<br/>WRITE_ON_FAILURE"]

    repo --> verifier
    verifier --> controller
    expected --> verifier
    verifier --> expected

    controller --> cpu
    controller --> pulsar
    controller --> wgpu
    controller --> hw

    cpu --> adapter
    pulsar --> adapter
    wgpu --> adapter
    hw --> adapter
    adapter --> comparator
    expected --> comparator

    comparator --> ledger
    ledger --> stdout
    ledger --> artifacts
```

## Core principles

- **Verifier-first execution:** `.sm` input must pass frontend compilation and verifier admission before any backend execution.
- **CPU reference authority:** CPU reference remains the baseline truth. Pulsar, WGPU/GPU, and future hardware backends are candidates that must match it.
- **Bit-exact quad comparison:** no tolerance mode for quad-state logic. Any bit mismatch is a failure.
[...]

PR #1927, opened 2026-09-14
fix(sm-sema): consume sm-front GrammarAdmission for surface dispatch (#1670)

### LINKED ISSUE #1670: FA-03-001: sm-sema surface dispatch swallows Logos parse/policy errors and retries as Rust-like source

skulmakov-oss:
Closes #1670

## Summary

Rewrites `check_source_with_profile` (`crates/sm-sema/src/std_adapters.rs`) to consume the `GrammarAdmission<T>` evidence `sm-front` already exposes (`admit_logos_program_with_profile`/`admit_program_with_profile`, SSF-09 Decision E, merged in PR #1925) instead of the old `if let Ok(logos) = ... { .. } else { retry as RustLike }` probe, which silently discarded any Logos parse/policy error and reinterpreted the same source text under the RustLike grammar.

- **Consumes `sm-front`'s `GrammarAdmission`** — lexes once, calls both `admit_*_with_profile` functions, and resolves ownership with a new `resolve_surface_authority`, implementing Decision E's frozen two-stage law (evidence strength `Exclusive > Shared > NoClaim`, then `Shared`/`Shared` resolved by parse outcome) verbatim.
- **No grammar classifier remains in `sm-sema`** — `resolve_surface_authority` is generic over each grammar's own program type and never inspects a token kind, keyword, or declaration shape. It owns none of "which forms are exclusive vs. shared" — that lives entirely in `sm-front`'s already-implemented, already-tested `admit_*` functions. This is the opposite of PR #1923's four rejected rounds, which each tried to reconstruct grammar ownership from raw tokens inside `sm-sema` and were each independently falsified.
- **Explicit profiles remain isolated** — moot for this call site: `check_source_with_profile` takes no `CompileProfile` parameter today (confirmed in the discovery note below), so it has only ever had one dispatch mode (Auto-equivalent). Nothing to isolate against.
- **Authoritative errors are preserved exactly** — the winning grammar's `FrontendError.message` is passed through unmodified into the existing `render_diag`/E0000 path; two regressions assert byte-exact equality against the direct parser call for the same source.
- **`#1919` and `#1920` are NOT fixed by this PR** — both describe the same fail-open pattern at other call sites (`smc-cli/src/app.rs`, `sm-ir/src/legacy_lowering.rs`) and remain open. `#1580` (SSF-09 umbrella) is not closed by this PR either.

## Deliberate, contract-required behavior changes

1. An authoritative Logos failure now surfaces instead of being silently retried as RustLike — this is `#1670`'s entire point.
2. Genuinely ambiguous input (both grammars independently admit it, or both hold `Exclusive` evidence) now yields a deterministic `AMBIGUOUS/CONFLICTING` error preserving both grammars' evidence (Decision E, "Evidence preservation"), instead of silently picking a side.
3. Genuinely empty/comment-only input (`NoClaim`/`NoClaim`) now yields a distinct `NO SURFACE CLAIM` error instead of silently succeeding as a trivially-empty RustLike program — required by Decision E's own Stage-1 table, confirmed no existing caller/test relied on the old trivial-success behavior.

## Discovery note

Full before/after control-flow trace, callers-affected census, and the reasoning behind each behavior change: see the discovery evidence note (attached to this PR description on request — produced before any code was touched, per the implementation brief's own discipline).

## Testing

22 new regressions in `crates/sm-sema/src/std_adapters.rs`:
- **10 direct `resolve_surface_authority` unit tests** with synthetic `GrammarAdmission` values, covering every cell of the frozen two-stage table — including `Exclusive(Err)` beating `Shared(Ok)` and `Shared`/`Shared` `Err`/`Err`, neither reliably reachable through real source text given the two grammars' own non-symmetric scan mechanics (RustLike aborts on its first unrecognized token; Logos recovers and continues).
- **12 end-to-end `check_source_with_profile` tests** using real source: malformed-Logos/malformed-RustLike non-reinterpretation, two exact-`FrontendError`-equality checks, quoted-Import ambiguity, unquoted-legacy-Import, Logos-exclusive-with-shared-Import, RustLike-exclusive-with-shared-Import, Decision E's own `Import "a.sm" as 123` worked example, bare-Pulse Exclusive evidence + policy rejection, blank-input `NO SURFACE CLAIM`, and lex-failure short-circuiting.

**Mutation/adversarial pass** (6 of 7 applicable — see below): each applied to the implementation, confirmed to break specific tests, then reverted:
- M1 (swallow owning `Err`, silently succeed) — caught by 4 tests.
- M2 (let `Shared(Ok)` beat `Exclusive(Err)`, priority inversion) — caught by 4 tests.
- M3 (resolve `Exclusive`/`Exclusive` by evaluation order instead of ambiguous) — caught by 1 test (the direct resolver unit test — no real-source fixture in this suite naturally reaches this cell, confirming Tier A's necessity).
- M4 (ownership only counts when the outcome is `Ok`, evidence-basis ignored on `Err`) — caught by 5 tests.
- M5 (collapse the `Shared`/`Exclusive` evidence-basis distinction, resolve by outcome only) — caught by 2 tests; notably the real-source end-to-end fixtures still passed under this mutation (right answer, wrong reason), confirming the direct resolver tests are load-bearing, not redundant.
- M7 (flatten the authoritative `FrontendError` to a generic message) — caught by 4 tests.
- M6 (bypass explicit-profile isolation) — **not applicable**: this call site has no `CompileProfile` parameter to isolate against (see discovery note).

**Qualification, all green on this exact HEAD:**
- `cargo fmt --check` (workspace)
- `cargo test -p sm-sema --features std --lib` — 68/68
- `cargo test --workspace` — 332/332 test-result blocks, 0 failures
- `cargo clippy -p sm-sema --features std --lib --tests -- -D warnings` — clean
- `cargo test --test legacy_guards / frontend_boundaries / ir_opt_boundaries` (boundary-enforcement) — clean
- `cargo test --test public_api_contracts` (public-api-guard) — 88/88, no new public surface
- `.\tools\7hell\run_ci.ps1` (PCC qualification fast gate) — `FAST 7HELL GATE PASSED!`

## Scope discipline

Per the implementation brief: no parser rewrite, no `FrontendError`/`CompileProfile`/`ParserProfile`/`Token` redesign, no re-lexing to classify a surface, no source-text keyword scanning, no heuristic declaration-boundary logic, no resurrection of PR #1923's production approach (its lex-error-preservation snippet is reused verbatim as explicitly blessed by Decision E's own text; its token-scanning classifier is not — that is exactly what this PR deletes the *need* for), no unrelated diagnostic cleanup, no changes to #1228/#1909/#1910, no umbrella lifecycle changes.


skulmakov-oss:
@codex review

